PR #157, opened 2017-01-10
Mark code languages for issue 146

kraigb:


karann-msft:
@kraigb the colorization isn't working. Instead we are seeing:

image

kraigb:
Yeah, I have the indentation wrong on the ```, just need to make that adjustment now across the board.

PR #159, opened 2017-01-10
Correct indentation on code markers

kraigb:


karann-msft:
do we not have a something to colorize this ?

image

kraigb:
There are a number of command line snippets like that, yes. What would you prefer from the list on http://highlightjs.readthedocs.io/en/latest/css-classes-reference.html? bash, perhaps?

kraigb:
I was also assuming to leave fixed-space snippets that are just for illustrative purposes (like a folder tree), in plain code.

PR #162, opened 2017-01-11
Adding more code tags and other corrections

kraigb:


karann-msft:
Looks great! Thanks @kraigb

PR #232, opened 2017-03-22
Update Dependency-Versions.md to reflect that version is required

geeeyetee:
Removed text concerning "If no version is specified for a dependency..." and "it's recommended to always specify a version...". This is for consistency with the following from docs/Schema/nuspec.md:

--------
The attributes for each <dependency> are as follows:...  version(Required) 
--------

Related commit: https://github.com/NuGet/docs.microsoft.com-nuget/commit/bd90112717c686b9b4b946588c20cac2a3e8deeb

### DIFF
--- a/docs/Create-Packages/Dependency-Versions.md
+++ b/docs/Create-Packages/Dependency-Versions.md
@@ -99,14 +99,6 @@ A few examples:
   <dependency id="ExamplePackage" version="[1.3.2,1.5)" />
 ```
 
-
-If no version is specified for a dependency, NuGet behaves as follows:
-
-- NuGet v2.7.2 and earlier: The **latest** package version will be used
-- NuGet v2.8 and later:  The **lowest** package version will be used
-
-For consistent behavior, it's recommended to always specify a version or version range for package dependencies.
-
 (For the curious, the NuGet version notation is inspired by Maven Version Range Specification, but is not identical to it.)
 
 ## Normalized version numbers


saldana:
### :white_check_mark: Validation status: passed


File | Status | Preview URL | Details
---- | ------ | ----------- | -------
[docs/Create-Packages/Dependency-Versions.md](https://github.com/geeeyetee/docs.microsoft.com-nuget/blob/patch-2/docs/Create-Packages/Dependency-Versions.md) | :white_check_mark:Succeeded | |

For more details, please refer to the [build report](https://opbuildstorageprod.blob.core.windows.net/report/2017%5C3%5C22%5C19a7edc0-f91d-0fd5-6e5f-4d131db05c11%5CPullRequest%5C201703221828383203-232%5Cworkflow_report.html).

**Note:** If you changed an existing file name or deleted a file, broken links in other files to the deleted or renamed file are listed only in the full build report.

docs/Create-Packages/Dependency-Versions.md
```diff
@@ -99,14 +99,6 @@ A few examples:
   <dependency id="ExamplePackage" version="[1.3.2,1.5)" />
 ```
 
-
-If no version is specified for a dependency, NuGet behaves as follows:
-
-- NuGet v2.7.2 and earlier: The **latest** package version will be used
-- NuGet v2.8 and later:  The **lowest** package version will be used
```
emgarten: I think 2.8.x also takes the highest version of the dependency. It was in 3.0.0 that this changed from what I remember. Have you tried this to verify?

docs/Create-Packages/Dependency-Versions.md
```diff
@@ -99,14 +99,6 @@ A few examples:
...
-
-For consistent behavior, it's recommended to always specify a version or version range for package dependencies.
```
emgarten: I would also mention that a minimum version should be included. Having a range such as ( , 10.0.0] makes it hard to determine the lowest version.

docs/Create-Packages/Dependency-Versions.md
kraigb: Yes, you're correct on the 2.8.x behavior--I tried it and it takes the latest.

docs/Create-Packages/Dependency-Versions.md
kraigb: I'll revise a little more.

kraigb:
@geeeyetee After looking at this more closely, I edited the file a little differently, first to include Justin's note about avoiding an open minimum version, and then also to put the recommendation to always use a version number into a callout, along with the warning about nuget.org rejecting packages without dependency versions, and also noting the change at NuGet 3.x+. See https://github.com/NuGet/docs.microsoft.com-nuget/commit/4a68c79250deab357ba8152e082baab8f1be7bfd for the exact change.

I do want to thank you for taking the time to create this PR, even though I'm closing it without merging.

geeeyetee:
Thanks, @kraigb and @emgarten! Glad to have helped.

PR #544, opened 2017-10-31
Fixing typo

sendhil:


kraigb:
Thanks for catching this and taking the time to fix it.

PR #574, opened 2017-11-20
Fix incorrect description of the /tools folder

kkm000:
The description of the `/tools` folder's usage is entirely incorrect. Visual Studio adds these package folders only to the PATH in the PM Console, but they are not added when building (which is pretty consistent with what they are used for, the PowerShell scripts that have access to the VS object model).

Double-checking.

PM console (VS 15.4.4)
```
PM> Write-Host $env:PATH
[..snip..]\TibMounter64;C:\projects\phemonoe\packages\Grpc.Tools.1.7.1\tools
```

Project file ("new" SDK-style .csproj file) with
```xml
  <Target Name="_test" BeforeTargets="Build">
    <Message Importance="high" Text="PATH=$(PATH)" />
  </Target>
```
prints
```
1>PATH=[...snip...]\TibMounter64
```
Double paranoid: Ctrl+F `Grpc` in the Output window confirms it's not there.

I have added the clarification because some NuGet packages are placing compilers into the `/tools` instead of `/build`, apparently on an invalid assumption that they would be "simply available" within the build scripts. So there is a confusion out there.

And there indeed no trace of these tool files "copied to the project folder," as the current documentation suggests (and not that there would be any reason to copy them).


kraigb:
Thanks!

PR #861, opened 2018-05-09
Make sample MSBuild code safer to copy and paste

AArnott:
Make the EmbeddingAssemblyNameFromPackageId target name illegal so that the user must make it unique, thereby mitigating the risk that it works great in private testing but stomps on other targets that also used that same name.
I also remove the `_InteropAssemblyName` property since it was unnecessary and another possible victim of stomping due to multiple packages setting the property.


kraigb:
Sounds like a good idea, thanks.

PR #875, opened 2018-05-15
fixed exclude files copy/paste error

Dzejms:
Source Files in the Exclude example were a pattern, not a discreet set of files. If I've misunderstood, let me know!

### DIFF
--- a/docs/reference/nuspec.md
+++ b/docs/reference/nuspec.md
@@ -384,9 +384,9 @@ Each `<file>` element specifies the following attributes:
 **Excluding files**
 
     Source files:
-        \tools\*.bak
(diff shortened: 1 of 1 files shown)


kraigb:
The point of that particular example is to say, if you have any number of .log and .bak, in a tools and tools\build folder, and want to exclude all of them (regardless of filename), then use wildcards in the .nuspec entries.  Perhaps is would help to say something like "tools\*.log (for example, build.log and test.log)" under Sources? Just giving a single filename is not the intent.

kraigb:
Thanks...looks more reasonable now. Thanks for the contribution!

PR #884, opened 2018-05-18
Add missing word

theryee:
I think the word "Library" was missing.


kraigb:
Yes, good catch. Thanks.

PR #1197, opened 2018-11-29
Fixed the license path

dotMorten:
The example doesn't work because the embedded file a few lines down and the license file link (corrected here) doesn't match.

### DIFF
--- a/docs/reference/nuspec.md
+++ b/docs/reference/nuspec.md
@@ -93,7 +93,7 @@ If you are using a license that hasn’t been assigned an SPDX identifier, or it
 <package>
   <metadata>
     ...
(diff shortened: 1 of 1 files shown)


nkolev92:
This is not correct. 

The target needs to match the value in license metadata.
The licenses/License.txt is a source. 

I think the example needs some more clarification.

dotMorten:
If I specify the license file to be "EULA.txt" but I pack the license file in `/licenses/EULA.txt` you'll get a packaging error.

dotMorten:
D'oh! I see what you mean now. That nuspec is being copied to the root folder.  I was using a slightly different packing approach so didn't include it like this.

PR #1414, opened 2019-04-25
Made contentFiles example clear that it goes inside of metadata section

samrueby:
contentFiles is similar to the files section. The files section went outside of metadata. It took several attempts before realizing that contentFiles needs to go inside of metadata. Made the example clearer that contentFiles goes inside of metadata.


karann-msft:
@samrueby thank you for your contribution!

Issue #2054, opened 2020-05-27
Document extension less file packing


nkolev92:
Related: https://github.com/NuGet/Home/issues/7601

nkolev92:
Related:master-nkolev92-extensionless

PR #2247, opened 2020-12-18
Document extensionless file packing

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2054

### DIFF
--- a/docs/reference/msbuild-targets.md
+++ b/docs/reference/msbuild-targets.md
@@ -251,6 +251,25 @@ When packing a license file, you need to use PackageLicenseFile property to spec
 
 [License file sample](https://github.com/NuGet/Samples/tree/master/PackageLicenseFileExample).
 
+### Packing a file without an extension
+
+In some scenarios, like when packing a license file, you might want to include a file without an extension.
+For historical reasons, NuGet & MSBuild treat paths without an extension as directories
+
+```xml
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <PackageLicenseFile>LICENSE</PackageLicenseFile>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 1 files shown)


docs/reference/msbuild-targets.md
```diff
@@ -251,6 +251,25 @@ When packing a license file, you need to use PackageLicenseFile property to spec
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="LICENSE" Pack="true" Visible="false" PackagePath=""/>
+  </ItemGroup>  
+```
+
+[File without an extension sample](https://github.com/NuGet/Samples/blob/master/PackageLicenseFileExtensionlessExample/).
+
+
```
loic-sharma, line 272: Super nit: there's an extra line here

```suggestion

```

docs/reference/msbuild-targets.md
```diff
@@ -251,6 +251,25 @@ When packing a license file, you need to use PackageLicenseFile property to spec
 
 [License file sample](https://github.com/NuGet/Samples/tree/master/PackageLicenseFileExample).
 
+### Packing a file without an extension
+
+In some scenarios, like when packing a license file, you might want to include a file without an extension.
+For historical reasons, NuGet & MSBuild treat paths without an extension as directories
```
loic-sharma, line 257: Woah this is some wizardy. Should we specifically call out what `Visible` is and what it does here? Is there a link to docs we could add?

loic-sharma, line 257: Don't forget a period! :)
```suggestion
For historical reasons, NuGet & MSBuild treat paths without an extension as directories.
```

loic-sharma (approved):

docs/reference/msbuild-targets.md
nkolev92: I'll just remove it :D 
I don't want to distract from the point of this change.

Issue #2149, opened 2020-09-02
Mention MSBuild -restore In 'Restore Target' Section


BenVillalobos:
See https://github.com/dotnet/msbuild/issues/5690#issuecomment-682299080
> Actually, this seems like more of a documentation issue, I accidentally shot myself in the foot: [here](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets)
> 
> I need to refrain from using the separate target:
> ![image](https://user-images.githubusercontent.com/29003333/91516457-283a8280-e8b1-11ea-8cb7-d353f4f66c30.png)
> 
> I got confused by this section:
> 
> ![image](https://user-images.githubusercontent.com/29003333/91516504-46a07e00-e8b1-11ea-850a-f6de2cc550f7.png)
> 
> which seems to imply running the restore target is fine, as long as you use PackageReference everywhere, but you also have to watch out for when you're running multiple targets. which is the _ONLY_ time I run restore

We should consider mentioning the 'don't run together with build' requirement in the restore section.

PR #2251, opened 2020-12-18
Add an additional note about calling restore and build in the same command

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2149

### DIFF
--- a/docs/reference/msbuild-targets.md
+++ b/docs/reference/msbuild-targets.md
@@ -379,6 +379,7 @@ An example:
 
 The `restore` target works for projects using the PackageReference format.
 `MSBuild 16.5+` also has [opt-in support](#restoring-packagereference-and-packagesconfig-with-msbuild) for the `packages.config` format.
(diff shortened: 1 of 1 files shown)


docs/reference/msbuild-targets.md
```diff
@@ -379,6 +379,7 @@ An example:
 
 The `restore` target works for projects using the PackageReference format.
 `MSBuild 16.5+` also has [opt-in support](#restoring-packagereference-and-packagesconfig-with-msbuild) for the `packages.config` format.
+Note that the `restore` target [should not be run](#restoring-and-building-with-one-msbuild-command) in combination with the `build` target.
```
loic-sharma, line 382: Consider making this have the call out color:

```suggestion

> ![Note]
> The `restore` target [should not be run](#restoring-and-building-with-one-msbuild-command) in combination with the `build` target.
```

loic-sharma (approved):

JonDouglas (approved):
Approving once it's fully rendering 👍

nkolev92:
pushed again cause it gets stuck sometimes.

Issue #787, opened 2018-03-22
Document the NuGet SDK


kraigb:
Transfer from https://github.com/NuGet/Home/issues/4062. 

* [x] Basic operations (list versions, download package, get package metadata, search for package, create package, read package): https://docs.microsoft.com/en-us/nuget/reference/nuget-client-sdk
* [x] Authenticated operations on NuGet.org (push, unlist)
* [x] Authenticated operations on feed with HTTP authentication (push, unlist, ~~restore~~)
* [x] Catalog using NuGet.Protocol.Catalog: https://docs.microsoft.com/en-us/nuget/guides/api/query-for-all-published-packages

karann-msft:
@nkolev92 - are the new ones already documented here - https://docs.microsoft.com/en-us/nuget/reference/nuget-client-sdk ?

nkolev92:
Sort of, it's not really complete.

loic-sharma:
Should we split up this issue into specific areas we'd like to add? Examples:

1. Authenticated operations on NuGet.org (push, unlist)
1. Authenticated operations on feed with HTTP authentication (push, unlist, restore)
1. Catalog using NuGet.Protocol.Catalog

Is there anything else you'd add @nkolev92?

nkolev92:
guess so...we can split into a few issues.
Personally don't really want to document restore specifics, but I see a use-case for push/unlist and authenticated feeds

nkolev92:
I think I'm covering the `authenticated operations on feed with http auth` in https://github.com/NuGet/docs.microsoft.com-nuget/pull/2254. 

It's not *all* the options, but it's the *easiest* one.

loic-sharma:
@nkolev92 Nice! I sent a pull request for nuget.org push/unlist here: https://github.com/NuGet/docs.microsoft.com-nuget/pull/2256

Do you think it'd be worth a sample for client certificate authentication?

nkolev92:
There are no examples for workign with plugins and client certs. 

I think they'd be nice to add, but are less valuable than the basic approach.

I'm not planning on adding them now, because they are significantly more work.
I'm ok with leaving this open or creating a issue tracking those 2 special examples.

loic-sharma:
OK we addressed all of the customer's asks from the original issue. Let's just close this until someone asks for more examples :)

PR #2256, opened 2020-12-22
Add package push and delete examples

loic-sharma:
Depends on: https://github.com/NuGet/Samples/pull/45

Part of: https://github.com/NuGet/docs.microsoft.com-nuget/issues/787

### DIFF
--- a/docs/reference/NuGet-Client-SDK.md
+++ b/docs/reference/NuGet-Client-SDK.md
@@ -15,18 +15,24 @@ The *NuGet Client SDK* refers to a group of NuGet packages:
 * [`NuGet.Packaging`](https://www.nuget.org/packages/NuGet.Packaging) - Used to interact with NuGet packages. `NuGet.Protocol` depends on this package
 
 You can find the source code for these packages in the [NuGet/NuGet.Client](https://github.com/NuGet/NuGet.Client) GitHub repository.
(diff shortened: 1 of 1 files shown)


zivkan (commented):
This can't be validated until the samples PR is merged. Please re-request review once the preview link shows the working sample and the docs compiler no longer reports warnings/errors.

nkolev92:
if you rebase on top of latest master it will clear some of the warnings the build is raising.

loic-sharma:
@nkolev92  @zivkan Rebased, please re-review

zivkan (approved):
My comments are probably out of scope for your PR, hence I'm approving it anyway, but:
* The first set of examples in the doc are related to NuGet.Protocol, then there are 2 NuGet.Packaging examples, and then you've added two NuGet.Protocol examples at the end. I think all the NuGet.Protocol should be grouped together.
* I think that there should be NuGet.Protocol and NuGet.Packaging headers, and they should be `H2` level headers, so they appear in the "In this article" menu, making it quick for customers to skip straight to the package that they're interested in a sample for.

nkolev92 (approved):
Looks good. 

I agree with @zivkan's feedback about separating them by package. 
I am also ok with making that out of scope

loic-sharma:
@zivkan Good idea, I added the two sections to the document

PR #2262, opened 2021-01-01
Update what-is-nuget.md

gdennie:


JonDouglas (changes requested):
Because the word "scope" is used quite a bit in package management, I believe that "means" is still the correct word to use. Is there any compelling reason to change this? Is there an alternative word that could be used altogether?

https://docs.microsoft.com/en-us/contribute/style-quick-start#use-the-microsoft-voice-principles

Common Words Used:

https://docs.microsoft.com/en-us/style-guide/welcome/

JonDouglas:
I am closing this PR for now due to inactivity. If you'd like to pick up this work again, please open this again & make the relevant changes.

Issue #1941, opened 2020-03-05
Document static graph restore (RestoreUseStaticGraphEvaluation) to increase speed of commandline restore


rrelyea:
We need to document the opt-in setting RestoreUseStaticGraphEvaluation in NuGet 5.5 and VS16.5 to speed up restore for large solutions.

@jeffkl

Grabbed the following from https://github.com/NuGet/Home/issues/8791#issuecomment-595309456:
```
<PropertyGroup>
  <RestoreUseStaticGraphEvaluation>true</RestoreUseStaticGraphEvaluation>
</PropertyGroup>
```
Or for one time use:

`msbuild /t:Restore /p:RestoreUseStaticGraphEvaluation=true`

/cc: @karann-msft @aortiz-msft

jeffkl:
Feel free to assign to me

karann-msft:
@aortiz-msft - here is the issue we spoke about.

anangaur:
Lets document it and talk about it a bit more for everybody to use it and provide us with feedback. @jeffkl can you document it? Should it go here: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets? 

P.S. We may want to refactor the pack and restore options separately. But another issue :)

nkolev92:
I fixed up the title cause this really goes beyond no-op. It's about commandline `restore` in general.

For large solutions, even actual restore could see significant improvements.

nkolev92:
https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets is one place, but I really think that we need to have in some other places too. 

That's a reference doc, not necessarily one where we advertise new features :)

JonDouglas:
Just throwing stuff out there. Does NuGet have a conceptual doc piece today just about improving restore performance? Nothing comes up other than half of Jeff's work and NuGet / CI release notes about improved caching / features. 

I feel like many aspects are scattered everywhere & not necessarily all MSBuild features are documented outside of release notes / specs, so it might be an opportunity to make up for the MSBuild 15/16 versions of features that we didn't document with respect to restore performance.

We can then expand upon the no-op / full restore scenarios alongside what all these flags mean with regards to improving overall NuGet restore performance & throw in some best practices as well.

Thoughts?

nkolev92:
> We can then expand upon the no-op / full restore scenarios alongside what all these flags mean with regards to improving overall NuGet restore performance & throw in some best practices as well.

[...]

PR #2268, opened 2021-01-07
Add static graph restore docs

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/1941.

### DIFF
--- a/docs/reference/msbuild-targets.md
+++ b/docs/reference/msbuild-targets.md
@@ -408,8 +408,8 @@ Additional restore settings may come from MSBuild properties in the project file
 | RestoreLockedMode | Run restore in locked mode. This means that restore will not reevaluate the dependencies. |
 | NuGetLockFilePath | A custom location for the lock file. The default location is next to the project and is named `packages.lock.json`. |
 | RestoreForceEvaluate | Forces restore to recompute the dependencies and update the lock file without any warning. |
-| RestorePackagesConfig | An opt in switch, that restores projects with packages.config. Support with `MSBuild -t:restore` only. |
-| RestoreUseStaticGraphEvaluation | Tells commandline restore to use static graph MSBuild evaluation instead of the standard evaluation. Static graph evaluation is an experimental feature that's significantly faster for large repos and solutions. |
+| RestorePackagesConfig | An opt-in switch, that restores projects with packages.config. Support with `MSBuild -t:restore` only. |
+| RestoreUseStaticGraphEvaluation | An opt-in switch to use static graph MSBuild evaluation instead of the standard evaluation. Static graph evaluation is an experimental feature that's significantly faster for large repos and solutions. |
 
 #### Examples
 
--- a/docs/reference/msbuild-targets.md
+++ b/docs/reference/msbuild-targets.md
@@ -474,6 +474,24 @@ This is an experimental feature that signficantly decreases the msbuild overhead
 msbuild -t:restore -p:RestoreUseStaticGraphEvaluation=true
 ```
 
+Alternatively you can enable it by setting the property in a Directory.Build.Props.
+
+> [!NOTE]
+> As of Visual Studio 2019.x and NuGet 5.x, this feature is considered experimental and opt-in. Follow [NuGet/Home#9803](https://github.com/NuGet/Home/issues/9803) for details on when this feature will be enabled by default.
+
+Static graph restore changes the msbuild part of restore, the project reading and evaluation, but not the restore algorithm! The restore algorithm is the same across all NuGet tools (NuGet.exe, MSBuild.exe, dotnet.exe and Visual Studio).
+
+In very few scenarios, static graph restore may behave differently from current restore and certain declared PackageReferences or ProjectReferences might be missing.
+
+To ease your mind and determine whether static graph works correctly for your scenario, consider running:
+
+```cli
+msbuild.exe -t:restore -p:RestoreUseStaticGraphEvaluation
+msbuild.exe -t:restore
+```
+
+NuGet should *not* report any changes. If you do see a discrepancy, please file an issue at [NuGet/Home](https://github.com/nuget/home/issues/new).
+
 ### Replacing one library from a restore graph
 
 If a restore is bringing the wrong assembly, it's possible to exclude that packages default choice, and replace it with your own choice. First with a top level `PackageReference`, exclude all assets:
--- a/docs/reference/msbuild-targets.md
+++ b/docs/reference/msbuild-targets.md
@@ -491,7 +491,7 @@ Static graph restore changes the msbuild part of restore, the project reading an
 
 In very few scenarios, static graph restore may behave differently from current restore and certain declared PackageReferences or ProjectReferences might be missing.
 
-To ease your mind and determine whether static graph works correctly for your scenario, consider running:
+To ease your mind, as a one time check, when migrating to static graph restore, consider running:
 
 ```cli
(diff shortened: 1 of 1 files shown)


docs/reference/msbuild-targets.md
```diff
@@ -464,6 +465,33 @@ msbuild -t:restore -p:RestorePackagesConfig=true
 > [!NOTE]
 > `packages.config` restore is only available with `MSBuild 16.5+`, and not with `dotnet.exe`
 
+### Restoring with MSBuild static graph evaluation
+
+With MSBuild 16.6+, NuGet has added the capability to use static graph evaluation from the commandline.
```
JonDouglas, line 470: Should use another note here to point out experimental:

> [!Note]
> With MSBuild 16.6+, NuGet has added an experimental feature to use static graph evaluation from the commandline that significantly decreases the restore time for large repositories.

docs/reference/msbuild-targets.md
```diff
@@ -464,6 +465,33 @@ msbuild -t:restore -p:RestorePackagesConfig=true
...
+This is an experimental feature that signficantly decreases the msbuild overhead of restore for large repos.
+
+```cli
+msbuild -t:restore -p:RestoreUseStaticGraphEvaluation=true
+```
+
+Alternatively you can enable it by setting the property in a Directory.Build.Props.
```
JonDouglas, line 477: Add an example of perhaps?:

<RestoreUseStaticGraphEvaluation>true</RestoreUseStaticGraphEvaluation>

docs/reference/msbuild-targets.md
```diff
@@ -464,6 +465,33 @@ msbuild -t:restore -p:RestorePackagesConfig=true
...
+
+> [!NOTE]
+> As of Visual Studio 2019.x and NuGet 5.x, this feature is considered experimental and opt-in. Follow [NuGet/Home#9803](https://github.com/NuGet/Home/issues/9803) for details on when this feature will be enabled by default.
+
+Static graph restore changes the msbuild part of restore, the project reading and evaluation, but not the restore algorithm! The restore algorithm is the same across all NuGet tools (NuGet.exe, MSBuild.exe, dotnet.exe and Visual Studio).
+
+In very few scenarios, static graph restore may behave differently from current restore and certain declared PackageReferences or ProjectReferences might be missing.
+
+To ease your mind and determine whether static graph works correctly for your scenario, consider running:
```
JonDouglas, line 486: Makes sense, maybe we just simplify to say "If static graph restore behaves differently, you can run a regular restore after".

We don't expect users to have to use two restores with this feature? Just in the case of a missing reference right?

JonDouglas (approved):
Approved with minor additions

docs/reference/msbuild-targets.md
nkolev92: I meant to suggest that they should do this only once when migrating.

docs/reference/msbuild-targets.md
nkolev92: Sure, I'll add one.

docs/reference/msbuild-targets.md
nkolev92: Played with the wording a bit. I called out that it should be a one time check.

Let me know what you think.

docs/reference/msbuild-targets.md
```diff
@@ -409,6 +409,7 @@ Additional restore settings may come from MSBuild properties in the project file
 | NuGetLockFilePath | A custom location for the lock file. The default location is next to the project and is named `packages.lock.json`. |
 | RestoreForceEvaluate | Forces restore to recompute the dependencies and update the lock file without any warning. |
 | RestorePackagesConfig | An opt in switch, that restores projects with packages.config. Support with `MSBuild -t:restore` only. |
+| RestoreUseStaticGraphEvaluation | Tells commandline restore to use static graph MSBuild evaluation instead of the standard evaluation. Static graph evaluation is an experimental feature that's significantly faster for large repos and solutions. |
```
loic-sharma, line 412: Does this also affect Visual Studio? Consider not mentioning command line:

```suggestion
| RestoreUseStaticGraphEvaluation | Opt-in switch to use static graph MSBuild evaluation instead of the standard evaluation. Static graph evaluation is an experimental feature that's significantly faster for large repos and solutions. |
```

If you prefer as is, please split "commandline" into two words :)

docs/reference/msbuild-targets.md
```diff
@@ -464,6 +465,41 @@ msbuild -t:restore -p:RestorePackagesConfig=true
 > [!NOTE]
 > `packages.config` restore is only available with `MSBuild 16.5+`, and not with `dotnet.exe`
 
+### Restoring with MSBuild static graph evaluation
+
+> [!NOTE]
+> With MSBuild 16.6+, NuGet has added an experimental feature to use static graph evaluation from the commandline that significantly decreases the restore time for large repositories.
```
loic-sharma, line 471:
```suggestion
> With MSBuild 16.6+, NuGet has added an experimental feature to use static graph evaluation from the command line that significantly improves the restore time for large repositories.
```

docs/reference/msbuild-targets.md
```diff
@@ -464,6 +465,41 @@ msbuild -t:restore -p:RestorePackagesConfig=true
...
+  </PropertyGroup>
+</Project>
+```
+
+> [!NOTE]
+> As of Visual Studio 2019.x and NuGet 5.x, this feature is considered experimental and opt-in. Follow [NuGet/Home#9803](https://github.com/NuGet/Home/issues/9803) for details on when this feature will be enabled by default.
+
+Static graph restore changes the msbuild part of restore, the project reading and evaluation, but not the restore algorithm! The restore algorithm is the same across all NuGet tools (NuGet.exe, MSBuild.exe, dotnet.exe and Visual Studio).
+
+In very few scenarios, static graph restore may behave differently from current restore and certain declared PackageReferences or ProjectReferences might be missing.
```
loic-sharma, line 492: Do we have a list of these scenarios? Is it worth including here?

loic-sharma (approved):

docs/reference/msbuild-targets.md
nkolev92: There's been a few bugs here and there, but we've been fixing those issues.

docs/reference/msbuild-targets.md
JonDouglas: When we say "migrating to static graph restore" that really just means that the user opted into the feature and should do this once to make sure everything is populated. I'm good with the wording as changed!

docs/reference/msbuild-targets.md
JonDouglas: Let's commit this change because users are very likely to put the MSBuild property into a .csproj & use VS. @nkolev92 any objections?

nkolev92: It doesn't affect VS restore at all.

The *entry* project is the one that needs it set. I wanted to avoid those details because it can get confusing.

docs/reference/msbuild-targets.md
JonDouglas: Good catch. "improvement" tends to be used more for perf related work, so let's use that.

docs/reference/msbuild-targets.md
nkolev92: changed it to make it equivalent to `RestorePackagesConfig`.

JonDouglas: Ah okay. I swore I tried using it in VS and it worked the same. It's been too long to remember though.

PR #2271, opened 2021-01-08
Update PackageIconUrl info to be current

gewarren:
Talk about now instead of the future.


JonDouglas (approved):
Thank you for the contribution @gewarren !

PR #2294, opened 2021-01-26
fix: add missing 'is' in second sentence

Geoff1900:
add missing 'is' in the second sentence of the first paragraph


JonDouglas (approved):
Thanks for the contribution @Geoff1900 !

PR #2304, opened 2021-02-05
update path to Package-restore-troubleshooting.md

trivalik:
fix dead link to troubleshooting

### DIFF
--- a/docs/consume-packages/Package-Restore.md
+++ b/docs/consume-packages/Package-Restore.md
@@ -239,4 +239,4 @@ To test the automatic package restore:
 
 ## Troubleshooting
 
-See [Troubleshoot package restore](package-restore-troubleshooting.md).
+See [Troubleshoot package restore](Package-restore-troubleshooting.md).


JonDouglas:
@trivalik Thank you for the contribution! The links are not case sensitive so this should work today. At least as I checked this out here vs. your contribution.

https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore#troubleshooting

Does the link not work for you on your end?

trivalik:
Your link is working but if you are here in github, and click the last link on https://github.com/NuGet/docs.microsoft.com-nuget/blob/master/docs/consume-packages/Package-Restore.md you come to https://github.com/NuGet/docs.microsoft.com-nuget/blob/master/docs/consume-packages/package-restore-troubleshooting.md which is not existing.

JonDouglas:
@trivalik Great catch! Let's add that in then for easier GitHub navigation. Thanks again for the contribution!

JonDouglas (approved):
Thanks for the contribution!

PR #2305, opened 2021-02-08
Fix typo

nabeelvalley:
This looks like the `gits` shouldn't be here


JonDouglas (approved):

PR #2321, opened 2021-02-13
Fix last two examples by removing hypen

RobJohnston:
`nuget` is the command, not `nuget-trusted-signers`.


JonDouglas (approved):
Thank you so much for your contribution! Great catch!

PR #2327, opened 2021-02-22
Add note about dependency confusion attack

Youssef1313:


JonDouglas:
Thank you for the contribution. Although this contribution adds a note about dependency confusion, it's in a location that is more reference than best practice. Instead, we have documentation for security here:

https://docs.microsoft.com/en-us/nuget/concepts/security-best-practices#nuget-feeds

PR #2350, opened 2021-03-11
Fix code snippet indent

austindrenski:
Silly little patch, but noticed live docs site is collapsing the markdown indent and code snippet indent for these two snippets, such that the rendered snippets are not indented.

Not entirely sure this will fix it on the live site, but it's reproducible/fixable in GFM:

- Before:
    ```xml
   <ItemGroup>
    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
   </ItemGroup>
    ```
- After:
    ```xml
    <ItemGroup>
      <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
    </ItemGroup>
    ```

Pages:
- https://docs.microsoft.com/nuget/consume-packages/install-use-packages-dotnet-cli
- https://docs.microsoft.com/nuget/quickstart/install-and-use-a-package-using-the-dotnet-cli

image

### DIFF
--- a/docs/consume-packages/install-use-packages-dotnet-cli.md
+++ b/docs/consume-packages/install-use-packages-dotnet-cli.md
@@ -42,9 +42,9 @@ This article shows you basic usage for a few of the most common dotnet CLI comma
    You can open the `.csproj` file to see the added reference:
 
     ```xml
-   <ItemGroup>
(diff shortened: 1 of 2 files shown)


JonDouglas (approved):
Thank you so much for the contribution!

PR #2356, opened 2021-03-18
Add new package ID casing policy

loic-sharma:


docs/nuget-org/nuget-org-faq.md
```diff
@@ -59,6 +59,17 @@ It is possible to share dependencies between multiple packages, making the total
...
+**How do I change my package ID casing?**
+
+NuGet.org displays the package's original casing (see [NuGetGallery#3349](https://github.com/NuGet/NuGetGallery/issues/3349)). If you'd like to change your package ID from `foo.bar` to `Foo.Bar`, please follow these steps:
+
+1. Ensure that the latest version of your package on nuget.org has the desired package ID casing. If needed, please upload a new version of your package with the desired casing.
+2. Login to nuget.org.
+3. Navigate to your package's page.
+4. Select the "Contact support" link in the sidebar and request a package ID case change.
+
+Our customer support will verify that your package's latest version has the desired package ID casing.
```
JonDouglas, line 71: Is this the only way to change this? This seems like a burden on support staff IMO. Does uploading a new package ID casing not resolve this?

JonDouglas (approved):
LGTM

docs/nuget-org/nuget-org-faq.md
anangaur: I agree with @JonDouglas here that the fix could be just take the latest version's ID casing. Easy self-serve.
The other option could be move to all small case like npm?

joelverhagen: >The other option could be move to all small case like npm?

This would be extremely disruptive, especially in our PascalCase world of .NET.

> Does uploading a new package ID casing not resolve this?

It only resolves it on Azure Search results and some V3 artifacts. Not gallery. As @loic-sharma mentioned in the change, there is a long standing issue in the GitHub.

skofman1: This is a knows bug (the casing is based on the first ever version). What is the purpose of this documentation? Are we getting lots of requests to change casing? If the requests are frequent we should just fix the bug.

docs/nuget-org/nuget-org-faq.md
```diff
@@ -59,6 +59,17 @@ It is possible to share dependencies between multiple packages, making the total
 
 Dependencies are mostly static and never change. When fixing a bug in code, the dependencies may not need to be updated. If you bundle dependencies, you end up reshipping larger packages every time. By splitting NuGet packages into related dependencies, upgrades are much more fine-grained for consumers of your package.
 
+**How do I change my package ID casing?**
+
+NuGet.org displays the package's original casing (see [NuGetGallery#3349](https://github.com/NuGet/NuGetGallery/issues/3349)). If you'd like to change your package ID from `foo.bar` to `Foo.Bar`, please follow these steps:
+
+1. Ensure that the latest version of your package on nuget.org has the desired package ID casing. If needed, please upload a new version of your package with the desired casing.
```
joelverhagen, line 66: Nit: do we have a standing on nuget.org vs. NuGet.org in our docs? I see some inconsistency in general but perhaps we can do our best for new changes.

joelverhagen: Maybe add a line to update the issue so that we can prioritize the bug fix appropriately?

skofman1 (changes requested):
I don't think we should add this to our documentation.

anangaur:
> I don't think we should add this to our documentation.

I agree, ideally the related bug should be fixed if this is a priority item for us.

docs/nuget-org/nuget-org-faq.md
JonDouglas: Yeah the bug sounds like why there is the customer support line. If someone has to manually intervene, we should just fix the bug instead.

joelverhagen (approved):
I think this is a good stop-gap until we fix the bug.

docs/nuget-org/nuget-org-faq.md
anangaur: @joelverhagen Agree it would be disruptive and hence needs to assessed based on the signals from customers if this is a real problem for many. 

> If the requests are frequent we should just fix the bug.

@skofman1 👍

shishirx34:
I agree with @skofman1, we should not document this change. This will only result in increase in requests, more often than not.. publishers want their casing to be fixed on display package page. Lets prioritize fixing the bug.

docs/nuget-org/nuget-org-faq.md
JonDouglas: There's no stance on nuget.org vs. NuGet.org. both are valid. Use of proper name is preferred, but common name is more practical.

docs/nuget-org/nuget-org-faq.md
JonDouglas: If we've had to do this more than a couple times, we should document it until a bug is fixed (however long that may be). I don't have a strong opinion as I see merit for both. Feel free to merge or close as you deem fit!

loic-sharma: Hello, thanks everyone for the discussion! We decided on Teams to fix the bug (tracked by https://github.com/NuGet/NuGetGallery/issues/3349) instead of having a manual fixup process. I will close this pull request as it is no longer needed.

@skofman1 I created this to document the process we were planning to take for the Azure SDK team to fix their IoT packages.

docs/nuget-org/nuget-org-faq.md
loic-sharma: Oops the inconsistency in my section was unintentional. This page uses NuGet.org, I meant to be consistent with that.

loic-sharma:
Closing this in favor of: https://github.com/NuGet/NuGetGallery/issues/3349

PR #2368, opened 2021-03-23
Add note for namespaces required for nuget.protocol samples

zivkan:
Fixes: https://github.com/nuget/home/issues/10023

### DIFF
--- a/docs/reference/NuGet-Client-SDK.md
+++ b/docs/reference/NuGet-Client-SDK.md
@@ -28,6 +28,9 @@ Install the `NuGet.Protocol` package to interact with HTTP and folder-based NuGe
 dotnet add package NuGet.Protocol
 ```
 
+> [!Tip]
+> `Repository.Factory` is defined in the `NuGet.Protocol.Core.Types` namespace, and the `GetCoreV3` method is an extension method defined in the `NuGet.Protocol` namespace. Therefore, you will need to add `using` statements for both namespaces.
+
 ### List package versions
 
 Find all versions of Newtonsoft.Json using the [NuGet V3 Package Content API](../api/package-base-address-resource.md#enumerate-package-versions):


JonDouglas:
@zivkan Note is fine & can merge as is, but should we just add the usings to the examples?

zivkan:
The samples on this page only show the method body, not the using statements at the top of the file. Otherwise the appropriate usings would already be visible. If we change the snippet to include the usings, then this page will get a lot longer.

What do you think is best?

JonDouglas:
Here's a few options:

1. New code snippet with all the usings
2. Usings in each snippet
3. Merge as is

My only concern with 3) is that there's other using statements like NuGet.Common & others to run the code.

zivkan:
@NuGet/nuget-client @loic-sharma  does anyone else have an opinion in which way we should go?

Do we make all our samples longer with all the using statements?  Merge my current note as-is?

For all the other usings, Intelliesense makes it very easy to find the correct using and add it. But extension methods aren't as easy to find with Intellisense, and changing the namespace would be a breaking change, hence why we haven't changed the NuGet SDK, as mentioned in the linked issue.

donnie-msft:
I say go with 3, as this just appears to be an unclear sample and that would help.

The discussion seems more about greatly improving all of our code samples (maybe we could use embedded code snippets... whatever the try.net plugin is called). In that case, you'd get option 2, plus a richer experience overall.

JonDouglas (approved):
Approving to consider merge. There's nothing blocking, just considerations.

dominoFire:
I prefer samples you can copy-paste directly on your code. It could be:

- `using` statements added as comments in the actual snippet
- Complete working code snippet
- Link to nuget/samples repo with a working sample

dominoFire (approved):

donnie-msft (approved):

zivkan:
So I had a bit of a look around to see how it's done elsewhere.

The [docs on XmlReader](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlreader?view=net-5.0) has a method that can be copy/pasted, but it doesn't have using statements, and it's not a complete example (needs a little extra code to call the method).

However, the [docs on the Version class](https://docs.microsoft.com/en-us/dotnet/api/system.version?view=net-5.0) have samples which have a `static void Main`, and have using statements, so could be copied into a `.cs` file and compiled & run as is.

The [docs on LINQ overview](https://docs.microsoft.com/en-us/dotnet/standard/linq/) don't have a single `using` statement example.

The [docs on using System.Text.Json](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?pivots=dotnet-5-0) start off with code snippets. Before using the APIs in a snippet, they have a snippet with the using statements, then a bunch of code snippets without using statements. However, at the end of the page they have two complete sample programs, including using statements, a `static void Main`, and comments showing expected output.

So, it seems to be all over the place. In general, if people copy paste our samples, Intellisense should give suggestions for which namespaces to add using statements for. For the specific API in the issue this PR fixes, it doesn't work very well if you type out the sample, rather than copy, paste, & use intellisense to add usings.

There's certainly scope for us to improve NuGet SDK samples, but our primary business is tooling, not the NuGet SDK, so I'm going to merge this as is.  Maybe if we do another docs day I'll try to remember to improve our API docs further.

Issue #2342, opened 2021-03-10
Document how to understand where a package comes/downloaded from


JonDouglas:
Conceptual or add-on piece to https://docs.microsoft.com/en-us/nuget/concepts/package-installation-process in which we discuss how to find the source of a package installation such as:

- .nupkg.metadata
- restore log
- package signature verification log
- package installation completion log
- http log

PR #2369, opened 2021-03-24
Add package source logging information

zivkan:
Fixes: #2342


docs/concepts/package-installation-process.md
```diff
@@ -47,3 +47,76 @@ The general process is as follows:
 7. (Visual Studio only) Display the package's readme file, if available, in a Visual Studio window.
 
 Enjoy your productive coding with NuGet packages!
+
+## Validating source of installed packages
```
zivkan, line 51: These headers become HTML anchors, so I feel like it's important to try to get them  right the first time, since changing them in the future will break links.

If you have any suggestsions for improvements to any of these headers, please let me know.

nkolev92: How about `Auditing`?

nkolev92 (commented):
I'm not sure this is the best location for this type of information. 

It makes the conceptual doc significantly longer by using many words to say the same thing. 

I don't have a good recommendation right now, but the conceptual docs seem like the incorrect place.

docs/concepts/package-installation-process.md
JonDouglas: +1 for Auditing as I think it's a great word for this.

JonDouglas:
I don't know of another location that makes sense for this right now. I think it logically makes sense where it is today, but I also can see it detracting value from the conceptual doc.

How about a compromise of a new doc called `Troubleshooting Installed Packages` or similar that can sit in the same concept tree?

docs/concepts/package-installation-process.md
nkolev92: We synced up offline, I think @JonDouglas's suggestion for a separate doc makes sense. 

I might've given conflicting feedback in the past, but the sheer size of content we're adding makes me want to reconsider it.

zivkan:
@JonDouglas @NuGet/nuget-client 🔔

docs/concepts/troubleshooting-installed-packages.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Troubleshooting Installed Packages
+description: How to find which package source was used for individual packages
+author: JonDouglas
```
JonDouglas, line 4: You can put your name on the doc if you'd like updates on it / feedback notifications!

JonDouglas (approved):
LGTM, consider changing the author & ms.author as you wrote the doc :)

docs/concepts/troubleshooting-installed-packages.md
```diff
@@ -0,0 +1,81 @@
...
+  "contentHash": "bw3R9q8cVNhWXNpnvWb0OGP4HadS4zvClq+T1zf7AF+tLY1haZ2AvbHidQekf4PDv1T40c6brZeT/V0IBq7cEQ==",
+  "source": "https://api.nuget.org/v3/index.json"
+}
+```
+
+> [!Note]
+> If your *global-packages* folder has packages extracted before you upgraded to a newer version of tools that has NuGet 5.9.0, the `.nupkg.metadata` file will be version 1 and will not contain the package source. You can [clear your *global-packages* folder](../consume-packages/managing-the-global-packages-and-cache-folders.md#clearing-local-folders) to ensure all packages will contain the package source.
+
+> [!Tip]
+> NuGet writes the `.nupkg.metadata` file to the *global-packages* folder only. Projects using `packages.config` use a solution packages folder, which does not create a `.nupkg.metadata` file. Unless `-NoCache` or `--no-cache` was used on the command line, the package will be in both the *global-packages*, with a `.nupkg.metadata` file, and solution packages folders, without the `.nupkg.metadata` file.
```
nkolev92, line 30: Did you verify this?

JonDouglas: I have verified the 5.9.0 item. Could be simplified to 16.9 if wanted.

nkolev92: My question is more about no-cache. 

no-cache should affect http cache only.

zivkan: You're right. I read the code not too long ago, but also not very recently, and my memory was that `NoCache` skipped the global packages folder. But that's wrong.

nkolev92 (approved):

kartheekp-ms:
I am going through this PR now. I have a quick question. Is it worth to mention somewhere that incase of `Azure Artifacts single feed` the package source URL written to `.nupkg.metada` file is single feed URL instead of the `upstream source`?

docs/concepts/troubleshooting-installed-packages.md
```diff
@@ -0,0 +1,81 @@
...
+CACHE https://api.nuget.org/v3-flatcontainer/moq/4.16.1/moq.4.16.1.nupkg
+```
+
+The URL format may be different for different NuGet HTTP server implementations, and whether it's implementing NuGet V2 or V3 HTTP protocol.
+
+If your `nuget.config` has multiple HTTP sources defined, you will see multiple requests to each package's `index.json` file, one for each source. But there will be only a single `nupkg` download for each version of the package.
+
+## Package signature log message
+
+If the package being downloaded is signed, NuGet will validate the signature and will log the following message at detailed verbosity:
```
kartheekp-ms, line 67: As per yesterday's announcement, https://github.com/dotnet/announcements/issues/180 where package verification feature has been disabled temporarily for `macOS and Linux` during `restore`.  I think `install` operation follows same code path for extracting packages. We can mention that this feature is supported only on `Windows` for now and things may change as we make progress on the above announcement.

kartheekp-ms (approved):

PR #2379, opened 2021-04-08
Symantec no longer creates certs

BrianBu01:
Removing dead link.

Well not dead, just to a service that no longer exists.


JonDouglas (approved):
Thank you! This is long overdue after the acquisition by DigiCert.

PR #2383, opened 2021-04-13
Create release notes for 5.7.1 and 5.9.1

nkolev92:

PR #2386, opened 2021-04-13
Merge main to live - Create release notes for 5.7.1 and 5.9.1 (#2383)

nkolev92:


nkolev92:
I think we should probably wait for the publishing of nuget.exe before we push the release notes live.

JonDouglas (approved):
Approving to unblock merge when binaries are ready.

PR #2390, opened 2021-04-17
Fix msbuild property links in best practice guide

loic-sharma:
It looks like the .NET guide for MSBuild properties removed their content on NuGet package properties. See: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#package-properties

They now link to NuGet's pack target documentation: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target

I've updated our links to reflect this.


JonDouglas (approved):
Thank you so much for the contribution! Let's merge!

PR #2392, opened 2021-04-18
Remove link to GoDaddy code certificate

eriawan:
Remove link to GoDaddy code certificate because GoDaddy won't provide code signing starting from June 1, 2021 and the current link will not go to code signing certificate anymore.

Reference link of code signing certificate ending schedule: https://www.godaddy.com/help/what-are-code-signing-and-driver-signing-certificates-4776

### DIFF
--- a/docs/create-packages/Sign-a-Package.md
+++ b/docs/create-packages/Sign-a-Package.md
@@ -14,7 +14,7 @@ Signed packages allows for content integrity verification checks which provides 
 
 ## Get a code signing certificate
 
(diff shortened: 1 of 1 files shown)


eriawan:
Could anyone please review this PR?

cc @zivkan @nkolev92

nkolev92:
Requesting review from folks that are better versed in signing & verification.

JonDouglas (approved):
Looks great. Thanks for the contribution!

Any issue if we merge & push this closer around June timeframe? 

I don't think it matters much, but perhaps would be best to leave up until certs cannot be created/renewed after that June 1st, 2021 date.

eriawan:
Done resolving conflict and rebased.

@JonDouglas 

As far as I know, although GoDaddy mentions that they won't provide renewal after June 1, 2021 they also stop offering new code signing certificate service. I have contacted them, and GoDaddy is not offering new code signing certificate anymore.

So IMHO I think it's better to delete link/references to GoDaddy in this page.

JonDouglas:
Sounds good, let's just merge now since we don't really have anything to wait on here.

PR #2397, opened 2021-04-21
Add .csproj instructions for package types

loic-sharma:
Reading material:

* https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D
* https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target
* https://github.com/NuGet/NuGet.Client/blob/f24bad0668193ce21a1db8cabd1ce95ba509c7f0/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTaskLogic.cs#L526-L545

I'm not sure who owns the NuGet pack experience now. Feel free to add reviewers if needed! :)

### DIFF
--- a/docs/create-packages/set-package-type.md
+++ b/docs/create-packages/set-package-type.md
@@ -28,7 +28,7 @@ Packages not marked with a type, including all packages created with earlier ver
 
 ## Custom package types
 
-You can mark your package with one or more custom package types if its use does not fit the known package types. For example, imagine if customers of the `Contoso` app can install extensions. Extension authors could use the custom package type `ContosoExtension` to identify these packages as `Contoso` app extensions.
+You can mark your package with one or more custom package types if its use does not fit the [known package types](#known-package-types). For example, imagine if customers of the `Contoso` app can install extensions. Extension authors could use the custom package type `ContosoExtension` to identify these packages as `Contoso` app extensions.
 
 > [!WARNING]
 > A package with a custom package type cannot be installed by Visual Studio or nuget.exe. See [NuGet/Home#10468](https://github.com/NuGet/Home/issues/10468) for more information.
@@ -63,7 +63,7 @@ Packages with multiple intended uses can be marked with multiple package types u
 </Project>
 ```
 
-Package types can be versioned using a `,` delimiter between the package type and its version:
+Package types can be versioned using a `,` delimiter between the package type and its [`Version`](/dotnet/api/system.version) string:
 
 ```xml
 <Project Sdk="Microsoft.NET.Sdk">
@@ -108,7 +108,7 @@ Packages with multiple intended uses may be marked with multiple package types:
 </package>
 ```
 
-Package types can be versioned:
+Package types can be versioned using a [`Version`](/dotnet/api/system.version) string:
 
 ```xml
 <?xml version="1.0" encoding="utf-8"?>
@@ -124,7 +124,6 @@ Package types can be versioned:
 ```
 
 ---
-
 
 The format of a package type string is exactly like a package ID. That is, a package type is a case-insensitive string matching the regular expression `^\w+([_.-]\w+)*$` having at least one character and at most 100 characters.
 
--- a/docs/create-packages/set-package-type.md
+++ b/docs/create-packages/set-package-type.md
@@ -128,4 +128,4 @@ Package types can be versioned:
 
 The format of a package type string is exactly like a package ID. That is, a package type is a case-insensitive string matching the regular expression `^\w+([_.-]\w+)*$` having at least one character and at most 100 characters.
 
-The format of a package type version is exactly like a [`Version`](/dotnet/api/system.version).
+The format of a package type version is exactly like a [`Version`](/dotnet/api/system.version) string. The package type version is optional and defaults to `0.0`.
--- a/docs/create-packages/set-package-type.md
+++ b/docs/create-packages/set-package-type.md
@@ -21,7 +21,8 @@ Packages can be marked with one more more *package types* to indicate its intend
 
 > [!NOTE]
 > Support for package types was added in NuGet 3.5.
-> If you don't need a custom package type, it's best to *not* explicitly set the package type. NuGet defaults to `Dependency` this type when no type is specified.
+> If you don't need a custom package type, it's best to *not* explicitly set the package type.
+> NuGet defaults to the `Dependency` type when no type is specified.
 
 # [Using dotnet CLI](#tab/dotnet)
 
--- a/docs/create-packages/set-package-type.md
+++ b/docs/create-packages/set-package-type.md
@@ -9,28 +9,58 @@ ms.topic: conceptual
 
 # Set a NuGet package type
 
-With NuGet 3.5+, packages can be marked with a specific *package type* to indicate its intended use. Packages not marked with a type, including all packages created with earlier versions of NuGet, default to the `Dependency` type.
+Packages can be marked with one more more *package types* to indicate its intended use. Packages not marked with a type, including all packages created with earlier versions of NuGet, default to the `Dependency` type.
 
 - `Dependency` type packages add build- or run-time assets to libraries and applications, and can be installed in any project type (assuming they are compatible).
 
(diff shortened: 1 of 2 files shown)


docs/create-packages/set-package-type.md
```diff
@@ -9,28 +9,55 @@ ms.topic: conceptual
...
+
+Package types can be set in the project file (`.csproj`):
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    
+    <PackageType>MyCustomType, 1.0.0.0;Dependency, 1.0.0.0</PackageType>
```
loic-sharma, line 36: Open questions:
* What is the right package type version for Dependency? The example [here](https://review.docs.microsoft.com/en-us/nuget/reference/msbuild-targets?branch=pr-en-us-2397#pack-target) ues version `2.0.0.0` for Dependency - is that correct?
* Should package type versions have segments? I followed the existing reference docs

joelverhagen: Dependency version is `0.0` (default).

What do you mean by segments?

This example is dumb. There is no such thing as Dependency 1.0.0.0 or 2.0.0.0 today.

joelverhagen: I think a better example is just `MyCustomType, 1.0.0.0;Dependency`.

loic-sharma: 4 segments as in `1.0.0.0` vs 3 as in `1.0.0`. Should it be 2 segments???

loic-sharma: Updated to `MyCustomType, 1.0.0.0;Dependency`

joelverhagen: Don't get me started on System.Version vs. NuGetVersion
https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRjz1QIE4AKKAUwHcACANWrAGcBLAeygoCJMAdJl4BKegF5x9Gg2ZsuPfkKGiRAbiA

joelverhagen: In other words, people can choose whatever segments they want and it has different meaning, at least w.r.t. to comparison.

loic-sharma: image

joelverhagen (approved):

docs/create-packages/set-package-type.md
```diff
@@ -9,28 +9,52 @@ ms.topic: conceptual
...
+
+Package types can be set in the project file (`.csproj`):
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    
+    <PackageType>MyCustomType, 1.0.0.0;Dependency</PackageType>
```
zivkan, line 37: Does `MyCustomType, 1.0.0.0` get treated as a "single" string, or does it get treated as a versioned `MyCustomType`? How/where do the versions come into play?

JonDouglas:
GitHub mobile removed @joelverhagen when requesting a review from @zkat who owns pack. Sorry for confusion!

docs/create-packages/set-package-type.md
```diff
@@ -9,28 +9,52 @@ ms.topic: conceptual
...
 
 - `Dependency` type packages add build- or run-time assets to libraries and applications, and can be installed in any project type (assuming they are compatible).
 
-- `DotnetTool` type packages are extensions to the [dotnet CLI](/dotnet/articles/core/tools/index) and are invoked from the command line. Such packages can be installed only in .NET Core projects and have no effect on restore operations. More details about these per-project extensions are available in the  [.NET Core extensibility](/dotnet/articles/core/tools/extensibility#per-project-based-extensibility) documentation.
+- `DotnetTool` type packages are .NET tools that can be installed by the [dotnet CLI](/dotnet/articles/core/tools/index).
 
 - `Template` type packages provide [custom templates](/dotnet/core/tools/custom-templates) that can be used to create files or projects like an app, service, tool, or class library.
 
 - Custom type packages use an arbitrary type identifier that conforms to the same format rules as package IDs. Any type other than `Dependency` and `DotnetTool`, however, are not recognized by the NuGet Package Manager in Visual Studio.
```
JonDouglas, line 20: Perhaps we need to add specifics as to the format rules here.

JonDouglas (approved):
Great contribution! Only a small comment on whether we should expand on the allowed input in the custom type. I have only seen strings and not versions, so maybe we should provide a note of it.

docs/create-packages/set-package-type.md
JonDouglas: And probably expand those notes https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets as well.

docs/create-packages/set-package-type.md
nkolev92: Currently a package with multiple package types can't be installed in the Package Manager UI. 

I would really prefer that we don't offer up example with multiple package types like this. 
The `Dependency` type doesn't really add any value here either. 

MyCustomType would not be installable in the PM UI, https://github.com/NuGet/NuGet.Client/blob/1328b1db6ef99fd0f0c7cf9201e4c75a366a88dd/src/NuGet.Core/NuGet.PackageManagement/InstallationCompatibility.cs#L140. 

I think that we should only offer up example of 1 package type and be explicit that those package types won't be installable through the PM UI. 

Note that dotnet add package won't do this check...sooo...complicated :D

joelverhagen: > Does `MyCustomType, 1.0.0.0` get treated as a "single" string, or does it get treated as a versioned `MyCustomType`? How/where do the versions come into play?

@zivkan, it gets parsed as a versioned `MyCustomType`. Package type names follow the same character limitations as package IDs so the comma and space would not be valid. The nuspec would look like this:

```xml
<package>
  <metadata>
    ...
    <packageTypes>
      <packageType name="MyCustomType" version="1.0.0.0" />
      <packageType name="Dependency" />
    </packageTypes>
    ...
  </metadata>
</package>
```

Side note, when the `Dependency` package type is parsed, it will get a version of `0.0`.

joelverhagen: > I think that we should only offer up example of 1 package type and be explicit that those package types won't be installable through the PM UI.

Seems fine to me. 

In my mind, an ideal situation would be if all client tools were more clear in their package type enforcement, per https://github.com/NuGet/Home/issues/10468. That is all fail for non-Dependency packages but Dependency + other package type combinations work everywhere. But before that DCR lands I agree perhaps the docs should guide people towards working scenarios.

Alternatively, we could link to the bug saying that this doesn't work in all NuGet client scenarios? I don't feel strongly one way or another.

docs/create-packages/set-package-type.md
loic-sharma: Note to self: code for parsing SDK style project's package types: https://github.com/NuGet/NuGet.Client/blob/f24bad0668193ce21a1db8cabd1ce95ba509c7f0/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTaskLogic.cs#L526-L545

docs/create-packages/set-package-type.md
loic-sharma: Thanks for the feedback y'all! I'll mark this as resolved as I:

* Added more details about delimiters and the package type format
* Removed the example where there was a custom package type plus the `Dependency` type
* Linked to https://github.com/NuGet/Home/issues/10468

Feel free to re-open this thread if you have additional suggestions

docs/create-packages/set-package-type.md
loic-sharma: Added! Let me know if you have more feedback

loic-sharma:
@JonDouglas could you take another look? I made some non-trivial changes since your last review. Thanks! :)

JonDouglas (approved):
Great examples & additions. I'm good to merge!

nkolev92 (approved):

PR #2408, opened 2021-04-26
NU5123 page does not render due to < > in <pre>

adamralph:
image


JonDouglas (approved):
Wow, thanks for the PR! The HTML must confuse the docs builder.

PR #2422, opened 2021-05-09
Typo "Nuget.Config"→"NuGet.Config"

hyoshioka0128:
https://docs.microsoft.com/en-us/nuget/consume-packages/overview-and-workflow
#PingMSFTDocs


JonDouglas (approved):
Thanks for the contribution!

PR #2426, opened 2021-05-09
Clarify how self update works

bbodenmiller:
Clarify how self update works


JonDouglas (approved):

PR #2444, opened 2021-05-29
Typo "NuGet.config"→"NuGet.Config"

hyoshioka0128:
https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-search
#pingMSFTDocs


JonDouglas (approved):
Thank you!

Issue #2399, opened 2021-04-21
Update document for 'nuget trusted-signers command add' about only 1 nupkg file acceptable on path


erdembayar:
Follow up on https://github.com/NuGet/Home/issues/10647

Due do [better security practice](https://github.com/nuget/home/issues/10647#issuecomment-801523491)  we made `nuget trusted-signers add nupkgPath` command accept only 1 nupkg file instead of many.

But current [doc](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-trusted-signers) says `where <package(s)> is one or more .nupkg files.`
So this issue is for updating the document both `nuget.exe` and `dotnet nuget trust`.

PR #2450, opened 2021-06-03
Update document for nuget trusted-signers only 1 signed package is can be passed.

erdembayar:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2399

Due do [better security practice](https://github.com/nuget/home/issues/10647#issuecomment-801523491) we made nuget trusted-signers add nupkgPath command accept only 1 nupkg file instead of many. It's consistent with other MS product practice.


heng-liu (approved):

erdembayar:
@kartheekp-ms 
Any comments here before I merge it later today?

PR #2459, opened 2021-06-14
Fix semver v2.0.0 floating version for "latest prerelease"

jzabroski:
The sample for creating a floating version for a SemVer version suffix is somewhat incorrect, in that it allows an unbounded number of suffixes without incrementing suffix numbers.

For more information, consider reading the documentation about NuGet versioning:
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#pre-release-versions

Or the SemVer spec:
https://semver.org/spec/v2.0.0.html

We may want to add an additional example for "latest build" as well as "latest prerelease", and possible add comments in the sample MSBuild XML so people can follow it more clearly.


JonDouglas (approved):
Thanks for the contribution!

PR #2478, opened 2021-07-06
packages.config must be in the project directory

ChrisGuzak:
my testing found that this file must be in the project root next to the .vcxproj file. attempts to share this file between multiple projects, putting it above (e.g. `..\packages.config`) and referencing it from the .vcxproj file, does not work.

### DIFF
--- a/docs/reference/packages-config.md
+++ b/docs/reference/packages-config.md
@@ -9,9 +9,9 @@ ms.topic: reference
 
 # packages.config reference
 
-The `packages.config` file is used in some project types to maintain the list of packages referenced by the project. This allows NuGet to easily restore the project's dependencies when the project is to be transported to a different machine, such as a build server, without all those packages.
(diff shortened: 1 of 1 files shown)


JonDouglas (commented):
This change seems to be mostly related to C++ scenarios which is less used. I think it would be best to add a note or similar in this doc rather than change the current use-case. Such note can go below these paragraphs to say "If using C++ / vcxproj the packages.config file must be located in the project directory". Thoughts?

ChrisGuzak:
Does the ..\ reference work in other languages? I assume not. My understanding is other languages don't need to use packages.config as this information is stored in  .*proj files. So, this is issue is more related to C++ and it seems appropriate here.

nkolev92 (approved):

nkolev92:
It's true for any proj files. 

I've removed the C++ reference.

PR #2364, opened 2021-03-22
Links: Nuget (2021-03)

DCtheGeek:
Monthly links cleanup. /cc @Mikejo5000

PR #2479, opened 2021-07-07
Fix broken link

riri-yu:


riri-yu:
Hey! I read the contributing guidelines before I fixed the link, but I couldn't do it in the ../folder format because the docs was in another repository.

zivkan:
The existing syntax is correct (start url with a `/`, and leave off the `.md`).  Your change will take customers off docs.microsoft.com and onto github (which won't render docs.microsoft.com specific markdown, and will show the page metadata at the top of the markdown file).

When you want to link to a page in a different "docset" (repo), find that page in your browser on docs.microsoft.com, then copy/paste the url, delete the https://, delete the hostname, and delete the locale. If the URL also contains a `?` query string, delete that too.

zivkan:
Also, when I go to https://docs.microsoft.com/en-us/nuget/reference/signed-packages-reference, the link works for me.  I'm not sure why you think it's broken?

riri-yu:
No, I didn't change it on the site. I changed it on the Github repo page. The site works fine, it was an issue on the Github pages that was assigned to me. The issue wasn't with https://docs.microsoft.com/en-us/nuget/reference/signed-packages-reference , it was a broken link **inside** that page. 

If you go here, https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/reference/Signed-Packages-Reference.md and press on **Related articles** down at the bottom, the "Verify signed packages using the dotnet CLI" has a 404.

JonDouglas:
Hey @riri-yu, 

Thanks again for helping improve our docs, it's really appreciated.

This is tricky because the link will 404 in GitHub because it's not within our product docset, yet the Microsoft docs will redirect as it's relative to the entire docset. We tend to focus on the user-facing documentation pages rather than the GitHub ones because of this.

The way to fix this within both locations is to use an absolute URL with the locale omitted https://docs.microsoft.com/dotnet/core/tools/dotnet-nuget-verify

**Reference:** https://docs.microsoft.com/en-us/contribute/how-to-write-links#links-to-pages-that-are-not-part-of-the-technical-documentation-set

We tend to avoid absolute links as they can require a bit more work to update, but this general area should be relatively fine.

Feel free to either close this PR to leave existing changes or make changes to add an absolute link & we'll review again!

zivkan:
@JonDouglas the link you referenced is for writing links to docs that are not on docs.microsoft.com.

Microsoft's technical writers periodically check for links abosolute URLs to docs.microsoft.com and remove them. For example: https://github.com/NuGet/docs.microsoft.com-nuget/pull/2364/files#diff-2342749f59a3eb5e3a14c8122a9c532cd2a481907ce9f3db2017bbb7d1128948R44

So, even if we get contributions changing the links to absolute URLs, the docs team are eventually going to change it back.

The reason is:
* If the absolute URL includes the locale code, then a customer reading the NuGet docs in one language will be redirected to another language when they click the link. Although I think the docs validator will warn when a pull request tries this.
* If the abosolute URL does not include the locale, then the webserver will redirect based on the browser's Accept-Lanaguage header, but say the customer is using French, but they intentionally navigate to the English NuGet docs, by following this link to the dotnet docs, they'll be switched back to French.

By using the existing docs URL, the "compiler" can keep the customer reading the docs on their currently selected language.

Ultimately this repo is our documentation "source code", and the docs are not intended to be viewed here on github. At least, that's my feedback.

zivkan:
Here's another PR from the docs team changing links from absolute URLs to relative URLs, to ensure they work correctly on various docs websites: https://github.com/NuGet/docs.microsoft.com-nuget/pull/2492

PR #2486, opened 2021-07-21
Fix missing text

drewnoakes:
Because the XML was not escaped in backticks, it was being removed from the rendered markdown. Consequently it was not clear what property the sentence is referring to.

The fix is to mark the property as an inline code snippet, which also makes it consistent with other property references in the document.


JonDouglas (approved):
Thanks for the contribution!

PR #2490, opened 2021-07-23
Document popularity transfers

loic-sharma:
Preview URL: https://review.docs.microsoft.com/en-us/nuget/nuget-org/deprecate-packages?branch=pr-en-us-2490

Addresses: https://github.com/NuGet/NuGetGallery/issues/7943

### DIFF
--- a/docs/nuget-org/Deprecate-packages.md
+++ b/docs/nuget-org/Deprecate-packages.md
@@ -61,3 +61,43 @@ Project `My.Test.Project` has the following deprecated packages
    > My.Sample.Lib        6.0.0      Legacy      My.Awesome.Package
 
 ```
+
+## Transfer popularity to a newer package
+
+Package authors who have deprecated a legacy package in favor of a newer package can choose to transfer the "popularity" of the legacy to the newer package to boost the its search ranking. 
+
+A common problem that package authors run into when they deprecate a legacy package in favor of a newer package is that the newer, less-popular package will rank lower in search results than the incumbent package for most queries.
+
+For example, let's say I have two packages for the same task: 
+* My deprecated legacy package, Contoso.Legacy with 3M downloads
+* My latest package, Contoso.Latest with 100K downloads
+
+NuGet.org heavily considers downloads/popularity when ranking search results. Given the search query "Contoso," Contoso.Legacy will likely rank above Contoso.Latest in the results, making it more difficult for users to discover my latest package.
+
+To solve this problem, I can apply to transfer the popularity of my deprecated legacy package to my latest package. This would cause Contoso.Latest to rank higher in search results as if had Contoso.Legacy's 3M downloads added to its existing 100K downloads. Contoso.Legacy would rank much lower, as if it had almost all of its downloads removed.  Only the internal popularity scores for the packages will be impacted, the actual download count shown for each package will not change on NuGet.org or on any client.
+
+> [!Note]
+> Popularity transfers can make it significantly harder for consumers to find the legacy package.
+
+See the table below to get a concrete idea of how a popularity transfer may impact search rankings for the query "Contoso":
+
+| Search ranking 	| Before popularity transfer     	| After popularity transfer      	|
+|----------------	|--------------------------------	|--------------------------------	|
+| 1              	| *Contoso.Legacy, 3M downloads*  	| *Contoso.Latest, 100K downloads* 	|
+| 2              	| Contoso.Scanner, 2M downloads  	| Contoso.Scanner, 2M downloads  	|
+| 3              	| Contoso.Core,  1.5M downloads  	| Contoso.Core,  1.5M downloads  	|
+| 4              	| Contoso.UI, 1M downloads       	| Contoso.UI, 1M downloads       	|
+| ...            	| ...                            	| ...                            	|
+| 20             	| *Contoso.Latest, 100K downloads* 	| *Contoso.Legacy, 3M downloads*   	|
+
+### Popularity transfer application process
+
+* The legacy packages and new packages must share all owners.
+* The new packages must be clearly related to the legacy packages in naming and function (i.e. an evolution or next generation).
+* All versions of the legacy packages must be deprecated and point to the new packages receiving the transfer.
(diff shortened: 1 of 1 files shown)


docs/nuget-org/Deprecate-packages.md
```diff
@@ -61,3 +61,82 @@ Project `My.Test.Project` has the following deprecated packages
...
+2. Email [account@nuget.org](mailto:account@nuget.org) with the deprecated package whose popularity should be transferred, and, the list of stable package(s) that should receive the popularity transfer.
+
+After the application is submitted, we will notify you of your application's acceptance or rejection (with the criteria that caused rejection). We may need to ask additional identifying questions to confirm owner identity.
+
+#### Popularity transfer requirements
+
+* The legacy packages and new packages must share all owners.
+* The new packages must be clearly related to the legacy packages in naming and function (i.e. an evolution or next generation).
+* All versions of the legacy packages must be deprecated and point to the new packages receiving the transfer.
+* The popularity transfer must not cause confusion for NuGet users or worsen the NuGet search experience.
```
jcjiang, line 104: This stands out as important to me. Looking at the table above, I can see users being confused by their search results because the actual download counts don't match the order of popularity, leading to mistrust in the nuget.org search results.

Do we have any way of communicating to users that a package's popularity has been transferred - or that a package is special/an outlier in some way? I'm thinking in a subtle, visual way that can be seen at a glance on the search results.

jcjiang (approved):
Looks great overall! Only concern I have is about communicating to users why they're getting search results that may not make sense at a glance (due to behind-the-scenes popularity adjustments) because there's a possibility that they'll just assume, "Oh, there's a bug with nuget.org search."

docs/nuget-org/Deprecate-packages.md
loic-sharma: > Looking at the table above, I can see users being confused by their search results because the actual download counts don't match the order of popularity, leading to mistrust in the nuget.org search results.

This is an excellent question. Keep in mind that search rankings tries to find results that are popular *and* relevant to the query. In other words, it's totally fine for the top results to have low downloads as long as they are the best matches for the search query. If we simply ordered results by their downloads, packages with many downloads would always drown out packages that are most relevant to the user's query. 

We did consider a visual cue but decided it wasn't necessary since we only allow popularity transfers if it *improves* the search results. If a customer were to complain that search results are unexpected, this would be a strong signal that we messed up and allowed a popularity transfer that regresses search rankings.

/cc @chgill-MSFT who drove the spec for popularity transfers

chgill-MSFT: Loic nailed the explanation exactly. The value of this feature is to actively _improve_ search results in cases where search ranking prioritizing downloads works against the best interest of our users. If we hear a lot of negative customer feedback or see in telemetry that search has been negatively impacted - we'll iterate on the experience accordingly 🙂

chgill-MSFT: Loic nailed the explanation exactly. The value of this feature is to actively _improve_ search results in cases where search ranking prioritizing downloads works against the best interest of our users. If we hear a lot of negative customer feedback or see in telemetry that search has been negatively impacted - we'll iterate on the experience accordingly 🙂

docs/nuget-org/Deprecate-packages.md
```diff
@@ -61,3 +61,82 @@ Project `My.Test.Project` has the following deprecated packages
...
+* The new packages must have a stable version.
+* The legacy package must have fewer than 5 new packages.
```
chgill-MSFT, line 106: I think mentioning this int he Package splits section might be sufficient. It may be confusing and unnecessary for most.

Otherwise, I would propose an alternative wording: "The popularity of a single legacy package can be transferred to and divided among no more than 5 newer packages. See package splits for more details about this scenario."

docs/nuget-org/Deprecate-packages.md
```diff
@@ -61,3 +61,82 @@ Project `My.Test.Project` has the following deprecated packages
...
+
+* My first deprecated legacy package, `Contoso.Legacy1`
+* My second deprecated legacy package, `Contoso.Legacy2`
+* My new consolidated package, `Contoso.Latest`
+
+After I deprecate `Contoso.Legacy1` and `Contoso.Legacy2`, I can apply to transfer their popularity to `Contoso.Latest`.
+
+#### Package splits
+
+A deprecated package's popularity can be transferred to up to 5 packages. This is useful if a package is deprecated of multiple new packages. For example, let's say I have 3 packages:
```
chgill-MSFT, line 123:
```suggestion
A deprecated package's popularity can be transferred to, and divided among, up to 5 newer packages. This is useful if the functionality of a deprecated package as been split among multiple new packages. For example, let's say I have 3 packages:
```

docs/nuget-org/Deprecate-packages.md
```diff
@@ -61,3 +61,82 @@ Project `My.Test.Project` has the following deprecated packages
...
+
+* My deprecated legacy package, `Contoso.Legacy`
+* My first new package, `Contoso.Web`
+* My second new package, `Contoso.Cloud`
+
+After I deprecate `Contoso.Legacy`, I can apply to transfer its popularity to `Contoso.Web` and `Contoso.Cloud`.
```
chgill-MSFT, line 129:
```suggestion
`Contoso.Legacy` includes both web and cloud functionality, but I decided to separate that functionality into different packages for the next generation. After I deprecate `Contoso.Legacy`, I can apply to transfer its popularity to both `Contoso.Web` and `Contoso.Cloud`.
```

chgill-MSFT (approved):
Overall it looks great! Some minor suggested changes.

docs/nuget-org/Deprecate-packages.md
loic-sharma: Nice, I like it!

docs/nuget-org/Deprecate-packages.md
loic-sharma: Agreed, I'll take it out as it is a corner case

loic-sharma: Removed

PR #2511, opened 2021-08-17
Add detail about devops not supporting snupkg

IanKemp:
Ref: https://developercommunity.visualstudio.com/t/add-snupkg-support-to-azure-devops-artifacts/657354

When (if) the support is added, this can be removed.


JonDouglas (approved):
Thank you so much for this note, this was really needed!

PR #2515, opened 2021-08-23
Add missing "=true" to RestoreUseStaticGraphEvaluation

cabadam:
Without this "=true", msbuild throws an error:

MSBUILD : error MSB1006: Property is not valid.
Switch: RestoreUseStaticGraphEvaluation

It is correct (=true) earlier in the document.


JonDouglas (approved):
Thank you for the missing value! Quite appreciated.

PR #2521, opened 2021-08-27
Quick grammar fixes

tgbarnett:
Changed "this constraints" to "this constraint" when talking about the constraint on nuget.org to not be able to delete packages.
Changed "enabling lock file" to "enabling the lock file".


JonDouglas (approved):
Thanks for the grammar improvements!

PR #2538, opened 2021-09-27
Clarify SemVer 2.0.0 prerelease sort order

jzabroski:
Note: EF Package name is not really a best practice for SemVer, so the documentation is a bit confusing.

### DIFF
--- a/docs/concepts/Package-Versioning.md
+++ b/docs/concepts/Package-Versioning.md
@@ -35,7 +35,7 @@ A specific version number is in the form *Major.Minor.Patch[-Suffix]*, where the
 1.0.1
 6.11.1231
 4.3.1-rc
-2.2.44-beta1
+2.2.44-beta.1
 ```
 
 > [!Important]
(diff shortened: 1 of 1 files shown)


docs/concepts/Package-Versioning.md
```diff
@@ -52,21 +52,24 @@ That said, package developers generally follow recognized naming conventions:
...
 1.0.1
 1.0.1-zzz
-1.0.1-rc
+1.0.1-rc.10
+1.0.1-rc.2
 1.0.1-open
 1.0.1-beta
 1.0.1-alpha2
-1.0.1-alpha
+1.0.1-alpha10
```
nkolev92, line 67: This definitely improves clarity. 👍

docs/concepts/Package-Versioning.md
```diff
@@ -52,21 +52,24 @@ That said, package developers generally follow recognized naming conventions:
...
+> NuGet 4.3.0+ supports [SemVer 2.0.0](https://semver.org/spec/v2.0.0.html), which supports pre-release numbers with dot notation, as in *1.0.1-build.23*. Dot notation is not supported with NuGet versions before 4.3.0. You can use a form like *1.0.1-build23*. It is recommended to use NuGet 4.3.0+, because nuget.org and newer NuGet clients will not treat "1.0.1-build23" as greater than "1.0.1-build3".
 
-When resolving package references and multiple package versions differ only by suffix, NuGet chooses a version without a suffix first, then applies precedence to pre-release versions in reverse alphabetical order. For example, the following versions would be chosen in the exact order shown:
+When resolving package references and multiple package versions differ only by suffix, NuGet chooses a version without a suffix first, then applies precedence to pre-release versions in reverse alphabetical order and treats dot notation numbers with numerical order. For example, the following versions would be chosen in the exact order shown:
```
nkolev92, line 57: I think this doc needs some more general work. 

For example, this section seems to refer to semver 1.x only, not semver 2.x (cause there's a 2.x section later on). 

Would tabs make it easier? 1.x and 2.x, default to 2.x and then have different explanations for both?

cc @NuGet/nuget-client @NuGet/nuget-pm what do we think?

nkolev92 (commented):
Thanks @jzabroski. 

Apologies for the overdue review. 
While I do like some parts of the change, I fear that this current doc structure is broken and some of the additions might make it worse. 

@NuGet/nuget-client @NuGet/nuget-pm 
I feel like this section is pretty broken, so it'd be nice to get some thoughts on reworking it.

docs/concepts/Package-Versioning.md
jzabroski: I guess, as a start, we may want to pick a different package than Entity Framework to demonstrate SemVer2.0 - but I am not an expert on Microsoft packages.  I just know that Joel Verhagen explained to me my incorrect application of SemVer2.0.  Alana at ProGet was also helpful in pointing out to me that SemVer 2.0 does not allow four part names, and that while the NuGet Client at various releases may allow this, the standard does not support this.

nkolev92 (approved):
I'll send a follow up PR clearing up the confusion. I think these changes have sufficient improvements to stand alone.

PR #2539, opened 2021-09-28
Fix not escaped asterisk

GKotfis:
Spaces between `*` and `-` are invalid.
Instead of: `1.1.* - *` should be: `1.1.*-*`
Same for `* - *`, should be: `*-*`

Message from `dotnet restore` command:
```bash
dotnet\sdk\5.0.300\NuGet.targets(131,5): error : '* - *' is not a valid version string
```


JonDouglas (approved):

PR #2546, opened 2021-10-15
Add package source mapping documentation

nkolev92:
Fixes https://github.com/nuget/client.engineering/issues/1027


nkolev92:
aka.ms/PackageSourceMapping should redirect to this document when merged.

PR #2570, opened 2021-10-27
Add note that search filter is specific to nuget.org

zivkan:


docs/consume-packages/Finding-and-Choosing-Packages.md
```diff
@@ -147,7 +147,8 @@ For the MSBuild equivalent to setting the license in the `.nuspec`, take a look
 NuGet package search works the same on nuget.org, from the NuGet CLI, and within the NuGet Package Manager extension in
 Visual Studio. In general, search is applied to keywords as well as package descriptions.
```
loic-sharma, line 148: This may be misleading. What do you think of this?

```suggestion
If nuget.org is your package source, you can use search syntax to further refine your search results. This works on nuget,org, in the NuGet CLI, or within the NuGet Package Manager UI in Visual Studio:
```

docs/consume-packages/Finding-and-Choosing-Packages.md
```diff
@@ -147,7 +147,8 @@ For the MSBuild equivalent to setting the license in the `.nuspec`, take a look
...
 
-- **Filtering**: You can apply a search term to a specific property by using the syntax `<property>:<term>` where
+- **Filtering**: Nuget.org supports search filters. Other package sources depend on server implementation.
```
loic-sharma, line 150: Personally I would move the "other package sources" disclaimer to the bottom

```suggestion
- **Filtering**: nuget.org supports search filters.
```

loic-sharma: You can add a note at the bottom with something like:

```
> [Note]
> The advanced search syntax is supported by nuget.org. Other package sources - like Azure Artifacts or GitHub Package Registry - may use different search syntaxes. 
```

loic-sharma (approved):

docs/consume-packages/Finding-and-Choosing-Packages.md
nkolev92: I like the version better as well, but don't feel too strong.y

nkolev92 (approved):

donnie-msft (approved):

JonDouglas (approved):

PR #2602, opened 2021-11-30
Add section on reproducible builds

### LINKED ISSUE #2586: Recommend using reproducible builds
Reproducible builds let you vet a package for unexpected code. This should be recommended in the best practice page to prevent supply chain attacks!

/cc @clairernovotny

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 07b4debc-1de6-547b-13a9-c124fc949325
* Version Independent ID: ebfff172-5713-5221-5745-0d1c9f818905
* Content: [Best practices for a secure software supply chain](https://docs.microsoft.com/en-us/nuget/concepts/security-best-practices)
* Content Source: [docs/concepts/Security-Best-Practices.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/concepts/Security-Best-Practices.md)
* Product: **nuget**
* GitHub Login: @JonDouglas
* Microsoft Alias: **jodou**

clairernovotny:
Fixes #2586

### DIFF
--- a/docs/concepts/Security-Best-Practices.md
+++ b/docs/concepts/Security-Best-Practices.md
@@ -167,7 +167,7 @@ To author sign a package, see [Sign a package](../create-packages/sign-a-package
 
 **📦🖊 Package Author**
 
-Reproducible builds creates binaries that are binary identical each time you build it, and contain source code links and compiler metadata that enable a package consumer to recreate the binary directly and validate that the build environment has not been compromised.
+Reproducible builds creates binaries that are byte-for-byte identical each time you build it, and contain source code links and compiler metadata that enable a package consumer to recreate the binary directly and validate that the build environment has not been compromised.
 
 To learn more about reproducible builds, see [Producing Packages with Source Link](https://devblogs.microsoft.com/dotnet/producing-packages-with-source-link/) and the [Reproducible Build Validation](https://github.com/dotnet/designs/blob/main/accepted/2020/reproducible-builds.md) spec.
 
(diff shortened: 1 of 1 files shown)


docs/concepts/Security-Best-Practices.md
```diff
@@ -163,6 +163,14 @@ To learn more about secret scanning, see [About secret scanning](https://docs.gi
 
 To author sign a package, see [Sign a package](../create-packages/sign-a-package.md).
 
+### Reproducible Builds
+
+**📦🖊 Package Author**
+
+Reproducible builds creates binaries that are binary identical each time you build it, and contain source code links and compiler metadata that enable a package consumer to recreate the binary directly and validate that the build environment has not been compromised.
```
loic-sharma, line 170: Can we use the term "package" instead of "binary"? This may be more approachable to newbies that are familiar with NuGet packages but not binaries.

```suggestion
Reproducible builds create packages that are identical each time you build it, and contain source code links and compiler metadata that enable a package consumer to recreate the package directly and validate that the build environment has not been compromised.
```

loic-sharma (approved):

docs/concepts/Security-Best-Practices.md
clairernovotny: The issue is that technically the package is not deterministic but the binary is. Deterministic packages had issues with time stamps and deployment that I don't know if they were addressed?

JonDouglas: The term "binary" is consistent with reproducible builds. We should keep that here in my opinion. The binaries in the package are reproducible today, not the package itself afaik.

loic-sharma: Good point @clairernovotny, those issues have not been addressed yet.

PR #2604, opened 2021-11-30
Update nuget-config-file.md

### LINKED ISSUE #2196: Clarify disabledPackageSources "value" attribute meaning
The [docs](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#disabledpackagesources) for the nuget.config specification includes a "value" attribute, like:

```xml
<disabledPackageSources>
    <add key="Contoso" value="true" />
</disabledPackageSources>
```

What isn't clear is whether the attribute `value` defaults to true, or what a value of `false` would mean -- would it make that tag a no-op? [This line](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/ConfigurationDefaults.cs#L67) in the NuGet client source code suggests that the `value` attribute is ignored.

anangaur:
Fixes #2196

### DIFF
--- a/docs/reference/nuget-config-file.md
+++ b/docs/reference/nuget-config-file.md
@@ -218,7 +218,7 @@ Stores keys for sources that use API key authentication, as set with the [`nuget
 
 ### disabledPackageSources
 
-Identified currently disabled sources. May be empty.
+Identified currently disabled sources. May be empty. Unless specific sources are disabled in this section, they are enabled.
 
 | Key | Value |
 | --- | --- |
@@ -234,6 +234,8 @@ Identified currently disabled sources. May be empty.
 <!-- Empty list -->
 <disabledPackageSources />
 ```
+
+In the above eample, `Contoso` source is disabled and this source will not be used to download or install packages.
 
 ### activePackageSource
 
--- a/docs/reference/nuget-config-file.md
+++ b/docs/reference/nuget-config-file.md
@@ -235,7 +235,7 @@ Identified currently disabled sources. May be empty. Unless specific sources are
 <disabledPackageSources />
 ```
 
-In the above eample, `Contoso` source is disabled and this source will not be used to download or install packages.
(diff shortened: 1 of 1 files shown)


zivkan (approved):

docs/reference/nuget-config-file.md
```diff
@@ -235,6 +235,8 @@ Identified currently disabled sources. May be empty.
 <disabledPackageSources />
 ```
 
+In the above eample, `Contoso` source is disabled and this source will not be used to download or install packages.
```
loic-sharma, line 238:
```suggestion
In the above example, `Contoso` source is disabled and this source will not be used to download or install packages.
```

Perhaps consider this:

```suggestion
In the example above, the package source `Contoso` is disabled and will not be used to download or install packages.
```

docs/reference/nuget-config-file.md
```diff
@@ -218,7 +218,7 @@ Stores keys for sources that use API key authentication, as set with the [`nuget
 
 ### disabledPackageSources
 
-Identified currently disabled sources. May be empty.
+Identified currently disabled sources. May be empty. Unless specific sources are disabled in this section, they are enabled.
```
loic-sharma, line 221: Is this always true? A package source can be disabled in other ways. For example, a package source from my shared nuget.config file can be disabled by adding a project specific nuget.config file that has `<clear />` in its `<packageSources>` node.

JonDouglas: That is more of a behavior of clear & merging of config files than it is with this section right?

anangaur: May be the text `section` is throwing different meanings. We can live without this line I guess and explain in the example?
Or just state:
Unless specific sources are disabled using `disabledPackageSources` , they are enabled.

JonDouglas (approved):
Approving as the scope of captured in this issue is addressed.

Issue #2415, opened 2021-05-06
There is no Solution Explorer on the right


sikosis:
This does not work. One is there is no Solution Explorer on the right, so starting off with that was bad instructions. Tried via the Tools menu and it doesn't work just says that package doesn't exist on nuget.org

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ef71b45e-0bb6-6e61-3b42-8e8f150db9fb
* Version Independent ID: 129fb470-7186-059a-fbe8-29b1caf96bf9
* Content: [Install and use a NuGet package in Visual Studio](https://docs.microsoft.com/en-us/nuget/quickstart/install-and-use-a-package-in-visual-studio)
* Content Source: [docs/quickstart/install-and-use-a-package-in-visual-studio.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/quickstart/install-and-use-a-package-in-visual-studio.md)
* Product: **nuget**
* GitHub Login: @JonDouglas
* Microsoft Alias: **jodou**

StingyJack:
@sikosis can you take a screenshot of what you do see ( sensitive details redacted)?  It will probably get rectified sooner with something like that.

PR #2606, opened 2021-11-30
Add Solution Explorer document reference

donnie-msft:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2415

This isn't part of NuGet, but it'd be convenient for there to be a link to documentation on how to open/use Solution Explorer.


JonDouglas (approved):
LGTM, merge when ready.

PR #2611, opened 2021-11-30
Clarify limitation of package source mapping

### LINKED ISSUE #2610: Document Azure DevOps limitation of Source Mapping w/ Generated Configs
Add a note on the source mapping documentation about azure pipelines limitations of feed names.

https://github.com/microsoft/azure-pipelines-tasks/issues/15542

JonDouglas:
Fixes #2610


docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -111,3 +111,4 @@ For an idea of how your source mappings may look like, refer to our [samples rep
 >
 > * There are no nuget.exe or dotnet.exe commands for managing the package source mapping configuration, see [NuGet/Home#10735](https://github.com/NuGet/Home/issues/10735).
 > * There are no means of mapping packages at package installation time, see [NuGet/Home#10730](https://github.com/NuGet/Home/issues/10730).
+> * There is a limitation when using the `DotNetCoreCLI@2` Azure Pipelines task which requires `feed-` prefixes in your source mapping configuration, see [microsoft/azure-pipelines-tasks#15542](https://github.com/microsoft/azure-pipelines-tasks/issues/15542).
```
nkolev92, line 114: I think we should recommend using NuGetAuthenticate and not using that command instead. 

That's what the eng recommended in our issue: https://github.com/NuGet/Home/issues/11406#issuecomment-982096484

JonDouglas: Are we saying the same thing here but in opposite ways?

JonDouglas: updated just to clarify

nkolev92: I think the wording of `requires you to do bla` implies that we want people to do it. 

It's not really a requirement imo, it's a bug. 

So I wouldn't call  out the whole `feed-` just say that PackageSourceMapping won't work with these tasks and tell them what to use.

JonDouglas: Are these comments to the outdated or most recent commit? I clarified majority of this already.

nkolev92: The latest version. 

> * There is a limitation when using the `DotNetCoreCLI@2` Azure Pipelines task which requires `feed-` prefixes in your source mapping configuration, it is recommended to use `NuGetAuthenticate` for authentication needs instead, see [microsoft/azure-pipelines-tasks#15542](https://github.com/microsoft/azure-pipelines-tasks/issues/15542).


I don't think we should directly or indirectly suggest that people should add `feed-` to their nuget.config. 

My suggestion is to avoid that. 

Maybe: 

> * There is a limitation when using the `DotNetCoreCLI@2` Azure Pipelines task with Package Source Mapping. 
it is recommended to use `NuGetAuthenticate` for authentication instead, and use [Command Line Task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/command-line) to invoke `dotnet`. See [microsoft/azure-pipelines-tasks#15542](https://github.com/microsoft/azure-pipelines-tasks/issues/15542).

You might need to replace the link if you go this direction.

nkolev92:
🔔 @JonDouglas

JonDouglas:
Ready to merge or need more clarification?

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -111,3 +111,4 @@ For an idea of how your source mappings may look like, refer to our [samples rep
 >
 > * There are no nuget.exe or dotnet.exe commands for managing the package source mapping configuration, see [NuGet/Home#10735](https://github.com/NuGet/Home/issues/10735).
 > * There are no means of mapping packages at package installation time, see [NuGet/Home#10730](https://github.com/NuGet/Home/issues/10730).
+> * There is a limitation when using the `DotNetCoreCLI@2` Azure Pipelines task which can be worked around by using `feed-` prefixes in your source mapping configuration, it is recommended however to use `NuGetAuthenticate` for your authentication needs instead, see [microsoft/azure-pipelines-tasks#15542](https://github.com/microsoft/azure-pipelines-tasks/issues/15542).
```
zivkan, line 114:
```suggestion
> * There is a limitation when using the `DotNetCoreCLI@2` Azure Pipelines task which can be worked around by using `feed-` prefixes in your source mapping configuration. It is recommended however to use `NuGetAuthenticate` for your authentication needs and call the dotnet cli directly from a script task. See [microsoft/azure-pipelines-tasks#15542](https://github.com/microsoft/azure-pipelines-tasks/issues/15542).
```

This feels like a run-on-sentence, so I suggest breaking it into two (or more) sentences.

The second part of my suggestion is that people inexperienced with Azure DevOps might have the impression that `NuGetAuthenticate` is used in the same scenarios as `DotNetCoreCli`, which is not the case. Therefore, maybe being explicit about needing to run the dotnet cli from a script task helps be clearer what, exactly, the recommendation is.

zivkan (approved):

Issue #2600, opened 2021-11-30
Update VS extensibility docs with latest changes


zivkan:
Possibly some new APIs have been created. Others have been deprecated. We need to update the docs.

Issue #2582, opened 2021-11-10
Document NuGet-VS version relationship


zivkan:
so customers writing VS extensions using our packages know which package version to use: https://docs.microsoft.com/en-us/nuget/visual-studio-extensibility/nuget-api-in-visual-studio

Issue #2513, opened 2021-08-19
Update background/free-thread VS extensibility APIs docs


zivkan:
Update https://docs.microsoft.com/en-us/nuget/visual-studio-extensibility/nuget-api-in-visual-studio

with the changes from https://github.com/NuGet/NuGet.Client/pull/4209

Issue #2242, opened 2020-12-18
Document NuGet.VisualStudio.Contracts


nkolev92:
cc @zivkan

PR #2612, opened 2021-11-30
Update VS extensibility docs

zivkan:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2600
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2582
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2513
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2242
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2240

### DIFF
--- a/docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
+++ b/docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
@@ -80,7 +80,7 @@ If you would like a new NuGet related API in Visual Studio, please search [NuGet
 
 1. Install the [`NuGet.VisualStudio.Contracts`](https://www.nuget.org/packages/NuGet.VisualStudio.Contracts/) package into your project, as well as [`Microsoft.VisualStudio.SDK`](https://www.nuget.org/packages/Microsoft.VisualStudio.SDK).
 
-1. Use the `IAsyncServiceProvider` to get Visual Studio's service broker, and use that to get NuGet's service. Note that [`AsyncPackage` extends `IVsAsyncServiceProvider2`](/dotnet/api/microsoft.visualstudio.shell.asyncpackage), so your class that implements `AsyncPackage` can be used as the `IAsyncServiceProvider`.
+1. Use the `IAsyncServiceProvider` to get Visual Studio's service broker, and use that to get NuGet's service. Note that [`AsyncPackage` extends `IVsAsyncServiceProvider2`](/dotnet/api/microsoft.visualstudio.shell.asyncpackage), so your class that implements `AsyncPackage` can be used as the `IAsyncServiceProvider`. Also see the docs on [`IBrokeredServiceContainer`](/dotnet/api/microsoft.visualstudio.shell.servicebroker.ibrokeredservicecontainer) and [`IServiceBroker`](/dotnet/api/microsoft.servicehub.framework.iservicebroker)
 
    ```cs
    // Your AsyncPackage implements IAsyncServiceProvider
--- a/docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
+++ b/docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
@@ -9,11 +9,29 @@ ms.topic: reference
 
 # NuGet API in Visual Studio
 
-In addition to the Package Manager UI and Console in Visual Studio, NuGet also exports some useful services through the [Managed Extensibility Framework (MEF)](/dotnet/framework/mef/index). This interface allows other components in Visual Studio to interact with NuGet, which can be used to install and uninstall packages, and to obtain information about installed packages.
+In addition to the Package Manager UI and Console in Visual Studio, NuGet also exports some useful services that other extension can use. These interface allows other components in Visual Studio to interact with NuGet, which can be used to install and uninstall packages, and to obtain information about installed packages.
 
-Over the years, NuGet has added many services all of which reside in the `NuGet.VisualStudio` namespace in the `NuGet.VisualStudio.dll` assembly:
+NuGet provides services via two different technologies, each of which have their interfaces defined in a different NuGet package. NuGet's older services are available via [the Managed Extensibility Framework (MEF)](/dotnet/framework/mef/), which are available in the package [NuGet.VisualStudio](https://www.nuget.org/packages/NuGet.VisualStudio) ([go to NuGet's MEF services](#mef-services)). There are newer APIs, designed to be usable with `async` code, available in the package `NuGet.VisualStudio.Contracts`, using a Visual Studio's `IServiceBroker` ([go to NuGet's Brokered Services](#brokered-services)).
 
-As of NuGet 3.3+, NuGet exports the following
+## Package Versions
+
+NuGet's product follows Visual Studio's version, but is 11.0 versions behind. For example, NuGet 6.0 corresponds to Visual Studio 2022 17.0, NuGet 5.11 corresponds to Visual Studio 2019 16.11, and so on.
+
+Starting from Visual Studio 17.1, NuGet's Visual Studio extensibility API packages match the version of Visual Studio that the APIs are targeting. For example, NuGet.VisualStudio and NuGet.VisualStudio.Contracts package version 17.1.0 should be used when your extension targets Visual Studio 17.1 and higher. In Visual Studio 17.0 and earlier, NuGet's package versions are the same as NuGet's product version. For example, if your extension targets Visual Studio 2022 version 17.0, you should use version 6.0 of NuGet's Visual Studio extensibility packages.
+
+## Services List
+
+### Brokered Services
+
+These services are available in the package [NuGet.VisualStudio.Contracts](https://nuget.org/packages/NuGet.VisualStudio.Contracts/).
+
+- [`INuGetProjectService`](inugetprojectservice): Methods to interact with a project. (5.7+)
+
+### MEF Services
+
+From NuGet 6.0, all of these APIs are available in the package [NuGet.VisualStudio](https://nuget.org/packages/NuGet.VisualStudio/). In NuGet 5.11 and earlier, the APIs in the namespace `NuGet.VisualStudio` are available in the package [NuGet.VisualStudio](https://nuget.org/packages.NuGet.VisualStudio/), and APIs in the namespace `NuGet.SolutionRestoreManager` are available in the package [NuGet.SolutionRestoreManager.Interop](https://www.nuget.org/packages/NuGet.SolutionRestoreManager.Interop/).
+
+#### NuGet.VisualStudio
 
 - [`IRegistryKey`](#iregistrykey-interface): Method to retrieve a value from a registry subkey. (3.3+)
 - [`IVsCredentialProvider`](#ivscredentialprovider-interface) Contains methods to get credentials for NuGet operations. (4.0+)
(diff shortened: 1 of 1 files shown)


docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
```diff
@@ -23,35 +41,72 @@ As of NuGet 3.3+, NuGet exports the following
...
+   var brokeredServiceContainer = await asyncServiceProvider.GetServiceAsync<SVsBrokeredServiceContainer, IBrokeredServiceContainer>();
+   var serviceBroker = brokeredServiceContainer.GetFullAccessServiceBroker();
+   var nugetProjectService = await serviceBroker.GetProxyAsync<INuGetProjectService>(NuGetServices.NuGetProjectServiceV1);
+   ```
+
+1. When your code no longer needs NuGet's brokered service, dispose it. For example, if you only need NuGet's brokered service during a single method call, you can wrap it in a C#`using` statement:
+
+   ```cs
+   InstalledPackagesResult installedPackagesResult;
+   using (nugetProjectService as IDisposable)
```
nkolev92, line 97: Wouldn't this throw if the service did not implement IDisposable?

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
```diff
@@ -23,35 +41,72 @@ As of NuGet 3.3+, NuGet exports the following
...
+       installedPackagesResult = await nugetProjectService.GetInstalledPackages(projectGuid, cancellationToken);
+   }
+   ```
+
+
+### MEF Services
+
+1. Install the [`NuGet.VisualStudio`](https://www.nuget.org/packages/NuGet.VisualStudio) package into your project, which contains the `NuGet.VisualStudio.dll` assembly.
+
+    In NuGet 5.11 and earlier, the package automatically sets the **Embed Interop Types** property of the assembly reference to **True**. In NuGet 6.0, NuGet followed a change in Visual Studio guidance to stop using embed interop types.
```
nkolev92, line 108: nit: phrasing, I'd say it's a Visual Studio change, rather than NuGet following a VS policy. 
So maybe say VS doesn't embed interop types in 2022 anymore.

Ideally link these?

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
```diff
@@ -23,35 +41,72 @@ As of NuGet 3.3+, NuGet exports the following
...
+
+### Brokered Services
+
+1. Install the [`NuGet.VisualStudio.Contracts`](https://www.nuget.org/packages/NuGet.VisualStudio.Contracts/) package into your project, as well as [`Microsoft.VisualStudio.SDK`](https://www.nuget.org/packages/Microsoft.VisualStudio.SDK).
+
+1. Use the `IAsyncServiceProvider` to get Visual Studio's service broker, and use that to get NuGet's service.
+
+   ```cs
+   // Your AsyncPackage implements IAsyncServiceProvider
+   IAsyncServiceProvider asyncServiceProvider = this;
```
nkolev92, line 87: Any chance we can link to relevant docs for other products?

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
```diff
@@ -335,207 +419,203 @@ public interface IRegistryKey
     }
 ```
 
-## IVsNuGetFramework interface
```
nkolev92, line 338: Why are we removing this? I don't see it added again.

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
```diff
@@ -23,35 +41,72 @@ As of NuGet 3.3+, NuGet exports the following
...
 - [`IVsPackageInstallerEvents`](#ivspackageinstallerevents-interface): Events for package install/uninstall. (3.3+)
 - [`IVsPackageInstallerProjectEvents`](#ivspackageinstallerprojectevents-interface): Batch events for package install/uninstall. (3.3+)
 - [`IVsPackageInstallerServices`](#ivspackageinstallerservices-interface): Methods to retrieve installed packages in the current solution and to check whether a given package is installed in a project. (3.3+)
-- [`IVsPackageManagerProvider`](#ivspackagemanagerprovider-interface): Methods to provide alternative Package Manager suggestions for a NuGet package. (3.3+)
-- [`IVsPackageMetadata`](#ivspackagemetadata-interface): Methods to retrieve information about an installed package. (3.3+)
-- [`IVsPackageProjectMetadata`](#ivspackageprojectmetadata-interface): Methods to retrieve information about a project where NuGet actions are being executed. (3.3+)
+- [`IVsPackageManagerProvider`](#ivspackagemanagerprovider-interface): Methods to provide alternative Package Manager suggestions for a NuGet package. (3.3 - 5.11)
```
donnie-msft, line 49: 👍

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
```diff
@@ -23,35 +41,72 @@ As of NuGet 3.3+, NuGet exports the following
...
+## Using NuGet Services
 
 > [!Warning]
-> Do not use any other types besides the public interfaces in your code, and do not reference any other NuGet assemblies, including `NuGet.Core.dll`.
+> Do not use any other types besides the public interfaces in your code, and do not reference any other NuGet assemblies, such as `NuGet.Protocol.dll`, `NuGet.Frameworks.dll`, and so on.
+
+In order to maximize the backwards compatibility promises we make, but also providing ourselves the flexibility to implement new features, performance improvements, and bug fixes in Visual Studio, we do not support the NuGet Client SDK being used in Visual Studio, and we do not provide binding redirects in `devenv.exe.config` to assemblies other than our VS extensibility contracts.
+
+If you would like a new NuGet related API in Visual Studio, please search [NuGet's Home repo](https://github.com/NuGet/Home/) and up vote any existing issues if you find a similar one. If you can't find an existing feature request to up vote, please create one.
```
donnie-msft, line 77:
```suggestion
If you would like a new NuGet related API in Visual Studio, please search [NuGet's Home repo](https://github.com/NuGet/Home/) and upvote any existing issues if you find a similar one. If you can't find an existing feature request to upvote, please create one.
```

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
zivkan: I haven't tested it myself. This was just the pattern suggested internally when I was learning about brokered services last year. In this specific case, we know that the service broker will definitely return a disposable object, even if the interface we cast to doesn't extend `IDisposable`.

I'm not sure why this pattern was used, I think it would be much better for package consumers if the interface extended `IDisposable`, but one side effect is that it will force the service implementation to implement `IDisposable`, even if it doesn't use any disposable objects itself. But that's not a big problem, we can have an empty `Dispose()` method. But changing our interface & service implementation is out of scope for this docs PR. This PR documents how it is now.

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
zivkan: My thought was that this isn't the dotnet/api reference. We're kind of treating it as a reference document (more so than any of our other NuGet docs), but at the same time we wrote at the top of the document that callers can MEF import these interfaces. `IVsNuGetFramework` can't be MEF imported. Technically `IVsPathContext` can't either, but I left it because of the following reason:

My intention was to document on this page the APIs that customers (VS extension developers) will care about. A way to use browser search to find interesting APIs like how to get the list of installed packages, how to install or uninstall, how to get configured package source, etc. I don't think `IVsNuGetFramework`'s properties or XMLDoc comments are interesting/insightful, or will help customers using find to search for an API they're looking for.

On the other hand, if they decide that one of the other interfaces has a method that they'd like to use, their code editor's intellisense or "go to definition" disassembly viewer, will show them what properties `IVsNuGetFramework` has, which is sufficient to empower them to achieve their goal.

Anyway, that was my intent, and why I deleted a bunch of interfaces that are not MEF importable, but are POCO definitions used as input/return values. Let me know if you don't agree with my intent, I can put it back.

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
zivkan: Whatever spell checker I'm using in VS Code didn't think that "upvote" was a word. Online searches seems to suggest it is 😕

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
zivkan: Also to note, the brokered service does not return the object that our service factory creates. The service broker returns a dynamically created (with Castle?) proxy, which is disposable, which in turn uses StreamJsonRpc, or other technologies, to do RPC between the proxy and our service (hence it can work over Live Share, or locally).

But all this is implementation detail that we the NuGet team don't own, we just use, plus I believe it's too much detail and would confuse customers trying to get started with using our APIs.

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
zivkan: No, we can't add links inside code blocks. I added a link to `AsyncPackage` above the code block, if that's what you think would be most useful, although you commented on the line with `IAsyncServiceProvider`, so maybe you wanted me to link to that interface's docs?  Or did you intend for me to link to all the types, the brokered service container and the type returned by `GetFullAccessServiceBroker`?

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
nkolev92: I am aware of what the service returns, but I didn';t think that you're always guaranteed to get a disposable proxy. 

Anyway, resolving.

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
nkolev92: Wondering if there's detail on how to detail with brokered services.

Maybe something that's in the extensibility docs.

nkolev92: Maybe Mads knows?

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
nkolev92: I don't have too strong of an opinion in either direciton.

nkolev92 (approved):
I'd lean 51%-49% toward including the POCO types because they're not documented anywhere else, but I'm not that invested in that right now and I certainly don't have the data to back up the value of those.

docs/visual-studio-extensibility/nuget-api-in-visual-studio.md
zivkan: Following an email thread, no, unfortunately there are no extensibility docs on using brokered services, only the reference docs on the APIs. They created (internal) issues to create some docs. I'm following the issues, and once they're done I'll be able to link them.

zivkan:
> I'd lean 51%-49% toward including the POCO types because they're not documented anywhere else

In that case, I think we should just publish our XMLDoc to docs.microsoft.com/dotnet/api. We can automate that process, and then when our interfaces change, we'll no longer need to manually update this page.

Issue #2548, opened 2021-10-19
Example is quite inadequate


achpru:
The example is quite mysterious about the suggested contents of the `<files>` element. The phrase "the files in the package will be" leaves the reader uncertain as to what it means: 1) the following list of files will be such because of some inner workings of NuGet, or 2) the package author is suggested to insert those five entries into the `<files>` element, or 3) something else.

In general, it is unclear under which conditions package authors should explicitly list references under `ref\...\` and whether under some conditions they may be omitted.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6efe48ec-f0f0-de4b-1c07-af1cc5c87d94
* Version Independent ID: fec47e16-9c71-8b73-0a41-d6808ac14700
* Content: [Select Assemblies Referenced By Projects](https://docs.microsoft.com/en-us/nuget/create-packages/select-assemblies-referenced-by-projects)
* Content Source: [docs/create-packages/Select-assemblies-referenced-by-projects.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/create-packages/Select-assemblies-referenced-by-projects.md)
* Product: **nuget**
* GitHub Login: @zivkan
* Microsoft Alias: **zivkan**

Issue #2362, opened 2021-03-20
Does NU5131 check "ref\" with "runtime\" or "runtimes\"


pete-yangpan:
According to https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks.
For different runtimes the dll should go to the "runtimes" folder. However on this wiki, it checks "runtime" without the 's'. I'm been hitting NU5131 and I am really not sure if it should have the 's' or not.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6efe48ec-f0f0-de4b-1c07-af1cc5c87d94
* Version Independent ID: fec47e16-9c71-8b73-0a41-d6808ac14700
* Content: [Select Assemblies Referenced By Projects](https://docs.microsoft.com/en-us/nuget/create-packages/select-assemblies-referenced-by-projects)
* Content Source: [docs/create-packages/Select-assemblies-referenced-by-projects.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/master/docs/create-packages/Select-assemblies-referenced-by-projects.md)
* Product: **nuget**
* GitHub Login: @zivkan
* Microsoft Alias: **zivkan**

PR #2613, opened 2021-12-01
Update docs on selecting referenced assemblies

zivkan:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2548
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2362


donnie-msft (approved):

docs/create-packages/Select-assemblies-referenced-by-projects.md
```diff
@@ -9,25 +9,44 @@ ms.topic: conceptual
...
+This gives your package consumers the flexibility to directly reference the package to get the compile time assets.
 
-> [!Note]
-> If the package contains a `runtime\` directory, NuGet may not use the assets in the `lib\` directory.
+## Not recommended: Multiple assemblies in one package
+
+`PackageReference` and `packages.config` have different features available. Whether you want to support your package consumers who use `PackageReference`, `packages.config`, or both, changes how you must author your package.
+
+NuGet's MSBuild Pack target does not support automatically including project references in the package. See [this issue](https://github.com/NuGet/Home/issues/3891), which lists some ways different developers have achieved their goal, or consider community developed tools such as [NuGetizer](https://github.com/devlooped/nugetizer). See the docs on [including content in a package](../reference/msbuild-targets.md#including-content-in-a-package) for how to use the `PackagePath` MSBuild item metadata to put files in any location in the package.
```
agr, line 31: Please don't link `this`. The text in absence of links should read as normal text. How about

> There is a [GitHub issue]((https://github.com/NuGet/Home/issues/3891)), which lists some ways...

agr (approved):

docs/create-packages/Select-assemblies-referenced-by-projects.md
```diff
@@ -9,25 +9,44 @@ ms.topic: conceptual
...
 
-## `PackageReference` support
+Our recommendation is to have one package per assembly, in which case you can use `PrivateAssets="compile"` in your `PackageReference` or `ProjectReference`. This provides more flexibility and minimizes risk that two different packages will contain different versions of the same assembly, preventing NuGet from being able to manage versions.
 
-When a project uses a package with `PackageReference` and the package contains a `ref\<tfm>\` directory, NuGet will classify those assembles as compile-time assets, while the `lib\<tfm>\` assemblies are classified as runtime assets. Assemblies in `ref\<tfm>\` are not used at runtime. This means it is necessary for any assembly in `ref\<tfm>\` to have a matching assembly in either `lib\<tfm>\` or a relevant `runtime\` directory, otherwise runtime errors will likely occur. Since assemblies in `ref\<tfm>\` are not used at runtime, they may be [metadata-only assemblies](https://github.com/dotnet/roslyn/blob/main/docs/features/refout.md) to reduce package size.
+```xml
+<ItemGroup>
+  <ProjectReference Include="..\OtherProject\OtherProject.csproj" PrivateAssets="compile" />
```
nkolev92, line 20: Given that you're overwriting PrivateAssets compile, the examples should probably include the defaults as well. 
Expect people to copy things like this.

docs/create-packages/Select-assemblies-referenced-by-projects.md
```diff
@@ -9,25 +9,44 @@ ms.topic: conceptual
 
 # Select Assemblies Referenced By Projects
 
-Explicit assembly references allows a subset of assemblies to be used for IntelliSense and compiling, while all assemblies are available at run-time. `PackageReference` and `packages.config` work differently, and as a result package authors need to take care to create the package to be compatible with both project types.
+Assemblies are used in two different ways during a build. The first is for compile, which allows the package consumer's code to compile against APIs in the assembly, and for Intellisense to give suggestions. The second is runtime, where the assembly is copied to the `bin` directory and is used at runtime.
 
-> [!Note]
-> Explicit assembly references are related to .NET assemblies. It is not a method to distribute native assemblies that are P/Invoked by a managed assembly.
+## Recommendation: Package dependencies and include/exclude assets
```
nkolev92, line 14: `Recommended` instead of `Recommendation` aligns with the language used below.

docs/create-packages/Select-assemblies-referenced-by-projects.md
```diff
@@ -9,25 +9,44 @@ ms.topic: conceptual
...
 
-## `PackageReference` support
+Our recommendation is to have one package per assembly, in which case you can add `compile` to the private assets. Note that the default private assets when none is supplied is `build;analyzers`. Therefore, you should use `PrivateAssets="compile;build;analyzers"` in your `PackageReference` or `ProjectReference`. This provides more flexibility and minimizes risk that two different packages will contain different versions of the same assembly, preventing NuGet from being able to manage versions.
```
nkolev92, line 16: > This provides more flexibility and minimizes risk that two different packages will contain different versions of the same assembly, preventing NuGet from being able to manage versions

Not sure this flows the same way you wanted it to.

Is this good or bad? I'd argue it's not great if you prevent NuGet from managing the versions.

docs/create-packages/Select-assemblies-referenced-by-projects.md
```diff
@@ -9,25 +9,44 @@ ms.topic: conceptual
 
 # Select Assemblies Referenced By Projects
 
-Explicit assembly references allows a subset of assemblies to be used for IntelliSense and compiling, while all assemblies are available at run-time. `PackageReference` and `packages.config` work differently, and as a result package authors need to take care to create the package to be compatible with both project types.
+Assemblies are used in two different ways during a build. The first is for compile, which allows the package consumer's code to compile against APIs in the assembly, and for Intellisense to give suggestions. The second is runtime, where the assembly is copied to the `bin` directory and is used at runtime.
```
nkolev92, line 12: I feel like the setup is gone here. 

Looking at the rendered doc: https://github.com/NuGet/docs.microsoft.com-nuget/blob/a0f8607854d354a9aa7f7e9da6c8beef0225366c/docs/create-packages/Select-assemblies-referenced-by-projects.md 

It's not clear what I'm trying to achieve beyond the title

docs/create-packages/Select-assemblies-referenced-by-projects.md
```diff
@@ -9,25 +9,44 @@ ms.topic: conceptual
...
+</ItemGroup>
+```
 
-> [!Important]
-> If a package contains the nuspec `<references>` element (used by `packages.config`, see below) and does not contain assemblies in `ref\<tfm>\`, NuGet will advertise the assemblies listed in the nuspec `<references>` element as both the compile and runtime assets. This means there will be runtime exceptions when the referenced assemblies need to load any other assembly in the `lib\<tfm>\` directory.
+This gives your package consumers the flexibility to directly reference the package to get the compile time assets.
 
-> [!Note]
-> If the package contains a `runtime\` directory, NuGet may not use the assets in the `lib\` directory.
+## Not recommended: Multiple assemblies in one package
```
nkolev92, line 27: If the not recommended is multiple assemblies in one package, shouldn't the top one be, recommended; one assembly per package?

nkolev92 (commented):
It might be just me, but I think the doc is now a bit more difficult to follow.

docs/create-packages/Select-assemblies-referenced-by-projects.md
```diff
@@ -9,22 +9,57 @@ ms.topic: conceptual
...
 
-## `packages.config` support
+## Not recommended: Multiple assemblies in one package
+
+`PackageReference` and `packages.config` have different features available. Whether you want to support your package consumers who use `PackageReference`, `packages.config`, or both, changes how you must author your package.
+
+NuGet's MSBuild Pack target does not support automatically including project references in the package. There is [an issue on GitHub](https://github.com/NuGet/Home/issues/3891), where community members shared ways they achieved this outcome, which usually involves using `PackagePath` MSBuild item metadata to place files anywhere in the package, as described in [the docs on including content in a package](../reference/msbuild-targets.md#including-content-in-a-package), and using [`SuppressDependenciesWhenPacking` to avoid the project references becoming package dependencies](../reference/msbuild-targets.md#pack-target-inputs). Also consider community developed tools such as [NuGetizer](https://github.com/devlooped/nugetizer).
```
nkolev92, line 50: Should we really be calling out NuGetizer? 
What warrants NuGetizer's inclusion compared to other alternatives (theoretically) ?

nkolev92 (approved):

docs/create-packages/Select-assemblies-referenced-by-projects.md
nkolev92: I approved, but please make this is addressed before merging.

zivkan: I'm not aware of any other alternatives. I'll change the docs so it says "there exists community created tooling that can do this", but I'm not sure how customers reading these docs are then supposed to discover nugetizer.

zivkan: Also, there's kind of a precedent in the hosting package page of listing a bunch of 3rd party products, some commercial: https://docs.microsoft.com/en-us/nuget/hosting-packages/overview

PR #2636, opened 2022-01-12
Highlight SolutionRestoreManager in 6.0 release notes

zivkan:
fixes: https://github.com/nuget/client.engineering/issues/1045

I didn't use the word "breaking change" anywhere, but I did explicitly tell customers using the old package they need to uninstall it. 🤷‍♂️

### DIFF
--- a/docs/release-notes/NuGet-6.0.md
+++ b/docs/release-notes/NuGet-6.0.md
@@ -122,6 +122,17 @@ The namespaces and classes remain the same, so it's compatible from an API persp
 Additionally, following [Visual Studio's change in policy](/visualstudio/extensibility/migration/migrated-assemblies?view=vs-2022&preserve-view=true), NuGet.VisualStudio no longer uses `EmbedInteropTypes`.
 Therefore, your extension will have a compile time reference to NuGet.VisualStudio.dll.
 Nuget instructs Visual Studio to use binding redirects, so your extension will not be affected when NuGet updates to newer versions and your extension is compiled against an older version of our assembly.
+For this reason, you can supress NuGet's assemblies from your vsix to reduce the download size.
+NuGet's packages will be updated to do this automatically in NuGet 6.2 (for Visual Studio 17.2).
+
+To supress NuGet's assemblies from your vsix, add the following to your project file:
+
+```xml
+<ItemGroup>
+  <SuppressFromVsix Include="NuGet.VisualStudio.dll" Visible="false" />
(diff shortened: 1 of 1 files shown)


docs/release-notes/NuGet-6.0.md
```diff
@@ -113,6 +113,16 @@ In Visual Studio 2022, NuGet has redefined the contract between NuGet package re
...
 
+### NuGet's SolutionRestoreManager Visual Studio APIs moved into NuGet.VisualStudio package
+
+NuGet.SolutionRestoreManager.Interop is no longer updated, and its APIs have been merged into the NuGet.VisualStudio package.
+If you are updating an existing Visual Studio extension to work with Visual Studio 2022 (17.0), and you were previously using NuGet.SolutionRestoreManager.Interop, you should uninstall that package and upgrade/install NuGet.VisualStudio to version 6.0.0.
+The namespaces and classes remain the same, so it's compatible from an API perspective.
+
+Additionally, following [Visual Studio's change in policy](/visualstudio/extensibility/migration/migrated-assemblies?view=vs-2022&preserve-view=true), NuGet.VisualStudio no longer uses `EmbedInteropTypes`.
+Therefore, your extension will have a compile time reference to NuGet.VisualStudio.dll.
+Nuget instructs Visual Studio to use binding redirects, so your extension will not be affected when NuGet updates to newer versions and your extension is compiled against an older version of our assembly.
```
nkolev92, line 124: Add `do not package nuget.visualstudio` to into your extension?

JonDouglas (approved):
🚢

docs/release-notes/NuGet-6.0.md
zivkan: I thought about that, but how would customers actually do that? NuGet.Client has some custom logic for our vsixignore file. 

With the binding redirect that NuGet instructs Visual Studio to create, with the CodeBase that points to our directory, it should force all extensions to use our assembly. Therefore, in theory there shouldn't be any problem. It will just reduce the customer's vsix size a little. But unless there's an easy/obvious way to achieve this, I think adding this to the docs will confuse customers.

zivkan: it also feels like something we should change in our package so it's default for customers, not something they should have to do manually. I'm going to spend a short time trying to figure out how the VS SDK does it (if it even does), but I'm not going to spend a whole lot of time on this

zivkan: ok, updated. Also created https://github.com/NuGet/Home/issues/11507

Issue #2587, opened 2021-11-15
Add video tutorial to Package Source Mapping docs


chgill-MSFT:
Consider adding https://www.youtube.com/watch?v=G6P38Dn69Ro from .NET conf to the [Package Source Mapping docs](https://docs.microsoft.com/en-us/nuget/consume-packages/package-source-mapping)

@JonDouglas what do you think?

JonDouglas:
Absolutely, send a pr

okayb7701:
#01732162794

joelverhagen:
This is in PROD.

PR #2654, opened 2022-01-31
Add link to source mapping video to doc

joelverhagen:
Address https://github.com/NuGet/docs.microsoft.com-nuget/issues/2587

### DIFF
--- a/docs/consume-packages/Package-Source-Mapping.md
+++ b/docs/consume-packages/Package-Source-Mapping.md
@@ -24,7 +24,7 @@ Older tooling will ignore the Package Source Mapping configuration. To use this 
 
 Package Source Mappings will apply to all project types – including .NET Framework – as long as compatible tooling is used.
 
-## Video walkthough
+## Video walkthrough
 
 For a video-based overview of the Package Source Mapping feature, consider watching the [Secure your NuGet packages with Package Source Mapping](https://www.youtube.com/watch?v=G6P38Dn69Ro) video on YouTube.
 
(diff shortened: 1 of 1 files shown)


docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -24,6 +24,10 @@ Older tooling will ignore the Package Source Mapping configuration. To use this
 
 Package Source Mappings will apply to all project types – including .NET Framework – as long as compatible tooling is used.
 
+## Related video
```
nkolev92, line 27: I love that we're adding this. 

I'm wondering if we can tweak the title a little bit. Should it maybe be a video walkthrough or something like that?

nkolev92 (commented):
Wondering if we need a better title.

docs/consume-packages/Package-Source-Mapping.md
joelverhagen: Sounds good! I didn't know what to put so I stole the heading on https://docs.microsoft.com/en-us/nuget/what-is-nuget. I'll do Video walkthrough.

joelverhagen:
```suggestion
## Video walkthough
```

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -24,6 +24,10 @@ Older tooling will ignore the Package Source Mapping configuration. To use this
 
 Package Source Mappings will apply to all project types – including .NET Framework – as long as compatible tooling is used.
 
+## Video walkthough
```
joelverhagen, line 27:
```suggestion
## Video walkthrough
```

docs/consume-packages/Package-Source-Mapping.md
joelverhagen: image
☠️

nkolev92 (approved):

Issue #2435, opened 2021-05-17
Document IVsProjectRestoreInfoSource and IVsSolutionRestoreService4


nkolev92:
Documentation for https://github.com/NuGet/NuGet.Client/pull/4058

PR #2657, opened 2022-02-02
Document IVsProjectRestoreInfoSource

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2435


zivkan (approved):

JonDouglas (approved):
LGTM. Is it worthwhile to get these interfaces and thus SDKs into API docs? 

I assume these comments are from the existing code so they are documented?

nkolev92:
> LGTM. Is it worthwhile to get these interfaces and thus SDKs into API docs?
> 
> I assume these comments are from the existing code so they are documented?

We could. These are the true APIs that we actually support.

Issue #2349, opened 2021-03-11
Add an explicit mention that floating versions are not affected by package listedness


nkolev92:
Covered in https://docs.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages already. 

I recall this mention existing, but might've been accidentally removed

Issue #1872, opened 2020-02-04
Document ExcludeRestorePackageImports


nkolev92:
See https://github.com/NuGet/Home/issues/3604

This is used to prevent build targets/props brought through packages from affecting restore because it would change the restore graph depending on whether the project has been restored earlier or not.

karann-msft:
@nkolev92 - I am going to need your help on this one :)

PR #2660, opened 2022-02-03
Document RestoreRepositoryPath, ExcludeRestorePackageImports and listendess impact to floating versions

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2389
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/1872
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2349

### DIFF
--- a/docs/concepts/Package-Versioning.md
+++ b/docs/concepts/Package-Versioning.md
@@ -163,6 +163,9 @@ Always specify a version or version range for package dependencies in project fi
 | \*-\* | 1.1.0 <br> 1.1.1 <br> 1.1.2-alpha <br> 1.3.0-beta  | 1.3.0-beta | The highest version including the not stable versions. | Available in Visual Studio version 16.6, NuGet version 5.6, .NET Core SDK version 3.1.300 |
 | 1.1.\*-\* | 1.1.0 <br> 1.1.1 <br> 1.1.2-alpha <br> 1.1.2-beta <br> 1.3.0-beta  | 1.1.2-beta | The highest version respecting the pattern and including the not stable versions. | Available in Visual Studio version 16.6, NuGet version 5.6, .NET Core SDK version 3.1.300 |
 
+> [!Note]
+> Whether a package is listed does not affected the floating version resolution.
+
 **References in `packages.config`:**
 
 In `packages.config`, every dependency is listed with an exact `version` attribute that's used when restoring packages. The `allowedVersions` attribute is used only during update operations to constrain the versions to which the package might be updated.
--- a/docs/concepts/Package-Versioning.md
+++ b/docs/concepts/Package-Versioning.md
@@ -164,7 +164,7 @@ Always specify a version or version range for package dependencies in project fi
 | 1.1.\*-\* | 1.1.0 <br> 1.1.1 <br> 1.1.2-alpha <br> 1.1.2-beta <br> 1.3.0-beta  | 1.1.2-beta | The highest version respecting the pattern and including the not stable versions. | Available in Visual Studio version 16.6, NuGet version 5.6, .NET Core SDK version 3.1.300 |
 
 > [!Note]
-> Whether a package is listed does not affected the floating version resolution.
+> Floating version resolution does not take into account whether or not a package is listed.
 
 **References in `packages.config`:**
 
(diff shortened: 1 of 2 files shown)


docs/concepts/Package-Versioning.md
```diff
@@ -163,6 +163,9 @@ Always specify a version or version range for package dependencies in project fi
 | \*-\* | 1.1.0 <br> 1.1.1 <br> 1.1.2-alpha <br> 1.3.0-beta  | 1.3.0-beta | The highest version including the not stable versions. | Available in Visual Studio version 16.6, NuGet version 5.6, .NET Core SDK version 3.1.300 |
 | 1.1.\*-\* | 1.1.0 <br> 1.1.1 <br> 1.1.2-alpha <br> 1.1.2-beta <br> 1.3.0-beta  | 1.1.2-beta | The highest version respecting the pattern and including the not stable versions. | Available in Visual Studio version 16.6, NuGet version 5.6, .NET Core SDK version 3.1.300 |
 
+> [!Note]
+> Whether a package is listed does not affected the floating version resolution.
```
JonDouglas, line 167: Should we expand on this?

We're trying to say that floating version resolution doesn't care if it's listed or not right?

Maybe:

`Floating version resolution does not take into account whether or not a package is listed or unlisted. It will find the most suitable match for the floating version defined.`

JonDouglas (approved):
Added a consideration for the note as I'm not fully clear on it.

docs/reference/msbuild-targets.md
```diff
@@ -442,8 +442,12 @@ Additional restore settings may come from MSBuild properties in the project file
 | `NuGetLockFilePath` | A custom location for the lock file. The default location is next to the project and is named `packages.lock.json`. |
 | `RestoreForceEvaluate` | Forces restore to recompute the dependencies and update the lock file without any warning. |
 | `RestorePackagesConfig` | An opt-in switch, that restores projects with packages.config. Support with `MSBuild -t:restore` only. |
+| `RestoreRepositoryPath` | packages.config only. Specifies the packages directory to which the packages should be restored. SolutionDirectory will be used if not specified. |
```
loic-sharma, line 445:
```suggestion
| `RestoreRepositoryPath` | packages.config only. Specifies the packages directory to which the packages should be restored. `SolutionDirectory` will be used if not specified. |
```

docs/reference/msbuild-targets.md
```diff
@@ -442,8 +442,12 @@ Additional restore settings may come from MSBuild properties in the project file
...
 | `RestoreUseStaticGraphEvaluation` | An opt-in switch to use static graph MSBuild evaluation instead of the standard evaluation. Static graph evaluation is an experimental feature that's significantly faster for large repos and solutions. |
 
+The `ExcludeRestorePackageImports` property is used by NuGet in PackageReference to ensure that build targets from packages brought in by restore cannot affect the next restore.
+As such, this is not a property that should be modified or set in any MSBuild files.
```
loic-sharma, line 449: Why document this property if users should never use it?

loic-sharma (approved):

docs/reference/msbuild-targets.md
nkolev92: Because people do :D 

I know of at least 4-5 instances where people used this property and they were getting inconsistent results.

docs/concepts/Package-Versioning.md
nkolev92: I like the first sentence better than what I have.

Do we need the 2nd sentence? To me it sounds a bit redundant.

docs/reference/msbuild-targets.md
loic-sharma: Should we rename this property to something like `_ExcludeRestorePackageImports`?

docs/concepts/Package-Versioning.md
erdembayar: Does it mean if can resolve to unlisted version?

nkolev92: Yes, it means exactly that.

docs/reference/msbuild-targets.md
nkolev92: The next time restore is run.

nkolev92: > Should we rename this property to something like _ExcludeRestorePackageImports?

Too late for that. It's been used in way too many places.

nkolev92: I tweaked it per your suggestion @loic-sharma. Thanks!

PR #2664, opened 2022-02-04
Document enhanced HTTP retry environment variables

jeffkl:
Document new environment variables for configuring enhanced HTTP retry.

Rendered: https://github.com/NuGet/docs.microsoft.com-nuget/blob/8f15af977ce90cf66b1505cbe0208bd97ad95336/docs/reference/cli-reference/cli-ref-environment-variables.md

Environment variables are named here: https://github.com/NuGet/NuGet.Client/pull/4436


docs/reference/cli-reference/cli-ref-environment-variables.md
```diff
@@ -34,4 +34,6 @@ The list of properties documented in [NuGet pack and restore as MSBuild targets]
...
 | NUGET_XMLDOC_MODE | Determines how assemblies XML documentation file extraction should be handled. | Supported modes are *skip* (do not extract XML documentation files), *compress* (store XML doc files as a zip archive) or *none* (default, treat XML doc files as regular files). |
 | NUGET_CERT_REVOCATION_MODE | Determines how the revocation status check of the certificate used to sign a package, is performed when a signed package is installed or restored. When not set, defaults to `online`.| Possible values *online* (default), *offline*.  Related to [NU3028](../errors-and-warnings/NU3028.md) |
-
+| NUGET_ENABLE_ENHANCED_HTTP_RETRY | Enables or disables enhanced HTTP retry in NuGet. | Possible values are `true`  or `false` (default) |
```
nkolev92, line 37: Do we want to keep this long term? Should we just not have it?

jeffkl: Not long term but I'm documenting it since its now on by default and people might need to turn it off as an escape hatch for some unforseen reason.  Once we get rid of the old code, I'll remove this env var from the docs.

Issue #2033, opened 2020-05-13
Document configure max http request in NuGet


nkolev92:
See spec https://github.com/NuGet/Home/wiki/Config-Max-Http-Request-In-NuGet

See https://github.com/NuGet/Home/issues/4538

RelateD:https://github.com/NuGet/docs.microsoft.com-nuget/issues/776

Example:

```xml
<config>
    <add key="maxHttpRequestsPerSource" value="16" />
</config>
```

fotto1:
Thanks for the hotfix @nkolev92 , I will try this. Anyway will a proper handling of 429 ever implemented?

I working for a large organization and will not the only one with this problem. Also other companies will face same issue when they using nuget package manager which follow HTTP Protocol Version 3

PR #2682, opened 2022-03-02
add maxHttpRequestsPerSource to docs

kartheekp-ms:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2033


erdembayar (approved):

docs/reference/nuget-config-file.md
```diff
@@ -35,6 +35,7 @@ Contains miscellaneous configuration settings, which can be set using the [`nuge
...
+| maxHttpRequestsPerSource (projects using PackageReference only)| Controls the maximum number of parallel requests sent from NuGet to every package source for package dependency resolution and downloads. The default value is `256` which is derived from `HttpClientHandler.MaxConnectionsPerServer` property. This setting has no impact on `.NET Core` for `Mac OS` because the throttling limit is set to `16` to avoid too many open files error. `nuget.exe` sets `ServicePointManager.DefaultConnectionLimit` property to `64` on `Windows` and to `1` on `Mono` to control maximum number of concurrent connections allowed for every package source. Hence configuring `maxHttpRequestsPerSource` property to a value less than `64` could impact performance of `nuget.exe` on `Windows`. |
```
nkolev92, line 38: How confident are we that it's PackageReference only? 
I was under the impression that it's both.

kartheekp-ms: Great question Nikolche. You are correct. After reading [wiki](https://github.com/NuGet/Home/wiki/Config-Max-Http-Request-In-NuGet) especially "Package.config use maxDegreeOfParallelism  which is set to Environment.ProcessorCount by default for [dependency fetching](https://github.com/NuGet/NuGet.Client/blob/f82a7ac6b54719225cda5f3317e3c31d5843279d/src/NuGet.Core/NuGet.PackageManagement/Resolution/ResolverGather.cs#L42) and [package download & extraction](https://github.com/NuGet/NuGet.Client/blob/9e1e8b52818b4bc5dbd7e8b6049fcdf15cad04b2/src/NuGet.Core/NuGet.PackageManagement/IDE/PackageRestoreManager.cs#L393)", I thought this setting is only applicable to `PR` projects which seems to be incorrect. I pushed a new commit with updated text. Please let me know what you think. I am happy to learn more if the content is incorrect.

docs/reference/nuget-config-file.md
```diff
@@ -35,6 +35,7 @@ Contains miscellaneous configuration settings, which can be set using the [`nuge
...
+| maxHttpRequestsPerSource | Controls the maximum number of parallel requests sent from NuGet to every package source for package dependency resolution and downloads. The default value on `.NET Core` is `Int32.MaxValue` which is derived from `HttpClientHandler.MaxConnectionsPerServer` property. This setting has no impact on `.NET Core` for `Mac OS` because the throttling limit is set to `16` to avoid too many open files error. `NuGet client tools` that runs on `.NET Framework` such as `Visual Studio` and `nuget.exe` sets `ServicePointManager.DefaultConnectionLimit` property to `64` on `Windows` and to `1` on `Mono` to control maximum number of concurrent connections allowed for every package source. Hence configuring `maxHttpRequestsPerSource` property to a value less than `64` could impact performance of `Visual Studio` and `nuget.exe` on `Windows`. Maximum degree of parallelism for `Packages.config` style projects is set to `Environment.ProcessorCount` by default. Hence configuring `maxHttpRequestsPerSource` property to a value less than the default could impact performance of `Packages.config` style projects. |
```
nkolev92, line 38: I think we might be adding a bit too much about the techgnical implementation by mentioning service point manager and the likes. 

.NET Core is also a term we probably shouldn't use as much going forward, but really just talk in the terms of tool like you did in the later part. 

Finally, the description is on the richer side. Call out the defaults and just say, lowering the parallelization may worsen performance.

kartheekp-ms: updated text in the latest commit. Please review again and let me know the feedback.

docs/reference/nuget-config-file.md
```diff
@@ -35,6 +35,7 @@ Contains miscellaneous configuration settings, which can be set using the [`nuge
...
+| maxHttpRequestsPerSource | Controls the maximum number of parallel requests sent from NuGet to every package source for package dependency resolution and downloads. The default value on `dotnet.exe` is `Int32.MaxValue` which is derived from `HttpClientHandler.MaxConnectionsPerServer` property. This setting has no impact on `dotnet.exe` for `Mac OS` because the throttling limit is set to `16` to avoid too many open files error. The default value for `NuGet client tools` that runs on `.NET Framework` such as `Visual Studio` and `nuget.exe` is `64` on `Windows` and `1` on `Mono`. The default value for `Packages.config` style projects is set to `Environment.ProcessorCount`. Hence configuring `maxHttpRequestsPerSource` property to a value less than the default could impact NuGet performance. |
```
nkolev92, line 38: You don't need the `Hence`.

nkolev92 (approved):

docs/reference/nuget-config-file.md
kartheekp-ms: Fixed in the latest commit.

PR #2687, opened 2022-03-04
Add scenarios where a PackageDetails leaf is produced

loic-sharma:
Update which scenarios produce a `PackageDetails` catalog leaf as Terrapin team may use NuGet Insight's catalog leaves data.

Please let me know if I forgot any scenarios!

### DIFF
--- a/docs/api/catalog-resource.md
+++ b/docs/api/catalog-resource.md
@@ -207,7 +207,7 @@ following scenarios:
 1. A package is **deprecated**.
 1. A package is **undeprecated**.
 1. A package is **reflowed**.
-1. A package is marked **vulnerable**.
+1. A package's **vulnerability status** is updated.
 
 A package reflow is an administrative gesture that essentially generates a fake push of an existing package with no
 changes to the package itself. On nuget.org, a reflow is used after fixing a bug in one of the background jobs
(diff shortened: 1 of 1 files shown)


docs/api/catalog-resource.md
```diff
@@ -204,7 +204,10 @@ following scenarios:
 1. A package is **pushed**.
 1. A package is **listed**.
```
joelverhagen, line 205: nit: 
```suggestion
1. A package is **relisted**.
```

docs/api/catalog-resource.md
```diff
@@ -204,7 +204,10 @@ following scenarios:
...
 1. A package is **unlisted**.
+1. A package is **deprecated**.
+1. A package is **undeprecated**.
 1. A package is **reflowed**.
+1. A package is marked **vulnerable**.
```
joelverhagen, line 210: I think if the vulnerability information changes a leaf is emitted too. Vulnerabilities can expand, be redacted, and edited in content IIRC. @drewgillies knows the full picture here I think.

loic-sharma: I think a better phrase would be `A package's **vulnerability status** has changed.` but I wanted to mirror the sentence structure of the other bullet points. I think clarity is better than my silly design preferences though so I'll update it :P

loic-sharma: Updated

joelverhagen (approved):

Issue #2674, opened 2022-02-23
Document central package management


jeffkl:
We need documentation for central package management which needs to cover the following

- [x] The whole feature is considered in preview/experimental
- [x] How to enable/disable it (`ManagePackageVersionsCentrally` and `Directory.Packages.props`)
- [ ] How to migrate to it (Retail migration, maybe `Microsoft.Build.CentralPackageVersions` -> NuGet CPM)
- [x] How to override package versions for a single PackageReference (`VersionOverride`)
- [x] How to disable `VersionOverride`
- [x] How to enable/disable overriding/pinning of transitive versions
- [ ] Error `NU1013` (`VersionOverride` is disabled)
- [ ] Warning `NU1504` (`VersionOverride` was ignored)

PR #2692, opened 2022-03-08
Central Package Management Doc

JonDouglas:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2674

This is a work of progress to get an initial set of docs for the various features & getting started experience of central package management. 

Please suggest conceptual sections, changes and commit directly where you see fit so we can make the best doc possible.

### DIFF
--- a/docs/consume-packages/Central-Package-Management.md
+++ b/docs/consume-packages/Central-Package-Management.md
@@ -9,7 +9,7 @@ ms.topic: conceptual
 
 # Central Package Management
 
-Dependency management is a core feature of NuGet. Managing depdendencies for a single project can be easy. Managing dependencies for multi-project solutions can prove to be difficult as they start to scale in size and complexity. In situations where you manage common dependencies for many different projects, you can leverage NuGet's central package management features to do all of this from the ease of a single location.
+Dependency management is a core feature of NuGet. Managing dependencies for a single project can be easy. Managing dependencies for multi-project solutions can prove to be difficult as they start to scale in size and complexity. In situations where you manage common dependencies for many different projects, you can leverage NuGet's central package management features to do all of this from the ease of a single location.
 
 Historically, NuGet package dependencies have been managed in two main locations:
 
--- a/docs/consume-packages/Central-Package-Management.md
+++ b/docs/consume-packages/Central-Package-Management.md
@@ -26,7 +26,7 @@ The feature is available across all NuGet integrated tooling.
 
 Older tooling will ignore Central Package Management configurations and features. To use this feature to the fullest extent, ensure all your build environments use the latest compatible tooling versions.
 
-Central Package Management will apply to all project types – including .NET Framework – as long as compatible tooling is used.
+Central Package Management will apply to all `<PackageReference>` projects – including [legacy .csproj](https://github.com/dotnet/project-system/blob/main/docs/feature-comparison.md) – as long as compatible tooling is used.
 
 ## Enabling Central Package Management
 
@@ -92,9 +92,9 @@ To fully onboard your repository, consider taking these steps:
 
 ## Transitive pinning
 
-You can automatically override a transitive package version even without an explicit top-level `<PackageReference />` by opting into a feature known as transitive pinning. This promotes a transitive dependency to a top-level dependency implicitly on your behalf.
+You can automatically override a transitive package version even without an explicit top-level `<PackageReference />` by opting into a feature known as transitive pinning. This promotes a transitive dependency to a top-level dependency implicitly on your behalf when necessary.
 
-You can enable this feature by setting the MSBuild property `CentralPackageTransitivePinningEnabled` to `true` in a project or `Directory.Build.props` import file:
+You can enable this feature by setting the MSBuild property `CentralPackageTransitivePinningEnabled` to `true` in a project or in a `Directory.Packages.props` or `Directory.Build.props` import file:
 
 ```xml
   <PropertyGroup>
@@ -126,7 +126,7 @@ You can override an individual package version by using the `VersionOverride` pr
 <Project>
 ```
 
-You can disable this feature by setting the MSBuild property `EnablePackageVersionOverride` to `false` in a project or `Directory.Build.props` import file:
+You can disable this feature by setting the MSBuild property `EnablePackageVersionOverride` to `false` in a project or in a `Directory.Packages.props` or `Directory.Build.props` import file:
 
 ```xml
   <PropertyGroup>
--- a/docs/consume-packages/Central-Package-Management.md
+++ b/docs/consume-packages/Central-Package-Management.md
@@ -146,7 +146,7 @@ If you'd like to disable central package management for any reason, you can disa
 
 ## Warning when using multiple package sources
 
-When using central package management, you will see a warning if you have more than one package source defined in your configuration. To resolve this warning, map your package sources with [package source mapping](https://aka.ms/nuget-package-source-mapping) or specify a single package source.
+When using central package management, you will see a `NU1507` warning if you have more than one package source defined in your configuration. To resolve this warning, map your package sources with [package source mapping](https://aka.ms/nuget-package-source-mapping) or specify a single package source.
 
 ```
 There are {0} package sources defined in your configuration. When using central package management, please map your package sources with package source mapping (https://aka.ms/nuget-package-source-mapping) or specify a single package source.
--- a/docs/consume-packages/Central-Package-Management.md
+++ b/docs/consume-packages/Central-Package-Management.md
@@ -144,6 +144,14 @@ If you'd like to disable central package management for any reason, you can disa
 <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>
 ```
 
+## Warning when using multiple package sources
+
+When using central package management, you will see a warning if you have more than one package source defined in your configuration. To resolve this warning, map your package sources with [package source mapping](https://aka.ms/nuget-package-source-mapping) or specify a single package source.
+
+```
+There are {0} package sources defined in your configuration. When using central package management, please map your package sources with package source mapping (https://aka.ms/nuget-package-source-mapping) or specify a single package source.
+```
+
 > [!Note]
 > This feature is in active development. We appreciate you trying it out and providing any feedback you may have at [NuGet/Home](https://github.com/nuget/home/issues).
 >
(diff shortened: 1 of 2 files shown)


docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+description: Manage your dependencies in a central location and how you can get started with central package management.
+author: jondouglas
+ms.author: jodou
+ms.date: 2/25/2022
+ms.topic: conceptual
+---
+
+# Central Package Management
+
+Dependency management is a core feature of NuGet. Managing depdendencies for a single project can be easy. Managing dependencies for multi-project solutions can prove to be difficult as they start to scale in size and complexity. In situations where you manage common dependencies for many different projects, you can leverage NuGet's central package management features to do all of this from the ease of a single location.
```
nkolev92, line 12:
```suggestion
Dependency management is a core feature of NuGet. Managing dependencies for a single project can be easy. Managing dependencies for multi-project solutions can prove to be difficult as they start to scale in size and complexity. In situations where you manage common dependencies for many different projects, you can leverage NuGet's central package management features to do all of this from the ease of a single location.
```

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+
+Historically, NuGet package dependencies have been managed in two main locations:
+
+- `packages.config` - An XML file used in older project types to maintain the list of packages referenced by the project.
+- `<PackageReference />` - An XML element used in new project types that manages NuGet dependencies directly within project files.
+
+Starting with [NuGet 7.0](..\release-notes\NuGet-7.0.md), you can centrally manage your dependencies in your projects with the addition of a `Directory.Packages.props` file.
```
nkolev92, line 19: versioning; It's not really going to be nuget 7.0, but NuGet 6.4 that matches .NET 7.

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+
+The feature is available across all NuGet integrated tooling.
+
+* [Visual Studio 2022 and later](https://visualstudio.microsoft.com/downloads/)
```
nkolev92, line 23: Update 4 probably.

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+* [.NET SDK 7.0.000 and later](https://dotnet.microsoft.com/download/dotnet/7.0)
+* [nuget.exe 7.0.0 and later](https://www.nuget.org/downloads)
```
nkolev92, line 25: 6.4

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+
+Older tooling will ignore Central Package Management configurations and features. To use this feature to the fullest extent, ensure all your build environments use the latest compatible tooling versions.
+
+Central Package Management will apply to all project types – including .NET Framework – as long as compatible tooling is used.
```
nkolev92, line 29: I'd probably say PackageReference projects - including old style csproj - while linking to the csproj.

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+    <TargetFramework>net6.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" />
+  </ItemGroup>
+</Project>
+```
+
+Now you're using central package management and managing your versions in a central location!
```
nkolev92, line 60: I think it'd be great if we can add a more detailed migration game, similar to what we did for source mapping. 

The rough steps would be: 

Run dotnet list packages or the equivalents. 
Start adding all top level dependencies in the Directory.Packages.props
Some guidance for resolving conflicts, including VersionOverride etc.
Remove *all* versions from PackageReference.

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+- Project1 will evaluate the `Directory.Packages.props` file in the `Repository\Solution1\` directory.
+- Project2 will evaluate the `Directory.Packages.props` file in the `Repository\` directory.
+
+## Get started
+
+To fully onboard your repository, consider taking these steps:
+
+1. Create a new file at the root of your repository named `Directory.Packages.props` that declares your centrally defined package versions in.
```
nkolev92, line 87: ignore my comment above :D The recommendation does apply though. 
Maybe provide some guidance about resolving conflicts, and advertise VersionOverride as a solution, but first encourage using the same version.

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+2. Declare `<PackageVersion />` items in your `Directory.Packages.props`.
+3. Declare `<PackageReference />` items without `Version` attributes in your project files.
+
+<!--For an idea of how central package management may look like, refer to our [samples repo](https://github.com/NuGet/Samples/tree/main/CentralPackageManagementExample).-->
+
+## Transitive pinning
+
+You can automatically override a transitive package version even without an explicit top-level `<PackageReference />` by opting into a feature known as transitive pinning. This promotes a transitive dependency to a top-level dependency implicitly on your behalf.
```
nkolev92, line 95: > This promotes a transitive dependency to a top-level dependency implicitly on your behalf

This is likely to raise a lot questions. 
Maybe clarify it further that only where the transitive dependency exists, it'll be effectively act as being promoted top level. 

I don't have a great recommendation here. 
I'll comment again if I tihnk of something.

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,150 @@
...
+
+You can enable this feature by setting the MSBuild property `CentralPackageTransitivePinningEnabled` to `true` in a project or `Directory.Build.props` import file:
```
nkolev92, line 97: Any reason why we shouldn't include Directory.Packages.props as an option?
In fact, I'd say it's maybe even the more appropriate place.

nkolev92, line 97: Any reason why we shouldn't include Directory.Packages.props as an option?
In fact, I'd say it's maybe even the more appropriate place.

docs/consume-packages/Central-Package-Management.md
nkolev92: Given that we're adding the functionality in 6.2, any reason why we shouldn't announce the functionality in 6.2?

nkolev92 (commented):
This looks great! Well done. 
Some suggestions for additions.

docs/consume-packages/Central-Package-Management.md
JonDouglas: Just placeholders. Will confirm with Jeff

docs/consume-packages/Central-Package-Management.md
JonDouglas: Good idea! Will fix.

docs/consume-packages/Central-Package-Management.md
JonDouglas: Good ideas, will add a couple more steps!

docs/consume-packages/Central-Package-Management.md
JonDouglas: Yeah, I tried to simplify what this means. I think an example could help, but also maybe we just wait until there's confusion on it and fix it then.

docs/consume-packages/Central-Package-Management.md
JonDouglas: Should we just include all three? I'm totally fine with adding that.

docs/consume-packages/Central-Package-Management.md
nkolev92: Does calling it a pseudo top-level dependency make it better? 
That it clearly indicates there's more to it and it's an `always like this` thing.

docs/consume-packages/Central-Package-Management.md
nkolev92: Yeah, I think that would be nice.

jeffkl (commented):
Looks good, let's just nail down the release versions.  I imagine we'll also be updating this doc over time as people ask questions that it doesn't answer.

docs/consume-packages/Central-Package-Management.md
nkolev92: Or `like` is just fine. 
The pseudo meaning in this case is the obvious one, `a pretend top-level dependency`.

JonDouglas: My main worry is that it's hard enough for devs to distinguish top level & transitive today. Adding the word `pseudo` or `pretend` might confuse that further?

nkolev92: Fair, but misleading them could have a similar effect no? 

Maybe we change it to: 

> This promotes a transitive dependency to a top-level dependency implicitly on your behalf **when necessary**.

**Caps** is the addition.

JonDouglas: yeah let's do that.

JonDouglas:
@jeffkl Yep that's right.

docs/consume-packages/Central-Package-Management.md
JonDouglas: Commit will come to fix :)

JonDouglas:
Putting this up ready for review so we can publish it alongside VS 17.2 P3 if wanted :)

nkolev92:
There's a few warnings raised by the build, in particular the link to the release notes, is that something we want to fix? Something comfortable leaving in for now?

docs/consume-packages/Central-Package-Management.md
```diff
@@ -0,0 +1,158 @@
+---
+title: Central Package Management
+description: Manage your dependencies in a central location and how you can get started with central package management.
+author: jondouglas
+ms.author: jodou
+ms.date: 2/25/2022
+ms.topic: conceptual
+---
+
+# Central Package Management
```
nkolev92, line 10: Awesome summary.

docs/consume-packages/Central-Package-Management.md
nkolev92: Did you mean to add more steps here? 

We can always improve the doc later. Maybe we can enhance during docs days, as long as we have an issue to know what we need to do that.

nkolev92 (approved):

JonDouglas:
Q: NuGet 6.2 won't likely go out for awhile, should we remove that for now or leave to 404 for now?

Issue #2691, opened 2022-03-08
Please clarify syntax of the packageSourceMapping section


gpwen:
Please clarify syntax of the packageSourceMapping section. It's not obvious how I should associate multiple package patterns with the same package source. Should I specify multiple "packageSource" items with the same key? Or should I specify a complex pattern as combination of multiple simple patterns? Is that even allowed? Or should I specify multiple "package" items?

In the end, it appears that:
* A "packageSource" key can only be specified **once**; and
* Each "packageSource" item allow **multiple** "package" items; and
* Each package pattern can only be a simple ID or wildcards, **combination won't  work**.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 15617693-33df-025b-26c3-2d80e4df68d3
* Version Independent ID: 15617693-33df-025b-26c3-2d80e4df68d3
* Content: [Package Source Mapping](https://docs.microsoft.com/en-us/nuget/consume-packages/package-source-mapping)
* Content Source: [docs/consume-packages/Package-Source-Mapping.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/Package-Source-Mapping.md)
* Product: **nuget**
* GitHub Login: @nkolev92
* Microsoft Alias: **nikolev**

nkolev92:
Thanks for trying out the feature @gpwen 

I'll see if I can tweak the docs to improve readability for the first 2 points. 
Key usually implies one occurence.

The 3rd part is already covered in https://docs.microsoft.com/en-us/nuget/consume-packages/package-source-mapping#package-pattern-syntax

gpwen:
Thanks for the prompt update @nkolev92 

It might be a little bit picky, but "Example Syntax" in "Package Pattern Syntax" section listed something like "NuGet.Common, Contoso.Contracts". That does leave an impression that package pattern might accept comma-delimited patterns, especially for those searching for ways to specify multiple patterns for one source. How about something like: "a", "b" or "c". If someone does try that as XML attribute value, it will be a syntax error, so it leaves no room for second guess.

I also think it would be very helpful if the sample code in "Enabling Package Source Mapping" includes a case where one source is associated with multiple patterns.

nkolev92:
@gpwen I'm not sure I agree about the feedback on the package prefix table. The fact that those patterns are separate is indicated by the quotes. 

[...]

PR #2693, opened 2022-03-09
Clarify the source mapping syntax

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2691

### DIFF
--- a/docs/consume-packages/Package-Source-Mapping.md
+++ b/docs/consume-packages/Package-Source-Mapping.md
@@ -32,9 +32,10 @@ For a video-based overview of the Package Source Mapping feature, consider watch
 
 To opt into this feature, you must have a `nuget.config` file. Having a single `nuget.config` at the root of your repository is considered a best practice. See [nuget.config documentation](../reference/nuget-config-file.md) to learn more.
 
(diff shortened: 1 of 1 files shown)


docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -33,6 +33,8 @@ For a video-based overview of the Package Source Mapping feature, consider watch
 To opt into this feature, you must have a `nuget.config` file. Having a single `nuget.config` at the root of your repository is considered a best practice. See [nuget.config documentation](../reference/nuget-config-file.md) to learn more.
 
 Declare your desired package sources in your `nuget.config` file. Following your source declarations, add a `<packageSourceMapping>` element that specifies the desired mappings for each source.
```
JonDouglas, line 35: What if we made these bullet items? I think it would be easier and more readable now that there are a few steps here.

JonDouglas (approved):
Great additions, see comment for consideration.

docs/consume-packages/Package-Source-Mapping.md
nkolev92: Good idea. 

Let me go through it.

nkolev92: The bullet items do look better. 

Pushed the change.

Issue #2683, opened 2022-03-02
Conflicting info on PrivateAssets default value


michaelw-au:
This page states:
_Note that the default private assets when none is supplied is build;analyzers. Therefore, you should use PrivateAssets="compile;build;analyzers" in ..._

which seems to conflict with https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets
which states the PrivateAssets default value (in the table) is
contentfiles;analyzers;build

which is the true default?
Or am I mis-understanding different contexts here?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6efe48ec-f0f0-de4b-1c07-af1cc5c87d94
* Version Independent ID: fec47e16-9c71-8b73-0a41-d6808ac14700
* Content: [Select Assemblies Referenced By Projects](https://docs.microsoft.com/en-us/nuget/create-packages/select-assemblies-referenced-by-projects)
* Content Source: [docs/create-packages/Select-assemblies-referenced-by-projects.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/create-packages/Select-assemblies-referenced-by-projects.md)
* Product: **nuget**
* GitHub Login: @zivkan
* Microsoft Alias: **zivkan**

nkolev92:
The confusion probably came from the fact that contentFiles are not included in the nuspec.

nkolev92:
PR out https://github.com/NuGet/docs.microsoft.com-nuget/pull/2710

PR #2710, opened 2022-04-01
Add content files where appropriate in select assemblies to be referenced by projects

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2683


JonDouglas (approved):
Thank you for the contribution! ;P

PR #2733, opened 2022-04-22
Add note on floating version resolution

aortiz-msft:
Clarify the resolution behavior when global package folder is not empty.


JonDouglas (approved):

PR #2738, opened 2022-04-28
update metadat from ms.prod to ms.service

huypub:


JonDouglas:
Hi,

Can we get more context as to this PR? An empty description isn't enough for us to make a decision to merge. Such as docs pointing to ms.prod vs. ms.service, and whether we even need this metadata would help! Thanks!

nkolev92:
Similar to the other one, closing per Jon's comment. 
Feel free to reopen if necessary.

PR #2741, opened 2022-04-29
Call out tooling preview state

SeanFeldman:
Currently, the document reads as if the feature is available with the stable VS/.NET.
But it's not, it requires preview bits.


nkolev92:
Apologies for the delay @SeanFeldman 

The stable versions have now shipped so we don't need this update anymore. 

Thank you for the contribution.

PR #2800, opened 2022-07-04
document The parameter is incorrect behavior

SimonCropp:
given this error message can take hours to diagnose the root cause


JonDouglas (approved):
Thanks for the contribution, Simon!

Issue #2804, opened 2022-07-13
The Create Package with NuGet.exe document describes concepts such as build targets, but that information is more generic


nkolev92:
Concepts such as the lib folder,  build folders etc should be introduced on a more general level. 
They are *not* specific to NuGet.exe pack, and they should be introduced as such.

Note that these things are sometimes generic and important in the context of authoring/consuming packages, so they simply cannot be pigeon holed into one or another section.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f35f4267-e5e3-ddcf-d3e4-479fb233ce9f
* Version Independent ID: 6d49d66f-f777-7e44-3a45-723e3ad09a59
* Content: [Create a NuGet package using nuget.exe CLI](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package)
* Content Source: [docs/create-packages/Creating-a-Package.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/create-packages/Creating-a-Package.md)
* Product: **nuget**
* GitHub Login: @JonDouglas
* Microsoft Alias: **feaguila**

PR #2806, opened 2022-07-13
Add guidances that props/targets should not modify NUGet specific targets/props

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2805

This ideally has its own page, but unfortunately our docs are short on the conceptual side, so I'm working with what I can. 

I have created this isuse that talks about the lack of `concepts` fro pack/restore.  https://github.com/NuGet/docs.microsoft.com-nuget/issues/2804


nkolev92:
Going to close this in favor of https://github.com/NuGet/docs.microsoft.com-nuget/pull/2807, which has a more comprehensive refactoring.

PR #2807, opened 2022-07-14
Add a dedicated page for props and targets

nkolev92:
Very often, our documentation is separated in create/consume packages.

Some concepts are so tightly coupled that separating them in create/consume only folder will never paint the full picture for the reader.

Furthermore it doesn't give us the ability to provider advice about general content, because usually you can pack in 3 different ways. 

Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2805 
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2804


nkolev92:
@aortiz-msft 

These docs are helpful in the context of https://developercommunity.visualstudio.com/t/Package-Manager-does-not-stop-restoring/10043518.

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+
+As NuGet has evolved, various different folders for build props and targets have been added.
+
+| Folder | NuGet Version | Use |
+|--------|---------------|-----|
+| build | 2.5+ | Build logic for every framework of a project. |
+| buildMultiTargetting | 4.0+ | Build logic for the `outer build` for projects that target multiple frameworks. PackageReference only. |
+| buildTransitive | 5.0+ | Build logic for assets that flow transitively to any consuming project. See the [feature](https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior) page. PackageReference only. |
+
+## Framework specific build folder
```
zivkan, line 25: I don't see anything on the page saying that `build/<packageid>.props` is also possible. I was looking for it, because I think it's important to warn package authors that this makes their package compatible with all projects, so there might be scenarios (like an otherwise .NET package being installed in a C++ project) where the props file will be imported, but not any `lib/<tfm>/*` assets.  At least, that's my understanding of how package compatibility works.

Basically, I want to discourage package authors from using build files without a TFM, but I guess that not documenting it possibly a strong discouragement too 😁

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+# MSBuild props and targets in a package
+
+In additional to the more traditional assemblies, NuGet packages may sometimes add custom build targets or properties to projects that consume that package.
+This can be achieved by adding a valid MSBuild file, in the form `<package_id>.targets` or `<package_id>.props` (such as `Contoso.Utility.UsefulStuff.targets`) within the build folders of the project.
+
+## Build folders
+
+As NuGet has evolved, various different folders for build props and targets have been added.
+
+| Folder | NuGet Version | Use |
```
JonDouglas, line 19: Great table, how does one differentiate when to use each of these? Should we talk about that outside the use?

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+MSBuild `.props` and `.targets` files for multi-framework targeting can be placed in the `\buildMultiTargeting` folder.
+When the imports are generated, a condition that the MSBuild property `$(TargetFramework)` is empty is set.
+
+### packages.config projects
+
+When NuGet installs a package with `\build` files, it adds MSBuild `<Import>` elements in the project file pointing to the `.targets` and `.props` files. (`.props` is added at the top of the project file; `.targets` is added at the bottom.) A separate conditional MSBuild `<Import>` element is added for each target framework.
+
+## Authoring packages with MSBuild props and targets
+
+You can use your tool of choice to include MSBuild props and targets your package.
```
JonDouglas, line 60:
```suggestion
You can author MSBuild props and targets for your package with the following tools:

```

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+
+- [NuGet.exe pack](..\create-packages\Creating-a-Package.md#include-msbuild-props-and-targets-in-a-package)
+- [dotnet.exe pack](..\create-packages\creating-a-package-dotnet-cli.md)
+- [MSBuild.exe pack](..\create-packages\creating-a-package-msbuild.md)
+
+
+### Guidance for the content of MSBuild props and targets
+
+The tooling itself cannot provide explicit guidance for how to author these props and targets as they will vary based on the need of the package author and the target projects themselves.
```
JonDouglas, line 69:
```suggestion
NuGet does not limit how you author prop and targets as they will vary based on the need of the package author and the target projects themselves.

```

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+
+There are a few things that must not be done in packages' props and targets, such not specifying properties and items that affect restore, as those will be automatically excluded.
+
+- Some examples of properties that must not be added or updated: TargetFramework, TargetFrameworkMoniker, TargetPlatforMoniker, AssetTargetFallback etc.
```
JonDouglas, line 73:
```suggestion
- Some examples of properties that must not be added or updated: TargetFramework, TargetFrameworkMoniker, TargetPlatformMoniker, AssetTargetFallback etc.

```

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+- Some examples of items that must not be added or updated: PackageReference, PackageVersion, PackageDownload etc.
```
JonDouglas, line 74:
```suggestion
- Some examples of items that must not be added or updated: PackageReference, PackageVersion, PackageDownload, etc.

```

JonDouglas (approved):
Great additions / refactoring. Added some feedback to consider!

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+    \build
+        \netstandard1.4
+            \Contoso.Utility.UsefulStuff.props
+            \Contoso.Utility.UsefulStuff.targets
+        \net462
+            \Contoso.Utility.UsefulStuff.props
+            \Contoso.Utility.UsefulStuff.targets
+```
+
+Note that if a package does not have any files in the `lib` or `ref` folders and only files under a framework specific build folder, that package will be considered compatible with all files. Up to date versions of the pack tooling, raise the  [NU5127](..\reference\errors-and-warnings\NU5127.md) warning when such packages are created.
```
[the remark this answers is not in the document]
erdembayar: Here does `all projects` means regarding PR vs PC or targetframework term?

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,74 @@
...
+description: Describes MSBuild props and targets in NuGet packages
+author: nkolev92
+ms.author: nikolev
+ms.date: 07/13/2022
+ms.topic: conceptual
+---
+
+# MSBuild props and targets in a package
+
+In additional to the more traditional assemblies, NuGet packages may sometimes add custom build targets or properties to projects that consume that package.
```
dominoFire, line 12: I suggest using .props and .targets instead of props and targets, respectively, since this is how it appears to be in MSBuild docs: https://docs.microsoft.com/visualstudio/msbuild/customize-your-build?view=vs-2022#choose-between-adding-properties-to-a-props-or-targets-file

Example below.

```suggestion
In additional to the more traditional assemblies, NuGet packages can add custom MSBuild `.targets` files or `.props` files to projects consuming the package.
```

docs/concepts/MSBuild-props-and-targets.md
nkolev92: I kind of assumed the `use` explains that.

docs/concepts/MSBuild-props-and-targets.md
nkolev92: I think I meant to use you can use the following tools to authot your packages, not the props and targets explicitly.

docs/concepts/MSBuild-props-and-targets.md
nkolev92: Fixed. all through the doc.

docs/concepts/MSBuild-props-and-targets.md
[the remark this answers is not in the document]
nkolev92: all NuGet compatible projects.

docs/concepts/MSBuild-props-and-targets.md
nkolev92: 2nd sentence says that. 
I've added some more clarifications.

docs/concepts/MSBuild-props-and-targets.md
JonDouglas: Like from the developer perspective which is most common? Which one should I use in my situation? I can piece together the answers but maybe we should have another column of common cases?

nkolev92: Do you have a suggestion?

I kind of see those 2 as equivalent to the current descriptions.

zivkan (approved):

docs/concepts/MSBuild-props-and-targets.md
```diff
@@ -0,0 +1,78 @@
...
+            \Contoso.Utility.UsefulStuff.props
+            \Contoso.Utility.UsefulStuff.targets
+        \net462
+            \Contoso.Utility.UsefulStuff.props
+            \Contoso.Utility.UsefulStuff.targets
+```
+
+Prefer using framework-specific build folders whenever appropriate to avoid false positive installations in projects that may not be supported by your package.
+
+Note that if a package does not have any files in the `lib` or `ref` folders and only files under a framework specific build folder, that package will be considered compatible with all projects. Up to date versions of the pack tooling, raise the  [NU5127](..\reference\errors-and-warnings\NU5127.md) warning when such packages are created.
```
erdembayar, line 45: What does it mean by `compatible with all projects`?

PR #2815, opened 2022-07-22

Sherlock01:


docs/consume-packages/Central-Package-Management.md
```diff
@@ -94,7 +94,7 @@ Repository
   <Project>
     <Import Project="..\Directory.Packages.props">
     <ItemGroup>
-      <PackageVersion Update="Newtonsoft.Json" Version="12.0.1" />
+      <PackageVersion Include="Newtonsoft.Json" Version="12.0.1" />
```
zivkan, line 97: This section of the docs is about having more than one `Directory.Packages.props` in the repo, so that one subdirectory uses a different version to other projects in the repo. Using `Update` is correct, to avoid MSBuild having two different items for the same package, and then command line de-duplication working differently to Visual Studio's project system de-duplication, and therefore the project behaving differently in VS vs command line.  Recent versions of NuGet will warn about duplicate items to minimize the risk.

zivkan (changes requested):

docs/consume-packages/Central-Package-Management.md
Sherlock01: Got it. Thank you for the response and explanation

PR #2829, opened 2022-07-28
Update NU1100.md

waldyd:
NU1100 may also be caused by the lack of a NuGet configuration file

### DIFF
--- a/docs/reference/errors-and-warnings/NU1100.md
+++ b/docs/reference/errors-and-warnings/NU1100.md
@@ -17,10 +17,15 @@ f1_keywords:
 <pre>Unable to resolve 'Dependency (>= 1.0.0)' for 'TargetFramework'</pre>
 
 ### Issue
-A dependency could not be resolved. This is a generic issue for types that are not packages or projects.
+A dependency could not be resolved. This is a generic issue for types that are not packages or projects. 
+
+This issue may also be caused by the lack of a NuGet configuration file.
 
 ### Solution
 Open the project file and examine the list of its dependencies. Check that each dependency exists on the package sources you're using, and that the package supports the project's target framework.
+
+In the case of a NuGet configuration file not being created, issue the following command from the root folder of the project:  
+`dotnet new nugetconfig`
 
 ## Scenario 2
 


JonDouglas (approved):
Thank you kindly for the contribution!

docs/reference/errors-and-warnings/NU1100.md
```diff
@@ -17,11 +17,16 @@ f1_keywords:
 <pre>Unable to resolve 'Dependency (>= 1.0.0)' for 'TargetFramework'</pre>
 
 ### Issue
-A dependency could not be resolved. This is a generic issue for types that are not packages or projects.
+A dependency could not be resolved. This is a generic issue for types that are not packages or projects. 
+
+This issue may also be caused by the lack of a NuGet configuration file.
```
nkolev92, line 22: NuGet would normally be creating the user-wide configuration file if it doesn't already exist. 

Can you elaborate how you hit this?

waldyd:
The issue can be reproduced by:

1. Installing .NET Core 
2. Installing VS Code
3. dotnet new mvc
4. dotnet build. 

On systems without NuGet (or tools that install it, i.e. Visual Studio). .NET app building will trigger this issue. 

Thanks.

nkolev92:
@JonDouglas I think we merged prematurely here. 

I think we need to confirm this repro first, as it doesn't really make sense to me immediately.

JonDouglas:
@nkolev92 A simple reproduction is to remove the contents of a nuget.config file as if it were "empty". i.e. leave only the xml & configuration tags. 

I repro'd it and the contribution here is helpful as the first scenario asks to check if dependencies exist on package sources, but the reality is the config may be generated by other tooling causing issues and one may want to check the config for accuracy.

This is similar to the problem of other nuget-powered tooling generating an empty config and the user has no idea how to resolve it.

nkolev92:
> A simple reproduction is to remove the contents of a nuget.config file as if it were "empty". i.e. leave only the xml & configuration tags.

That's not what the update says. It talks about a missing nuget.config, which is different.

JonDouglas:
@nkolev92 The assumption here is that there exists no config in the project/solution and the error is thrown because the user-level is problematic (i.e. empty from other tooling or similar).

using the dotnet new nugetconfig generates a proper configuration at the project level that will override the user-level and thus works.

Perhaps the language we want to instill here is fixing these items at the user-level or inspecting the various configs to ensure the sources exist in the first place?

I still think the contribution is fine as is, but we can improve more later.

nkolev92:
I don't think we can say `missing config` because a missing config altogether causes NuGet to create one.
It is a critical onboarding experience.

The context with which it gets created is irrelevant imo. 
The docs shouldn't contradict the actual tool behavior even if it's not the most user friendly description. 

> Perhaps the language we want to instill here is fixing these items at the user-level or inspecting the various configs to ensure the sources exist in the first place?

I'll take a crack at updating this and create a PR. I think it'll be easier to make progress that way :D

JonDouglas:
@waldyd As one last question, can you let us know the contents of your `user` config?

https://docs.microsoft.com/nuget/consume-packages/configuring-nuget-behavior#config-file-locations-and-uses

nkolev92:
Hoping the following version improves clarity. 

https://github.com/NuGet/docs.microsoft.com-nuget/pull/2838.

waldyd:
I did not have a nuget.config file at a user level neither at global level. When the issue arose I had to create it at project level by issuing dotnet new nugetconfig.

nkolev92:
NuGet creates a settings file if there's none. 
That's the onboarding experience when you install new tools. There's no other component that creates a config file by default other than NuGet. 

cc @v-luzh

Can you please try to repro this?

Start off with a new windows machine and follow the steps here:

https://github.com/NuGet/docs.microsoft.com-nuget/pull/2829#issuecomment-1203440980

PR #2831, opened 2022-07-29
Updates to ms.date values based on file history

Mikejo5000:
Updates for ms.date. Updating the value if the last update appeared to be a non-trivial update, and if the last update was <1 year ago (otherwise, it's always considered "stale"). I'm reviewing only the articles of >1000 PVs (June), which looks like 46 articles.


Mikejo5000:
@JonDouglas Ready to go. See notes in the Description, above.

nkolev92 (approved):

JonDouglas (approved):
Thanks for the contribution Mike!

PR #2843, opened 2022-08-07
Add MSBuildSdk as well known package type

kant2002:


kant2002:
I assume https://review.docs.microsoft.com/en-us/help/onboard/faq link is for MS only. Maybe if there nothing special, would be good to have that link public.

docs/create-packages/set-package-type.md
```diff
@@ -19,6 +19,8 @@ Packages can be marked with one more more *package types* to indicate its intend
 
 - `Template` type packages provide [custom templates](/dotnet/core/tools/custom-templates) that can be used to create files or projects like an app, service, tool, or class library.
 
+- `MSBuildSdk` type packages provides [MSBuild SDK](/visualstudio/msbuild/how-to-use-project-sdk) that can be used to augment behaviour of projects. Examples of such SDK can be found on [following GitHub repo](https://github.com/microsoft/MSBuildSdks).
```
nkolev92, line 22: I'm not 100% sure how I feel about this change, as while it's commonly understood what the MSBuildSDK package is, https://github.com/NuGet/Home/issues/6484 is still not resolved, so if you `ask the client`, this isn't really known. 

It's probably cheap to fix, but I just haven't had the time.

@jeffkl thoughts?

JonDouglas: quick bump for @jeffkl so we can revise or merge.

jeffkl: Its probably fine to document it and implement it later honestly.  There's no enforcement as far as I know so you can put any value in there you want.  At least if we document it then packages could be correct.

nkolev92: I'm comfortable with whatever you 2 decide. Don't block merging/changes on me :)

JonDouglas:
Closing this PR in favor of https://github.com/NuGet/docs.microsoft.com-nuget/pull/3036 (simplified language)

Thank you for kicking off this contribution. We really appreciate it!

PR #2844, opened 2022-08-09
Mention about NuGet.PackageSourcemapper tool for package source mapping feature

erdembayar:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2779

To create visibility for [NuGet.PackageSourceMapper tool](https://www.nuget.org/packages/NuGet.PackageSourceMapper/0.1.3-preview.22218.1#readme-body-tab) mention it in doc. It'll drive customers onboarding to reducing time they spend on creating source mapping patterns.

PR #2846, opened 2022-08-09
added dotnet restore step and clear global packages folder step while onboarding onto source mapping feature

kartheekp-ms:
While reviewing https://github.com/NuGet/docs.microsoft.com-nuget/pull/2844, I noticed that the onboarding steps can be improved for source mapping feature. The reason is that `dotnet list package` requires restore to be executed prior to produce any results which was missing in the original docs. Also added a step to clean global packages folder before running the restore again so that customers can verify that the packages are restored from expected sources based on the source mapping configuration.

### DIFF
--- a/docs/consume-packages/Package-Source-Mapping.md
+++ b/docs/consume-packages/Package-Source-Mapping.md
@@ -103,13 +103,13 @@ To learn more about how package installation works, see [the conceptual document
 To fully onboard your repository you may take the following steps:
 
 1. Declare a new [global packages folder for your repo](../reference/nuget-config-file.md#config-section).
-2. Run [dotnet restore](/dotnet/core/tools/dotnet-restore) to restore dependencies.
-3. Run [`dotnet list package --include-transitive`](/dotnet/core/tools/dotnet-list-package#synopsis) to view all top-level and transitive packages in your solution.
+1. Run [dotnet restore](/dotnet/core/tools/dotnet-restore) to restore dependencies.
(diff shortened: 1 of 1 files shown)


docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -103,11 +103,13 @@ To learn more about how package installation works, see [the conceptual document
 To fully onboard your repository you may take the following steps:
 
 1. Declare a new [global packages folder for your repo](../reference/nuget-config-file.md#config-section).
-1. Run [`dotnet list package --include-transitive`](/dotnet/core/tools/dotnet-list-package#synopsis) to view all top-level and transitive packages in your solution.
+2. Run [dotnet restore](/dotnet/core/tools/dotnet-restore) to restore dependencies.
+3. Run [`dotnet list package --include-transitive`](/dotnet/core/tools/dotnet-list-package#synopsis) to view all top-level and transitive packages in your solution.
```
nkolev92, line 107: Thanks for adding these steps. 

Can you please revert the numbering changes?

kartheekp-ms: These numbers are added by GitHub when I clicked on `Edit this document` on docs.microsoft.com. I am happy to revert those changes.

kartheekp-ms: Reverted number changes in the document. Now the changes proposed in this PR can be noticed easily when compared to earlier. Thanks Nikolche.

nkolev92: Using the `1.` numbers allows easier reordering as well, because when you use `1.` it autogens the numbers.

nkolev92 (approved):

PR #2852, opened 2022-08-16
Update and refresh 4 articles: Install and use packages

v-thepet:
User Story https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1973816.


v-thepet:
@Mikejo5000 - These three articles about installing and managing packages are ready for your review and signoff when approved. I wasn't sure who all is supposed to review and approve, so could you pull in Josee and/or Jon Douglas to review if needed? Some notes:

- [docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md](https://github.com/v-thepet/docs.microsoft.com-nuget/blob/nuget1/docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md). The code has changed due to the new .NET 6.0 template. The namespace needs to be explicitly declared, so to that end, I provided an explicit name for the project. (VS also asked for the Name and Email to be nullable, but this might not actually be necessary?)

- [docs/consume-packages/install-use-packages-powershell.md](https://github.com/v-thepet/docs.microsoft.com-nuget/blob/nuget1/docs/consume-packages/install-use-packages-powershell.md). The existing screenshots are accurate, but the rounded-corner red boxes aren't Docs standard (they should be square corners). I don't think it's a blocker or necessary to redo the screenshots for this, and I notice that all the red boxes in this content set have rounded corners, so maybe it's best to be consistent. But I'd be glad to redo the screenshots to make the boxes square, so let me know.

  I'm unable to replicate the added console commands under **Extend the Package Manager Console**. The MvcScaffolding project (even the updated one for VS2019) targets .NET 4.6-4.7, so it won't install in a .NET 6.0 project. I spent awhile looking for some .NET 6 packages that would put some extra commands in the console, but couldn't find any. So for now, I just left the current content verbatim. Let me know if you know of a package that would work for .NET 6.0, or if I should remove this section or screenshot, or perhaps add a warning that it only works for .NET 4.6 projects.

Thanks for your review.

JonDouglas:
Thank you kindly for the contribution! Give me some time to add some comments.

v-thepet:
@JonDouglas @Mikejo5000 - I realized there's another install/use article on my list: https://docs.microsoft.com/en-us/nuget/consume-packages/install-use-packages-nuget-cli. So if I get to it before you review this, I'll add it to this PR. It shouldn't affect the articles I already edited. If you already completed your review, I'll put that article in a new PR. Thanks for your review.

v-thepet:
@Mikejo5000 @JonDouglas @j-martens - I've added the fourth "manage" article, [docs/consume-packages/install-use-packages-nuget-cli.md](https://github.com/v-thepet/docs.microsoft.com-nuget/blob/nuget1/docs/consume-packages/install-use-packages-nuget-cli.md) and its include, to this PR. 

I have 5 remaining articles to refreshfor this User Story by Aug. 31, 2022. I will have the drafts done by that time, but the PRs might not be done if there are extensive review revisions. Any reviews at this point will be very helpful for completing the remaining drafts and finishing the project as soon as possible. Thanks!

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,120 @@
...
+```cli
+nuget install packages.config
+```
 
-Alternatively, you can use the following command to install a NuGet package using an existing `packages.config` file to the *packages* folder. This does not add the package to your project dependencies, but installs it locally.
+For example, the following command installs all the packages listed in the *packages.config* file at *C:\myPath* to the *packages* subdirectory in the current directory:
 
 ```cli
-nuget install packages.config -OutputDirectory packages
+nuget install C:\myPath\packages.config -OutputDirectory packages
```
Mikejo5000, line 60: @v-thepet Just to confirm, looks like the full path is required for packages.config? @nkolev92 in case any clarification is needed.

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,120 @@
...
 ```cli
 nuget list -Source <source>
 ```
 
-For example, list packages in the *packages* folder.
+For example, to list packages in the *packages* subdirectory, use:
 
 ```cli
-nuget list -Source C:\Users\username\source\repos\MyProject\packages
+nuget list -Source packages
```
Mikejo5000, line 88: Opposite question to the previous. Testing showed no path is required here?

docs/consume-packages/install-use-packages-nuget-cli.md
v-thepet: @Mikejo5000 - I'll experiment with using a relative path.

docs/consume-packages/install-use-packages-nuget-cli.md
v-thepet: @Mikejo5000 - Right. I think this was an example path. (VS projects don't have to be at source/repos.) Using just "packages" lists the packages in a subfolder named packages in the current folder. The folder name "packages" seems arbitrary, at least, any folder name works with the command. I don't know if folder name "packages" is necessary for some other functionality. If you use just `nuget list` you get all the packages at nuget.org (or default location).

docs/consume-packages/includes/restore-nuget-exe-cli.md
```diff
@@ -1,13 +1,21 @@
...
 
-`restore` only adds packages to disk but does not change a project's dependencies. To restore project dependencies, modify `packages.config`, then use the `restore` command.
+To restore packages, run the following command:
 
-As with the other `nuget.exe` CLI commands, first open a command line and switch to the directory that contains your project file.
+```cli
+nuget restore <projectPath>
+```
+
+The `restore` command uses a solution file or a *package.config* file in the specified project path.
```
Mikejo5000, line 11: Also tested this change? @nkolev92 to verify

Mikejo5000 (approved):
Looks good, thanks! I left a couple comments.

docs/consume-packages/includes/restore-nuget-exe-cli.md
v-thepet: @Mikejo5000 - Yes, I tested, but I'm not completely clear whether/when it looks for a .sln file first, as opposed to a package.config file. I can test further.

nkolev92: It's solution first.

docs/consume-packages/install-use-packages-nuget-cli.md
nkolev92: Relative paths should work too.

v-thepet: @Mikejo5000 @nkolev92 - Correct, relative paths work.

docs/consume-packages/install-use-packages-nuget-cli.md
v-thepet: @Mikejo5000 @nkolev92 - I was incorrect, full path (or URL) is needed for `-Source`.

v-thepet:
@Mikejo5000 and @nkolev92 - Apologies, it looks like I was incorrect with the paths. I retested and it appears that:

- The `-Source` option for `list` must be a full path (or URL). (If no source specified, uses global *NuGet.Config* or default source.)
- The *packages.config* path for `install` can be relative, e.g. `<subdirectory>/packages.config`. 
- The `-OutputDirectory` path for `install` or `restore` can be relative, e.g. `<subdirectory>`.
- For `restore`, looks like if you want to use a *packages.config* rather than a solution for source, you must also specify either `-OutputDirectory` or `PackagesDirectory` options. (I think I'll cover those details in the Restore Packages article, which I'll be working on next.)

I've made some updates to clarify these points. Thanks for your reviews and sorry for my incomplete testing.

nkolev92:
> For restore, looks like if you want to use a packages.config rather than a solution for source, you must also specify either -OutputDirectory or PackagesDirectory options. (I think I'll cover those details in the Restore Packages article, which I'll be working on next.)

What do you mean by source here? 
I'd probably call it the target as it's what's getting restored. 

OutputDirectory and PackagesDirectory should not be mandatory.
Remember that nuget.exe restore can restore PackageReference projects as well, so it can restore without needing OutputDirectory/PackagesDirectory.

docs/consume-packages/includes/restore-nuget-exe-cli.md
```diff
@@ -1,13 +1,21 @@
-Use the [restore](../../reference/cli-reference/cli-ref-restore.md) command, which downloads and installs any packages missing from the *packages* folder.
+The NuGet CLI [restore](../../reference/cli-reference/cli-ref-restore.md) command downloads and installs any packages that are missing from the packages directory.
 
-For projects migrated to PackageReference, use [msbuild -t:restore](../package-restore.md#restore-using-msbuild) to restore packages instead.
+Like `install`, the `restore` command only adds packages to disk, but doesn't change the project's dependencies. To change project dependencies, modify the *packages.config* file manually, then use `restore`.
```
nkolev92, line 3: I think we can clarify that nuget.exe does support PackageReference as well. We just recommend using msbuild or dotnet restore baed on the project flavor. 

Now it kind of sounds like it doesn't.

v-thepet:
@nkolev92 - Thank you!

> > For restore, looks like if you want to use a packages.config rather than a solution for source, you must also specify either -OutputDirectory or PackagesDirectory options.
> 
> What do you mean by source here? I'd probably call it the target as it's what's getting restored.
> 
Yes, "target" is better than "source," although must distinguish between "input target" and "output target."

> OutputDirectory and PackagesDirectory should not be mandatory. Remember that nuget.exe restore can restore PackageReference projects as well, so it can restore without needing OutputDirectory/PackagesDirectory.

Yes, seems they are mandatory only for *package.config* projects. 

I'll clarify these details in the next two articles I am working on: https://docs.microsoft.com/en-us/nuget/consume-packages/finding-and-choosing-packages and https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore.

docs/consume-packages/includes/restore-nuget-exe-cli.md
v-thepet: @nkolev92 - Added "you can." Do you think that's enough?

nkolev92: Can you point where that was added? 
My idea was more about adding it in the description that mentions a packages.config rather than the note itself.

v-thepet: @nkolev92 - I just changed the note to say "you can," meaning they don't have to, they can still use nuget restore if they want. 
image

However it sounds like you want larger revisions, maybe at the beginning of this article you want to not advise using PackageReference or dotnet CLI? 
image

Please advise! Thanks.

nkolev92: I'll try to cover what we our guidance is wrt to using the tools. 

NuGet.exe - supports packages.config and PackageReference, both SDK and non-SDK PackageReference. 
MSBuild.exe - Supports PackageReference, both SDK and non-SDK PackageReference and packages.config next to project through an opt-in
dotnet.exe - Supports PackageReference, only SDK PackageReference. Does not support packages.config. 

When people have PackageReference, we prefer that they use msbuild.exe or dotnet.exe, respectively. 

For example, SDK projects -> dotnet.exe
non-SDK PackageReference - msbuild.exe
mix of SDK and non-SDK -> msbuild.exe. 

We don't have to clarify all of that. 

My thought was:

- say nuget.exe can restore both packages.config and PackageReference
- say that if they have non-SDK PackageReference projects, they should msbuild.exe -t restore.

zivkan, line 3: > To change project dependencies, modify the *packages.config* file manually, then use `restore`.

Customers will not be successful if they take this advice. Editing the packages.config file and then restoring will cause NuGet to download and extract the package to disk. However, the project won't actually use the package in any way. The list of changes to the project file is complex and both project and package dependant, so it's much too complex to document for customers to try to do manually. It's necessary to install the package in Visual Studio so that the project updates are automated.

Hand editing the packages.config file is appropriate for packages.config files that are not associated with projects, however.

v-thepet:
@nkolev92 @zivkan - Thanks! See the changes I made to the Restore section and to the Important alert in the Install section: https://review.docs.microsoft.com/en-us/nuget/consume-packages/install-use-packages-nuget-cli?branch=pr-en-us-2852. Making it clearer that the nuget.exe restore command works on both PackageReference and packages.config, SDK and non-SDK style projects, but if you have non-SDK PackageReference, use msbuild. 

Also removing the advice to modify the packages.config file manually. However I wonder if it still does work to modify <PackageReference> in the csproj file manually... but that and modifying packages.config outside of a project might be edge cases.

zivkan:
> However I wonder if it still does work to modify \<PackageReference\> in the csproj file manually... but that and modifying packages.config outside of a project might be edge cases.

Yes, editing `PackageReference` manually is fine, because it's just a single thing to edit, and everything else happens behind the scenes. I think it was one of the design goals, to be as easy as possible to edit, either through tooling, or hand editing.

docs/consume-packages/includes/restore-nuget-exe-cli.md
v-thepet: Clarified this in both the Restore and Install sections.

v-thepet: @zivkan - I've removed the advice to hand-edit the packages.config file from both the Install and Restore sections.

v-thepet:
@Mikejo5000 @nkolev92 @zivkan - I've made changes to address all comments. Let me know if anything more is needed or whether you approve, and I'll sign off. Thanks for your help!

docs/consume-packages/install-use-packages-dotnet-cli.md
```diff
@@ -1,34 +1,34 @@
...
-ms.date: 06/03/2019
+ms.date: 08/17/2022
 ms.topic: conceptual
 ---
 
-# Install and manage packages using the dotnet CLI
+# Install and manage NuGet packages with the dotnet CLI
 
-The CLI tool allows you to easily install, uninstall, and update NuGet packages in projects and solutions. It runs on Windows, Mac OS X, and Linux.
+You can use the dotnet CLI tool on Windows, MacOS, or Linux to easily install, uninstall, and update NuGet packages in .NET projects and solutions. This article describes the most common dotnet CLI commands for managing NuGet packages.
```
JonDouglas, line 12: nit: `macOS`

docs/consume-packages/includes/restore-nuget-exe-cli.md
```diff
@@ -1,13 +1,21 @@
-Use the [restore](../../reference/cli-reference/cli-ref-restore.md) command, which downloads and installs any packages missing from the *packages* folder.
+The NuGet CLI [restore](../../reference/cli-reference/cli-ref-restore.md) command downloads and installs any packages that are missing from project package directories. The command works on projects that use either `PackageReference` or *packages.config* for package references.
```
JonDouglas, line 1: Links to https://docs.microsoft.com/en-us/nuget/reference/packages-config and https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files may help here

docs/consume-packages/install-use-packages-dotnet-cli.md
```diff
@@ -63,30 +65,32 @@ dotnet add package Newtonsoft.Json --version 12.0.1
...
 ```
 
 ## Remove a package
 
-Use the [dotnet remove package](/dotnet/core/tools/dotnet-remove-package?tabs=netcore2x) command to remove a package reference from the project file.
+Use the [dotnet remove package](/dotnet/core/tools/dotnet-remove-package) command to remove a package reference from the project file.
 
 ```dotnetcli
-dotnet remove package <PACKAGE_NAME>
+dotnet remove package <PackageName>
```
JonDouglas, line 79: We should keep the same parameters as the CLI:

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-remove-package#arguments

docs/consume-packages/install-use-packages-dotnet-cli.md
```diff
@@ -37,22 +37,24 @@ This article shows you basic usage for a few of the most common dotnet CLI comma
...
 ## Install a specific version of a package
 
-If the version is not specified, NuGet installs the latest version of the package. You can also use the [dotnet add package](/dotnet/core/tools/dotnet-add-package?tabs=netcore2x) command to install a specific version of a NuGet package:
+The `dotnet add package` command installs the latest version of the package, unless you specify a different version.
+
+To install a specific version of a NuGet package, use the optional `-v` or `--version` switch:
 
 ```dotnetcli
-dotnet add package <PACKAGE_NAME> --version <VERSION>
+dotnet add package <PackageName> -v <version>
```
JonDouglas, line 57: We should keep the arguments the same as the CLI.

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,120 @@
...
+For most commands, the NuGet CLI tool uses the current directory, unless you specify a different location in the command. To run NuGet CLI commands, open a command line and switch to the directory that contains your project file.
+
+For a complete list of commands and their arguments, see the [NuGet CLI reference](../reference/nuget-exe-cli-reference.md).
 
 ## Prerequisites
 
-- Install the `nuget.exe` CLI by downloading it from [nuget.org](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe), saving that `.exe` file to a suitable folder, and adding that folder to your PATH environment variable.
+Download the NuGet CLI from [nuget.org](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe). Save the *nuget.exe* file to a suitable directory, and make sure the directory is in your PATH environment variable.
```
JonDouglas, line 25: Should we tell users that they can also now use `winget`(Windows) or `brew`(macOS)?

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,120 @@
...
 ```cli
-nuget install packages.config -OutputDirectory packages
+nuget install config\packages.config -OutputDirectory packages
+
 ```
 
 ## Install a specific version of a package
 
-If the version is not specified when you use the [install](../reference/cli-reference/cli-ref-install.md) command, NuGet installs the latest version of the package. You can also install a specific version of a Nuget package:
+The `install` command installs the latest version of a package, unless you specify a different version. To install a specific version of a package, use the `-Version` option:
```
JonDouglas, line 63:
```suggestion
The `install` command installs the latest version of a package unless you specify a different version. To install a specific version of a package, use the `-Version` option:
```

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,120 @@
...
 ```cli
 nuget list -Source <source>
 ```
 
-For example, list packages in the *packages* folder.
+For example, to list packages in the *packages* subdirectory of *MyProject*, use:
 
 ```cli
-nuget list -Source C:\Users\username\source\repos\MyProject\packages
+nuget list -Source C:\Users\<username>\source\repos\MyProject\packages
```
JonDouglas, line 86:
```suggestion
nuget list -Source C:\Users\%USERNAME%\source\repos\MyProject\packages
```

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,120 @@
...
 ```
 
-If you use a search term, the search includes names of packages, tags, and package descriptions.
+If you don't specify a source, `list` uses all sources defined in the *NuGet.Config* global configuration file. If *NuGet.Config* doesn't specify any sources, `list` uses the default feed, for example nuget.org.
```
JonDouglas, line 89:
```suggestion
If you don't specify a source, `list` uses all sources defined in the *NuGet.Config* global configuration file. If *NuGet.Config* doesn't specify any sources, `list` uses a default feed of nuget.org.
```

docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md
```diff
@@ -44,51 +44,52 @@ NuGet packages can be installed into a .NET project of some kind. For this walkt
...
+    string json = JsonConvert.SerializeObject(account, Formatting.Indented);
+    Console.WriteLine(json);
     ```
 
-1. Replace the `Main` function with the following:
+1. Add the following class definition after the `Console.WriteLine(json);` line:
 
     ```cs
-    static void Main(string[] args)
+    namespace nuget1
```
JonDouglas, line 81: Probably want to omit this namespace.

docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md
```diff
@@ -1,40 +1,40 @@
...
-- The [.NET Core SDK](https://www.microsoft.com/net/download/), which provides the `dotnet` command-line tool. Starting in Visual Studio 2017, the dotnet CLI is automatically installed with any .NET Core related workloads.
+- The [.NET SDK](https://www.microsoft.com/net/download), which provides the `dotnet` command-line tool. Starting in Visual Studio 2017, the dotnet CLI automatically installs with any .NET or .NET Core related workloads.
 
 ## Create a project
 
-NuGet packages can be installed into a .NET project of some kind. For this walkthrough, create a simple .NET Core console project as follows:
+You can install NuGet packages into a .NET project. For this walkthrough, create a simple .NET console project by using the dotnet CLI, as follows:
 
-1. Create a folder for the project.
+1. Create a folder named *nuget1* for the project.
```
JonDouglas, line 27: Can we remove the exact name for a folder and tell users to create a folder of their choosing?

JonDouglas (changes requested):
Added a few comments that should be considered before merge.

Thank you so much for your contributions!

If possible in the future, smaller PRs in the scope of each page will be quite appreciated from the review perspective!

docs/consume-packages/includes/restore-nuget-exe-cli.md
v-thepet: @JonDouglas- This include is used in the consume-packages article, so a link to consume-packages would be circular.

docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md
v-thepet: @JonDouglas - From all I have learned, the namespace declaration is necessary for the new template. If you do not declare the namespace the build fails. @nkolev92 @zivkan - Can you comment if this is incorrect and let me know how to arrange the code so the namespace declaration is unnecessary?

docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md
v-thepet: @JonDouglas - Due to the namespace declaration being necessary, and because we want to present quickstart code for the user, the only easy way to do that is to have them create a named folder. I could use placeholders, but then the code would not run without modification.

JonDouglas: I'm just saying the previous version was fine. If we keep the folder name, let's please use a more descriptive name than nuget1.

v-thepet: @JonDouglas - Changed namespace and project name to Nuget.Quickstart.

v-thepet:
@JonDouglas - Thanks for your review. I've implemented all your comments. Future PRs are one article :-).

Mikejo5000 (approved):

v-thepet:
#sign-off. Thanks @Mikejo5000 @nkolev92 @zivkan @JonDouglas.

docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md
nkolev92: The namespaces are top level, so if you're recommending the class be added in the same file, then you likely don't need the namespace as it would've been declared top level.

docs/consume-packages/includes/restore-nuget-exe-cli.md
```diff
@@ -1,13 +1,21 @@
-Use the [restore](../../reference/cli-reference/cli-ref-restore.md) command, which downloads and installs any packages missing from the *packages* folder.
+The NuGet CLI [restore](../../reference/cli-reference/cli-ref-restore.md) command downloads and installs any packages that are missing from project package directories. The command works on projects that use either [PackageReference](/nuget/consume-packages/package-references-in-project-files) or [packages.config](/nuget/reference/packages-config) for package references.
```
nkolev92, line 1: In PackageReference, the packages go to a `global package folder. 

In packages.config, the packages go to a packages folder in the solution root (usually). 

So I think we don't have to be super specifinc and just say missing packages. 

```suggestion
The NuGet CLI [restore](../../reference/cli-reference/cli-ref-restore.md) command downloads and installs any missing packages. The command works on projects that use either [PackageReference](/nuget/consume-packages/package-references-in-project-files) or [packages.config](/nuget/reference/packages-config) for package references.
```

docs/consume-packages/includes/restore-nuget-exe-cli.md
```diff
@@ -1,13 +1,21 @@
...
 
-For projects migrated to PackageReference, use [msbuild -t:restore](../package-restore.md#restore-using-msbuild) to restore packages instead.
+Like `install`, the `restore` command only adds packages to disk, but doesn't modify the project file or *packages.config*. To add project dependencies, use the Visual Studio Package Manager UI or Console, then run `install` or `restore`.
```
nkolev92, line 3:
```suggestion
Like `install`, the `restore` command only adds packages to disk, but doesn't modify the project file or *packages.config*. To add project dependencies, use the Visual Studio Package Manager UI or Console.
```

If you run a package installation through the PM UI or PMC you don't need to run install/restore as the packages will already be downloaded and installed.

docs/consume-packages/install-use-packages-dotnet-cli.md
```diff
@@ -37,22 +37,24 @@ This article shows you basic usage for a few of the most common dotnet CLI comma
     dotnet add package Newtonsoft.Json
     ```
 
-3. After the command completes, look at the project file to make sure the package was installed.
+1. After the command completes and you save the project, you can open the project file to see the package reference.
```
nkolev92, line 40: Given that there's no active process, there's no saving required. It's updated immmediately.

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,123 @@
...
-ms.date: 06/03/2019
+ms.date: 08/29/2022
 ms.topic: conceptual
 ---
 
-# Manage packages using the nuget.exe CLI
+# Manage NuGet packages with the NuGet CLI
 
-The CLI tool allows you to easily update and restore NuGet packages in projects and solutions. This tool provides all NuGet capabilities on Windows, and also provides most features on Mac and Linux when running under Mono.
+You can use the `nuget.exe` CLI tool to install, list, update, remove, and restore NuGet packages in Visual Studio projects and solutions. The NuGet CLI provides all NuGet capabilities on Windows, and provides most NuGet capabilities on Mac and on Linux with Mono. This article describes the most common NuGet CLI commands for managing NuGet packages.
```
nkolev92, line 12: NuGet.exe can't do all the listed things. 

There's no proper list and remove functionalities. 
I think we should only say `restore` and the install/update/remove commands are completed through the PM UI.

List just happens to work for packages.config projects because of a design decision, but it's not really guaranteed to be 100% correct every time

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,123 @@
...
-## Remove a package
-
-To delete one or more packages, delete the packages you want to remove from the *packages* folder.
-
-If you want to reinstall packages, use the `restore` or `install` command.
-
 ## List packages
 
-You can display a list of packages from a given source using the [list](../reference/cli-reference/cli-ref-list.md) command. Use the `-Source` option to restrict the search.
+Use the [list](../reference/cli-reference/cli-ref-list.md) command to display a list of packages. Use the `-Source` option to restrict the list.
```
nkolev92, line 80: I  think we should tweak this a bit to indicate that this shows the packages installed in the packages folder.

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,123 @@
...
-For example, list packages in the *packages* folder.
+For example, to list packages in the *packages* subdirectory of *MyProject*, use:
 
 ```cli
-nuget list -Source C:\Users\username\source\repos\MyProject\packages
+nuget list -Source C:\Users\%USERNAME%\source\repos\MyProject\packages
 ```
 
-If you use a search term, the search includes names of packages, tags, and package descriptions.
+If you don't specify a source, `list` uses all sources defined in the *NuGet.Config* global configuration file. If *NuGet.Config* doesn't specify any sources, `list` uses a default feed of nuget.org.
```
nkolev92, line 92: This information deviates from the context of the doc which is nuget.exe for installing packages.

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,123 @@
...
 
-Use the [update](../reference/cli-reference/cli-ref-update.md) command to update all packages. Updates all packages in a project (using `packages.config`) to their latest available versions. It is recommended to run `restore` before running `update`.
+Use the [update](../reference/cli-reference/cli-ref-update.md) command to update all packages in a project *packages.config* file to their latest available versions. It's best to run `restore` before you run `update`.
 
 ```cli
 nuget update
 ```
 
+## Remove a package
+
```
nkolev92, line 109: I'd recommend that they use the PMC/PM UI. 

Just deleting a package from the packages folder does nothing for the project itself.
Notable that the packages folder tends to be in the solution file repo.

docs/consume-packages/install-use-packages-powershell.md
```diff
@@ -1,178 +1,205 @@
...
 ms.topic: conceptual
 f1_keywords: 
   - "vs.nuget.packagemanager.console"
 ---
 
-# Install and manage packages with the Package Manager Console in Visual Studio (PowerShell)
+# Manage packages with the Visual Studio Package Manager Console (PowerShell)
 
-The NuGet Package Manager Console lets you use [NuGet PowerShell commands](../reference/powershell-reference.md) to find, install, uninstall, and update NuGet packages. Using the console is necessary in cases where the Package Manager UI does not provide a way to perform an operation. To use `nuget.exe` CLI commands in the console, see [Using the nuget.exe CLI in the console](#use-the-nugetexe-cli-in-the-console).
+The Package Manager Console in Visual Studio uses PowerShell commands to interact with NuGet packages. You can use the console when there's no way to do an operation through the [Package Manager UI](install-use-packages-visual-studio.md). You can also use [dotnet CLI](../reference/dotnet-commands.md) or [NuGet CLI](#use-the-nugetexe-cli-in-the-console) commands in the console.
```
nkolev92, line 14: I think it's still be great to link the powershell reference.

docs/consume-packages/install-use-packages-powershell.md
```diff
@@ -1,178 +1,205 @@
...
+- To find and list packages whose name begins with a string:
 
-- The Console displays applicable license terms in its window with implied agreement. If you do not agree to the terms, you should uninstall the package immediately.
-- Also a reference to the package is added to the project file and appears in **Solution Explorer** under the **References** node, you need to save the project to see the changes in the project file directly.
+  ```powershell
+  Find-Package <string> -StartWith
+  ```
 
-## Uninstall a package
+- By default, `Get-Package` returns a list of 20 packages. Use `-First` to show more packages. For example, to show the first 100 packages, use:
```
nkolev92, line 87: did you mean to say `Find-Package` here?

docs/consume-packages/install-use-packages-powershell.md
```diff
@@ -1,178 +1,205 @@
...
+  ```
 
-# List all versions of the package with the ID of "jquery"
-Find-Package jquery -AllVersions -ExactMatch
-```
+<a name="use-the-nugetexe-cli-in-the-console"></a>
+## Use the NuGet CLI in the console
 
-See [Find-Package](../reference/ps-reference/ps-ref-find-package.md). In Visual Studio 2013 and earlier, use [Get-Package](../reference/ps-reference/ps-ref-get-package.md) instead.
+You can also do all operations that are available in the console with the [NuGet CLI](../reference/nuget-exe-cli-reference.md). However, console commands operate within the context of Visual Studio saved project and solution, and often do more than their equivalent NuGet CLI commands. For example, installing a package through the console adds a reference to the project file, but the NuGet CLI command doesn't. For this reason, developers working in Visual Studio typically prefer to use the console commands rather than the NuGet CLI.
```
nkolev92, line 166: Given that we give a disclamier later in ther sentence, should we start with you can do `most operations` or something like that?

nkolev92 (commented):
This looks great. Thanks for all the great work. 

I have some left some suggestions and some places where I think we need further clarification.

docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md
v-thepet: @nkolev92 @zivkan - The namespace has to be declared somewhere. In the old style template it was at the top of the file but in the new style it's supposed to be implied. It seems awkward partway through so I made a change to have them basically replacing the whole contents (one line) of the file rather than doing it in steps. This seems more straightforward. Let me know what you think.

nkolev92: Seems fine by me. 
Sometimes our quickstarts get complicated because we try to follow a bunch of simple steps and we end up having to explain the steps in more detail that its needed. 

I like the simplification.

docs/consume-packages/install-use-packages-nuget-cli.md
v-thepet: @nkolev92 - How about "manage," as that's what the article states it's about, and if we only say "restore" people might think they're in the wrong article.

nkolev92: Yep, great idea.

docs/consume-packages/install-use-packages-nuget-cli.md
v-thepet: @nkolev92 - Not sure what you mean - this series of articles includes the 'list' capability.

docs/consume-packages/install-use-packages-nuget-cli.md
v-thepet: @nkolev92 - Added some language to recommend using VS/Console.

docs/consume-packages/install-use-packages-powershell.md
v-thepet: @nkolev92 - It's linked in the next paragraph.

docs/consume-packages/install-use-packages-powershell.md
v-thepet: @nkolev92 - Also clarified that this is talking about running the nuget.exe commands in the console, instead of PS.

docs/consume-packages/install-use-packages-nuget-cli.md
nkolev92: The part talking about sources being used from nuget.config. 

The `list` functionality lists the packages from a specific source. It just so happens that the packages folder itself has the shape of a source, so it can be used as an approximation to view the packages installed the cunrrent repo/projects.

v-thepet: @nkolev92 - Right but if you start a new project with no dependencies and then do nuget list, it starts listing from nuget.org, is what I'm trying to say.

nkolev92: Unfortunately the command itself is misnamed when compared to commands such as dotnet list package, https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-list-package, which is a true command that lists your installed packages (only works for SDK based projects).

This list command is more of a `search` command, where it searches for a package on certain sources. 
It's equivalent to the `Find-Package` command in PMC. 

The current usage is not the primary intended one for this command. 
I think we should tweak it to link to the command description and then explain this is a way of getting an approximation of installed packages on the commandline

v-thepet: @nkolev92 - OK, I removed that sentence. (The first sentence already links to the command.)

nkolev92 (approved):

v-thepet:
@nkolev92 @JonDouglas @zivkan - I have implemented all corrections and this should be ready for merge, thank you! @Mikejo5000 FYI.

docs/consume-packages/includes/restore-nuget-exe-cli.md
```diff
@@ -1,13 +1,21 @@
...
-For projects migrated to PackageReference, use [msbuild -t:restore](../package-restore.md#restore-using-msbuild) to restore packages instead.
+Like `install`, the `restore` command only adds packages to disk, but doesn't modify the project file or *packages.config*. To add project dependencies, use the Visual Studio Package Manager UI or Console.
 
-`restore` only adds packages to disk but does not change a project's dependencies. To restore project dependencies, modify `packages.config`, then use the `restore` command.
+To restore packages, run the following command:
 
-As with the other `nuget.exe` CLI commands, first open a command line and switch to the directory that contains your project file.
+```cli
+nuget restore <projectPath>
+```
```
zivkan, line 9: I think there's value in duplicating the following here as well: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#restoring-packagereference-and-packagesconfig-projects-with-msbuild

MSBuild 16.5+ supports restoring packages.config that are in the same directory as a solution project `msbuild -t:restore -p:RestorePackagesConfig=true`.

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -1,118 +1,123 @@
...
+For a complete list of commands and their arguments, see the [NuGet CLI reference](../reference/nuget-exe-cli-reference.md).
 
 ## Prerequisites
 
-- Install the `nuget.exe` CLI by downloading it from [nuget.org](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe), saving that `.exe` file to a suitable folder, and adding that folder to your PATH environment variable.
+Download the NuGet CLI from [nuget.org](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe). Save the *nuget.exe* file to a suitable directory, and make sure the directory is in your PATH environment variable.
+
+> [!NOTE]
+> You can also use the [winget](/windows/package-manager/winget) tool for Windows or [Homebrew](https://brew.sh/) for macOS.
```
zivkan, line 28: I really hate nuget.exe, so it makes me sad that we're making it easier for customers to keep depending on it 😢

docs/consume-packages/install-use-packages-powershell.md
```diff
@@ -1,178 +1,205 @@
 ---
-title: Install and manage NuGet packages using the console in Visual Studio
-description: Instructions for using the NuGet Package Manager Console in Visual Studio for working with packages.
+title: Manage NuGet packages with the Visual Studio Package Manager Console
+description: See how to work with NuGet packages by using PowerShell commands in the Visual Studio Package Manager Console.
 author: JonDouglas
 ms.author: jodou
-ms.date: 07/08/2019
+ms.date: 08/29/2020
```
zivkan, line 6:
```suggestion
ms.date: 08/29/2022
```

docs/quickstart/install-and-use-a-package-using-the-dotnet-cli.md
```diff
@@ -1,40 +1,40 @@
...
 
-NuGet packages contain reusable code that other developers make available to you for use in your projects. See [What is NuGet?](../What-is-NuGet.md) for background. Packages are installed into a .NET Core project using the `dotnet add package` command as described in this article for the popular [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json/) package.
+NuGet packages contain compiled binary code that developers make available for other developers to use in their projects. For more information, see [What is NuGet](../What-is-NuGet.md). This quickstart describes how to install the popular [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json) NuGet package into a .NET project by using the [dotnet add package](/dotnet/core/tools/dotnet-add-package) command.
 
-Once installed, refer to the package in code with `using <namespace>` where \<namespace\> is specific to the package you're using. You can then use the package's API.
+You refer to installed packages in code with a `using <namespace>` directive, where `<namespace>` is the package name. You can then use the package's API in your project.
```
zivkan, line 14:
```suggestion
You refer to installed packages in code with a `using <namespace>` directive, where `<namespace>` is often the package name. You can then use the package's API in your project.
```

There's nothing preventing a package's namespace to be different from the package id. Maybe I'm just being overly pedantic.

zivkan (approved):
I don't consider any of my comments blocking issues.

@JonDouglas would you like to review and merge?

docs/consume-packages/includes/restore-dotnet-cli.md
```diff
@@ -1,9 +1,9 @@
-Use the [dotnet restore](/dotnet/core/tools/dotnet-restore?tabs=netcore2x) command, which restores packages listed in the project file (see [PackageReference](../../consume-packages/package-references-in-project-files.md)). With .NET Core 2.0 and later, restore is done automatically with `dotnet build` and `dotnet run`. As of NuGet 4.0, this runs the same code as `nuget restore`.
+The [dotnet restore](/dotnet/core/tools/dotnet-restore) command restores packages that the project file lists with `<PackageReference>`. For more information, see [PackageReference in project files](../../consume-packages/package-references-in-project-files.md).
 
-As with the other `dotnet` CLI commands, first open a command line and switch to the directory that contains your project file.
+.NET Core 2.0 and later `dotnet build` and `dotnet run` commands restore packages automatically. As of NuGet 4.0, `dotnet restore` runs the same code as `nuget restore`.
```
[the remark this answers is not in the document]
nkolev92: I think that's a harmless change that we can make if we choose to.

nkolev92:
@v-thepet Is this merge ready now? 

There's a few comments from @zivkan left that might need addressing.

zivkan:
I'm happy for this to be merged without additional changes.

PR #2859, opened 2022-08-27
Update and refresh Find and choose packages

v-thepet:
User Story https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1973816


v-thepet:
@Mikejo5000 @nkolev92 @zivkan @JonDouglas - This article is ready for your review. It's pretty straightforward except that the nuget.org UI has changed, so I took all new screenshots and updated the text accordingly. 

One nit: the original article stated that in the Visual Studio NuGet Package Manager UI, the include prerelease tab is NOT checked. I'm about 95% sure that in my Visual Studio Package Manager, that box IS (was) checked by default. (But as stated in the original article, the Console, dotnet CLI, and nuget CLI do NOT include prerelease by default.) 

Thanks for your review and approval if approved.

nkolev92 (approved):

docs/consume-packages/Finding-and-Choosing-Packages.md
```diff
@@ -1,171 +1,182 @@
...
-ms.date: 11/03/2021
+ms.date: 08/29/2022
 ms.topic: conceptual
 ---
 
-# Finding and evaluating NuGet packages for your project
+# Find and evaluate NuGet packages for your project
 
-When starting any .NET project, or whenever you identify a functional need for your app or service, you can save yourself lots of time and trouble by using existing NuGet packages that fulfill that need. These packages can come from the public collection on [nuget.org](https://www.nuget.org/packages/), or a private source that's provided by your organization or another third party.
+When you start a .NET project, or identify a functional need in your app or service, you can often install existing NuGet packages to save the time and trouble of [creating your own packages](../create-packages/overview-and-workflow.md). Existing packages can come from the [nuget.org](https://www.nuget.org/packages) public collection, or from private sources that your organization or another party provide.
```
zivkan, line 12: I only glanced over all the changes in this file. Maybe I missed it, but I didn't see anything to explicitly tell customers that all these advanced filtering options are server dependent. Especially because of the last sentence here, saying private sources, I think it's important to tell customers it might not work as expected.  NuGet.org only started supporting the syntax since last year, so there was 10 years where even nuget.org didn't support the syntax. So it's likely that many non-nuget.org package sources do not support any of these advanced syntaxes.

I don't have an idea on how to briefly say this, but hopefully you can come up with something. I hope that you also agree it's worth mentioning to customers.

v-thepet: @zivkan - It's here, first sentences under Search syntax:

### Search syntax
Package search queries at nuget.org, from the NuGet CLI, and from within Visual Studio all use the same syntax. Other package sources, like Azure Artifacts or GitHub Package Repository, might use different syntax or might not support advanced filtering.

Do you think that's enough?

zivkan: Oops, sorry I missed that. After reviewing 5+ long docs PRs, I was rushing.  I had hoped to do some coding today 😁

PR #2860, opened 2022-08-27
Update and refresh Restore packages

v-thepet:
https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1973816


v-thepet:
@Mikejo5000 @nkolev92 @zivkan @JonDouglas - This article is ready for your review. A couple of notes:

- I haven't yet tested/worked through the more advanced procedures for setting environment variables and editing the global NuGet.Configs. It seemed more important to get this draft reviewed asap. I tested the basic procedures and the others seem like they should work, but I can test more later or if there are questions.
- The dotnet CLI and nuget.exe CLI restore procedures use the includes that I edited in PR https://github.com/NuGet/docs.microsoft.com-nuget/pull/2852. To prevent confusion or merge conflicts, I've just included that note for now. As soon as PR 2852 merges, I'll uncomment the calls to the includes. 

Thanks for your review.

docs/consume-packages/Package-Restore.md
```diff
@@ -1,242 +1,222 @@
...
-
-> [!Note]
-> NuGet doesn't indicate a failure to restore a package until all the sources have been checked. At that time, NuGet reports a failure for only the last source in the list. The error implies that the package wasn't present on *any* of the other sources, even though errors aren't shown for each of those sources individually.
-
-## Restore packages
+Package Restore tries to install all package dependencies to the state that matches the `<PackageReference>`s in a project file, such as *.csproj*, or `<package>`s in a *packages.config* file. In Visual Studio **Solution Explorer**, you can see the package references under **Dependencies** > **Packages**.
```
nkolev92, line 18: Only PackageReference projects have their packages listed in the SolutionExplorer. 
Should we tweak this to clarify it or at least to not mistakenly suggest that packages.config has the same capability.

docs/consume-packages/Package-Restore.md
```diff
@@ -1,242 +1,222 @@
...
-1. If the package references in your project file are correct, use your preferred tool to restore packages.
+Package Restore first installs the direct dependencies of a project as needed, then installs any dependencies of those packages throughout the entire dependency graph.
 
-   - [Visual Studio](#restore-using-visual-studio) ([automatic restore](#restore-packages-automatically-using-visual-studio) or [manual restore](#restore-packages-manually-using-visual-studio))
-   - [dotnet CLI](#restore-using-the-dotnet-cli)
-   - [nuget.exe CLI](#restore-using-the-nugetexe-cli)
-   - [MSBuild](#restore-using-msbuild)
-   - [Azure Pipelines](#restore-using-azure-pipelines)
-   - [Azure DevOps Server](#restore-using-azure-devops-server)
+If a needed package isn't already installed, NuGet first attempts to retrieve it from the [NuGet cache](../consume-packages/managing-the-global-packages-and-cache-folders.md). If the package isn't in the cache, NuGet tries to download it from all sources configured in Visual Studio at **Tools** > **Options** > **NuGet Package Manager** > **Package Sources**.
```
nkolev92, line 22: Maybe we link the how package installation works doc? 

Seems like we've had multiple descriptions of the same concepts and it doesn't shock me they've diverged. 

The tldr is that cache is an oversimplification and a generic buzzword that doesn't really explain much.

docs/consume-packages/Package-Restore.md
```diff
@@ -1,242 +1,222 @@
...
+To use MSBuild restore:
 
-   ```cmd
-   # Uses the project file in the current folder by default
-   msbuild -t:restore
-   ```
+1. Open a Developer Command Prompt by searching for *developer command prompt* and starting the prompt from the Windows **Start** menu, which configures all the necessary paths for MSBuild.
 
-3. Type the following command to rebuild the project.
+1. Switch to the folder containing the project or *packages.config* file, and enter `msbuild -t:restore`.
```
nkolev92, line 144: while msbuild.exe /t:restore does support packages.config restore through an opt in property `RestorePackagesConfig`, https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#restore-properties it only supports it in the context of a project, so just saying switching to the project folder is best.

docs/consume-packages/Package-Restore.md
```diff
@@ -1,242 +1,222 @@
...
-ms.date: 08/05/2019
+ms.date: 08/29/2022
 ms.topic: conceptual
 ---
 
-# Restore packages using Package Restore
+# Restore packages with NuGet Package Restore
 
-To promote a cleaner development environment and to reduce repository size, NuGet **Package Restore** installs all of a project's dependencies listed in either the project file or `packages.config`. The .NET Core 2.0+ `dotnet build` and `dotnet run` commands do an automatic package restore. Visual Studio can restore packages automatically when it builds a project, and you can restore packages at any time through Visual Studio, `nuget restore`, `dotnet restore`, and xbuild on Mono.
+NuGet Package Restore restores all of a project's dependencies that are listed in either a project file or a *packages.config* file. You can restore packages at any time with `nuget restore`, `dotnet restore`, or through Visual Studio. The `dotnet build` and `dotnet run` commands automatically restore packages, and you can configure Visual Studio to restore packages automatically when it builds a project.
```
zivkan, line 12:
```suggestion
NuGet Package Restore restores all of a project's dependencies that are listed in either a project file or a *packages.config* file. You can restore packages at any time with `nuget restore`, `dotnet restore`, `msbuild -t:restore`, or through Visual Studio. The `dotnet build` and `dotnet run` commands automatically restore packages, and you can configure Visual Studio to restore packages automatically when it builds a project.
```

docs/consume-packages/Package-Restore.md
zivkan: Regardless of customers calling the global packages folder the "nuget cache", we generally do not. If you run `dotnet nuget locals all -l`, it will list all the special folders NuGet uses, and there is an HTTP cache, but the linked docs filename (managing the global packages and cache folders) calls out "global packages" differently from "cache", so I don't think we should use "NuGet cache" anywhere in our docs.

v-thepet: @nkolev92 @zivkan - Although, unfortunately:

image

docs/consume-packages/Package-Restore.md
v-thepet: @zivkan - Decided I didn't like "at any time," replaced with "manually."

docs/consume-packages/Package-Restore.md
v-thepet: Clarified the language.

nkolev92 (approved):

v-thepet:
@nkolev92 @zivkan @JonDouglas @Mikejo5000 - I've implemented all comments/corrections, and I've also uncommented the includes, so this PR should be ready to merge, thanks. (The includes will update when PR #2852 merges.)

docs/consume-packages/Package-Restore.md
zivkan: `dotnet nuget locals all -l` outputs:

```text
http-cache: C:\Users\zivkan\AppData\Local\NuGet\v3-cache
global-packages: C:\Users\zivkan\.nuget\packages\
temp: C:\Users\zivkan\AppData\Local\Temp\NuGetScratch
plugins-cache: C:\Users\zivkan\AppData\Local\NuGet\plugins-cache
```

so, there are two cache directories (http and plugins) that will get cleared, and nuget's temp directory, and also the global-packages directory. In other words, the button clears more than just the global packages directory.

v-thepet: @zivkan - Right, I had to demonstrate that to myself because the label is misleading. Just saying that if you're trying to get people away from calling it "the cache" or "NuGet Cache," the button label is unfortunate.

zivkan: Honestly, I'm extremely tempted to change the label to "Clear NuGet directories" or "Clear NuGet storage". Changes wouldn't take effect until VS17.4 at the earliest, but I might create that PR today.

PR #2864, opened 2022-08-30
Update allowlist

lyndaidaii:
We are adding new domain allowlist: 
https://github.com/NuGet/NuGetGallery/pull/9217


joelverhagen (approved):

shishirx34 (approved):

shishirx34:
Looks like the validation doesn't like accounts which no longer exists in the authoring section for authoring and readme. Should we update these authors to current PMs? Perhaps Jon or Claire, that way someone has the ownership of docs? @joelverhagen - thoughts?

joelverhagen:
Nice catch @shishirx34. @lyndaidaii, to clear those warnings you can put Claire (`cnov`) for the `ms.author` and `clairernovotny` for the `author`.

JonDouglas:
Also consider taking a look at the doc and updating the `ms.date` if it is still "fresh" :)

PR #2884, opened 2022-09-12
New FAQ 

anangaur:
Documenting https://github.com/NuGet/Announcements/issues/60 as a FAQ

### DIFF
--- a/docs/nuget-org/nuget-org-faq.yml
+++ b/docs/nuget-org/nuget-org-faq.yml
@@ -264,28 +264,28 @@ sections:
           
           Our customer support will process this request and perform the account deletion.
           
-        - question: Error connecting to nuget.org or restoring packages from `https://api.nuget.org/v3/index.json`?
-          answer: |
-            If you are experiencing connectivity issues while connecting to nuget.org or restoring packages from https://api.nuget.org/v3/index.json, such as:
+      - question: Error connecting to nuget.org or restoring packages from `https://api.nuget.org/v3/index.json`?
+        answer: |
+          If you are experiencing connectivity issues while connecting to nuget.org or restoring packages from https://api.nuget.org/v3/index.json, such as:
 
-            > "The SSL connection could not be established"
-            
-            , or
-            
-            > “[nuget.org] Unable to load the service index for source https://api.nuget.org/v3/index.ison. An error occurred while sending the request.
+          > "The SSL connection could not be established"
+
+          , or
+
+          > “[nuget.org] Unable to load the service index for source https://api.nuget.org/v3/index.ison. An error occurred while sending the request.
 The underlying connection was closed: An unexpected error occurred on a send. Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
 An existing connection was forcibly closed by the remote host.”
 
-            , or
-            
-            > “This page can’t be displayed. Turn on TLS 1.0, TLS 1.1, and TLS 1.2 in Advanced settings and try connecting to https://api.nuget.org again. If this error persists, it is possible that this site uses an unsupported protocol or cipher suite such as RC4 (link for the details), which is not considered secure. Please contact your site administrator.”
-          
-          , this could be related to [the changes to nuget.org supported cipher suites](https://github.com/NuGet/Announcements/issues/60). 
-          
-          Please ensure that TLS 1.2 and any of the following cipher suites have been enabled on your machine:
+          , or
 
-          `TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384`
-          `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`
-          `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384`
-          `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256`
-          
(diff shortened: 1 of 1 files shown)


docs/nuget-org/nuget-org-faq.yml
```diff
@@ -263,3 +263,29 @@ sections:
...
+The underlying connection was closed: An unexpected error occurred on a send. Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
+An existing connection was forcibly closed by the remote host.”
+
+          , or
+
+          > “This page can’t be displayed. Turn on TLS 1.0, TLS 1.1, and TLS 1.2 in Advanced settings and try connecting to https://api.nuget.org again. If this error persists, it is possible that this site uses an unsupported protocol or cipher suite such as RC4 (link for the details), which is not considered secure. Please contact your site administrator.”
+
+        , this could be related to [the changes to nuget.org supported cipher suites](https://github.com/NuGet/Announcements/issues/60). 
+
+        Please ensure that TLS 1.2 and any of the following cipher suites have been enabled on your machine:
```
zivkan, line 285: Is it worth linking to the Azure DevOps TLS checker? https://github.com/microsoft/azure-devops-tls12

I don't know if it checks the same cyphers as what you listed below, but if it does, it might be an easy way for customers to check.  Despite all my knowledge of Windows, I honestly wouldn't have a clue how to check, and this docs update doesn't link to anything that tells customers how to check, though I recognise that it depends on the operating system, and maybe even which version of an operating system family.

JonDouglas: does https://www.ssllabs.com/ssltest/ cover this?

erdembayar (approved):

JonDouglas:
We need some cleanup to the YAML now that I fixed tabs. If anyone wants to do that, feel free! (See the preview to see what I mean)

docs/nuget-org/nuget-org-faq.yml
zivkan: No. The website tests your browser's TLS/SSL settings, but browsers don't use Windows' SChannel APIs, they ship their own SSL library (probably OpenSLL). So, checking that website in your browser does not validate that .NET apps can use the appropriate TLS ciphers.

JonDouglas: Ah. I just mostly meant for an end user to test what cipher suites have been enabled:

i.e.

image

Either way sounds good to me!

zivkan: > Ah. I just mostly meant for an end user to test what cipher suites have been enabled:

I'm not 100% sure about Edge (the new "Edgium", not the old Edge that is now deprecated), but I'm reasonably confident that Chrome and Firefox ship their own OpenSSL and therefore don't use Windows' configuration. So my point stands that opening that web page in a browser will test the browser's cipher suite, not the "Windows" configuration that .NET uses.

zhhyu: Chrome has its own cipher suites, while .NET/Windows keeps a different set:
https://support.venafi.com/hc/en-us/articles/360046174112-Validate-Cipher-Suites-Offered-to-Servers-from-Windows

docs/nuget-org/nuget-org-faq.yml
```diff
@@ -263,3 +263,29 @@ sections:
...
+          The underlying connection was closed: An unexpected error occurred on a send. Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
+          An existing connection was forcibly closed by the remote host.”
+
+          , or
+
+          > “This page can’t be displayed. Turn on TLS 1.0, TLS 1.1, and TLS 1.2 in Advanced settings and try connecting to https://api.nuget.org again. If this error persists, it is possible that this site uses an unsupported protocol or cipher suite such as RC4 (link for the details), which is not considered secure. Please contact your site administrator.”
+
+          , this could be related to [the changes to nuget.org supported cipher suites](https://github.com/NuGet/Announcements/issues/60). 
+
+          Please ensure that TLS 1.2 and any of the following cipher suites have been enabled on your machine:
```
zhhyu, line 285: >the following cipher suites [](http://example.com/codeflow?start=48&length=27)

I am a little concerned about listing the cipher suites here:
1. This list of cipher suites is now used by one of the two CDN providers we have, and the two CDN providers support different sets of cipher suites today. We run the active failover in different regions to ensure the continuous operation of the service, so there is no guarantee that applying these cipher suites work for customers (especially the latter two for legacy systems). Though we know that what we have at the two CDN providers support the known systems for customers.
2. For the CDN provider who is using this list of cipher suites, they plan to switch to a different set by the end of this year, so it will be changed soon. Once they finish the work on their side, we have the chance to align the cipher suites used at the two CDN provider early next year. If we'd like to show what cipher suites our endpoint supports in the document, it may be better to do that until we align the two CDN providers next year.
3. Checking SSL Labs helps but it doesn't mean that the result is correct. Because the node that SSL Labs runs the test may locate in a different region from the customer, and in that location, it may be a different CDN provider with a different set of cipher suites.

zhhyu:
Please review this PR: https://github.com/NuGet/docs.microsoft.com-nuget/pull/3001. Thanks!

zhhyu:
Let's follow this PR to finalize this issue: https://github.com/NuGet/docs.microsoft.com-nuget/pull/3001. Thanks!

Closed this PR.

PR #2904, opened 2022-09-30
Fix project end tag in example xmls

Speeedy01:


erdembayar (approved):
Good catch.

JonDouglas (approved):
Thank you kindly for your contribution!

Issue #2525, opened 2021-09-03
Update Deprecation Screenshots


dominoFire:
New Deprecation experiences were introduced in https://github.com/NuGet/NuGet.Client/pull/4144

We should update Deprecation screenshots and update instructions, if needed

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a01e821e-fdaa-ddeb-584c-e3d1af0ce964
* Version Independent ID: f919df6a-2e9e-d289-4628-f43d3330a9fe
* Content: [Deprecating packages on nuget.org](https://docs.microsoft.com/en-us/nuget/nuget-org/deprecate-packages#visual-studio)
* Content Source: [docs/nuget-org/Deprecate-packages.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/nuget-org/Deprecate-packages.md)
* Product: **nuget**
* GitHub Login: @anangaur
* Microsoft Alias: **anangaur**

JonDouglas:
Steps for whoever picks this up:

1. Download the most recent Visual Studio
2. Create a new project 
3. Install a couple of packages from the top list in Browse
4. Find a deprecated package such as `My.Sample.Lib` - https://www.nuget.org/packages/My.Sample.Lib/6.0.1
5. Take a screenshot of the installed tab.

nkolev92:
Fixed in https://github.com/NuGet/docs.microsoft.com-nuget/pull/2906

PR #2906, opened 2022-10-03
Update Deprecation Screenshots as per guidance in #2525

ianrathbone:
Updated the screenshot for deprecations for latest Visual Studio  as per #2525


JonDouglas (commented):
Thank you so much for your contribution!

I did notice the screenshot we are replacing here is a little larger than the one provided here. Could we enhance this further with a 2300x750 px image? The one here is 1800x400 and somewhat gets shrunk in the previews.

PR #2932, opened 2022-11-04
Suggest a more generally appropriate import tag

AArnott:
The sample import assumes the next parent level import comes from one directory above. My proposed change will work regardless of the number of directories that separate the two levels.

### DIFF
--- a/docs/consume-packages/Central-Package-Management.md
+++ b/docs/consume-packages/Central-Package-Management.md
@@ -92,7 +92,7 @@ Repository
 - Project1 will evaluate the `Directory.Packages.props` file in the `Repository\Solution1\` directory and it must manually import the next one if so desired.
   ```xml
   <Project>
-    <Import Project="..\Directory.Packages.props">
+    <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)../, Directory.Packages.props))\Directory.Packages.props" Condition=" '$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)../, Directory.Packages.props))' != '' " />
     <ItemGroup>
       <PackageVersion Update="Newtonsoft.Json" Version="12.0.1" />
     </ItemGroup>


docs/consume-packages/Central-Package-Management.md
```diff
@@ -92,7 +92,7 @@ Repository
 - Project1 will evaluate the `Directory.Packages.props` file in the `Repository\Solution1\` directory and it must manually import the next one if so desired.
   ```xml
   <Project>
-    <Import Project="..\Directory.Packages.props">
+    <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)../, Directory.Packages.props))\Directory.Packages.props" Condition=" '$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)../, Directory.Packages.props))' != '' " />
```
jeffkl, line 95: Thanks for improving our documentation.  What do you think about using `GetPathOfFileAbove` instead and not having the condition?

```suggestion
    <Import Project="<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Packages.props', '$(MSBuildThisFileDirectory)../'))" />" />
```

AArnott: Oh! I like `GetPathOfFileAbove`. I didn't know about that. Is it new?
Removing the condition will cause it to fail when no such file exists above. That may be fine given the context of this sample. I could go either way.

nit: your proposed line change nests an extra import tag (invalid syntax).

jeffkl: > Oh! I like GetPathOfFileAbove. I didn't know about that. Is it new?

I technically shipped it in 2016 (https://github.com/dotnet/msbuild/pull/1448) but a lot of people still use the `GetDirectoryNameOfFileAbove()` method instead

> Removing the condition will cause it to fail when no such file exists above. That may be fine given the context of this sample. I could go either way.

I _think_ its sort of necessary if you have this import that you have a file to be imported and in my opinion its better to get an error than nothing because in that case something above wasn't imported?  What do you think?

> nit: your proposed line change nests an extra import tag (invalid syntax).

Oops copy/paste error 🤦‍♂️

jeffkl (approved):

docs/consume-packages/Central-Package-Management.md
AArnott: Great points. And thank you for shipping it then. :)

BTW, I came up with a more generalized idiom. Should we use it here?
```xml
<Import Project="$([MSBuild]::GetPathOfFileAbove($(MSBuildThisFile), $(MSBuildThisFileDirectory)..))" />
```

zivkan (approved):

PR #2765, opened 2022-05-26
Update Central-Package-Management.md

KyleMcMaster:
This line should be removed as central package management is now supported in Visual Studio 17.2 and .NET 6.0.300 SDK.

PR #2950, opened 2022-11-30

dtaylor84:
Remove outdated and contradicted warning.


nkolev92:
Thanks for creating this pr @dtaylor84 

https://github.com/NuGet/docs.microsoft.com-nuget/pull/2765 is an older PR that's attempting to make the same change. So at this point, we'd like to give priority to that chnage.

PR #2968, opened 2022-12-19
Fix command to set symbolic link for NuGet.Config

paulroho:


docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -33,7 +33,7 @@ These other tools look for the user config file under the `~/.config/NuGet` fold
 
 If the tooling you use involves both locations, consider consolidating them by following these steps to allow you to work with only one user-level config file:  
 1. Check the contents of the two user-level config files and keep the one you want under `~/.nuget/NuGet` folder. 
-2. Set symbolic link from `~/.nuget/NuGet` to `~/.config/Nuget`. E.g. Run bash command: `ln -s ~/.nuget/NuGet ~/.config/Nuget`.
+2. Set symbolic link from `~/.nuget/NuGet` to `~/.config/Nuget`. E.g. Run bash command: `ln -s ~/.nuget/NuGet/NuGet.Config ~/.config/NuGet/NuGet.Config`.
```
erdembayar, line 36: Doesn't directory level symbolic link work this case? Why we're using file level symbolic link?

zivkan: +1 to Erick's question.

Since NuGet added the "[additional user wide configuration](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#additional-user-wide-configuration)", if a directory symlink works, then it'll automatically pick up these additional `config/*.config` files. Otherwise to be 100% equivalent, additional symlinks would be needed to ensure both mono nuget.exe and dotnet cli use the same config.

paulroho: I see your point. My problem started, that the suggested command `ln -s ~/.nuget/NuGet ~/.config/Nuget` creates the symlink to `~/.nuget/NuGet` within `~/.config/NuGet/`, so basically providing NuGet.config in `~/.config/NuGet/NuGet/NuGet.config`.

Now using `ln -s ~/.nuget/NuGet/ ~/.config` (of course after deleting the old NuGet directory) puts it in the proper place to me at least. Now both `nuget`, `dotnet nuget` and Rider are happy and consistent.

So to me it seems, the command should be changed as stated above. What's your take on it?

erdembayar: Last approach ltgm. Which OS are you using? Probably we need to check combination matrix for MAC/Linux/dotnet/nuget/mono before approving this, don't one combination get fixed but other combination fails after this change.

zivkan:
@kartheekp-ms the PR author never pushed a commit to implement https://github.com/NuGet/docs.microsoft.com-nuget/pull/2968/files#r1053212241. I don't know if I noticed this when this PR first came out, but the PR's change also makes it so the example command no longer matches the description of the mitigation, earlier on the same line. Plus GitHub says there's a file in conflict, so this PR can't be merged until it's resolved. Therefore, this PR isn't ready for review/approval at this time.

nkolev92:
Closing this. Happy to reconsider if/when the PR is up to date.

Issue #2963, opened 2022-12-16
Examples for registration base url need placeholder not hardcoded values


erdembayar:
Please check this example: https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#sample-request-2
It has hard coded baseurl so it fails if I make call now because that `registration3` was deprecated long ago. Instead, we should use placeholder in examples and refer how to get latest baseurl with sdk.
![image](https://user-images.githubusercontent.com/8766776/207996141-da79bc16-d44d-4859-9798-3f3745d2e71d.png)

Failing example: `GET https://api.nuget.org/v3/registration3/text2xml.lib/1.1.4.json`
![image](https://user-images.githubusercontent.com/8766776/207996906-08268074-f5a5-4d1c-ac6b-283e633e65db.png)

As of today baseurl looks like below and proper request url would be https://api.nuget.org/v3/registration5-gz-semver2/text2xml.lib/1.1.4.json

![image](https://user-images.githubusercontent.com/8766776/207996385-d22133dd-79c4-455d-a9b2-b55703ff8c56.png)

erdembayar:
Fyi @joelverhagen

PR #2971, opened 2022-12-21
Use sample values for the base URLs that change sometimes on NuGet.org

joelverhagen:
Resolve https://github.com/NuGet/docs.microsoft.com-nuget/issues/2963

### DIFF
--- a/docs/api/_data/package-registration-index.json
+++ b/docs/api/_data/package-registration-index.json
@@ -2,11 +2,11 @@
   "count": 1,
   "items": [
     {
-      "@id": "https://api.nuget.org/v3/registration3/nuget.server.core/index.json#page/3.0.0-beta/3.0.0-beta",
+      "@id": "https://api.nuget.org/v3/registration-sample/nuget.server.core/index.json#page/3.0.0-beta/3.0.0-beta",
       "count": 1,
       "items": [
         {
-          "@id": "https://api.nuget.org/v3/registration3/nuget.server.core/3.0.0-beta.json",
+          "@id": "https://api.nuget.org/v3/registration-sample/nuget.server.core/3.0.0-beta.json",
           "catalogEntry": {
             "@id": "https://api.nuget.org/v3/catalog0/data/2017.10.05.18.41.33/nuget.server.core.3.0.0-beta.json",
             "authors": ".NET Foundation",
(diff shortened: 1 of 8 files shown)


camigthompson (approved):

docs/api/_data/package-registration-index.json
```diff
@@ -2,11 +2,11 @@
   "count": 1,
   "items": [
     {
-      "@id": "https://api.nuget.org/v3/registration3/nuget.server.core/index.json#page/3.0.0-beta/3.0.0-beta",
+      "@id": "https://api.nuget.org/v3/registration-sample/nuget.server.core/index.json#page/3.0.0-beta/3.0.0-beta",
```
erdembayar, line 5: Is there any example or guidance on how to get this `registation-sample` value from index page? Otherwise, takes some time to figure out what it's.

docs/api/search-query-service-resource.md
```diff
@@ -143,9 +143,11 @@ name      | string  | yes      | The name of the package type.
 ### Sample request
 
 ```
-GET https://azuresearch-usnc.nuget.org/query?q=NuGet.Versioning&prerelease=false&semVerLevel=2.0.0
+GET https://search-sample.nuget.org/query?q=NuGet.Versioning&prerelease=false&semVerLevel=2.0.0
```
erdembayar, line 146: Same as https://github.com/NuGet/docs.microsoft.com-nuget/pull/2971/files#r1054651582.
How to get this value if someone coming to this page 1st time?

docs/api/_data/package-registration-index.json
joelverhagen: It's in the service index page, and at the top of the document.

docs/api/search-query-service-resource.md
joelverhagen: https://github.com/NuGet/docs.microsoft.com-nuget/pull/2971#discussion_r1054656517

docs/api/_data/package-registration-index.json
erdembayar: Do you mean this part? I don't see anything particular about selecting correct registration url.

image

joelverhagen: image

erdembayar: Here we have 2 base urls. Currently they're same, but in case they're different then which one to use?
https://user-images.githubusercontent.com/8766776/207996385-d22133dd-79c4-455d-a9b2-b55703ff8c56.png

joelverhagen: The `/Versioned` one is undocumented (in line with https://learn.microsoft.com/en-us/nuget/api/overview#undocumented-resources-on-nugetorg) and should not be used. You select the `@type` you want to use based on https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#versioning.

erdembayar (approved):

PR #2988, opened 2023-02-01
Update install-and-use-a-package-in-visual-studio.md

albarry4:
I got an error when trying to run this that stated: "Formatting" is an ambiguous reference between "Newtonsoft.Json.Formatting" and "System.Xml.Formatting", so I had to insert the "Newtonsoft.Json." before the "Formatting" myself. Assuming this could be an issue for other users, we should include this text here by default to make the tutorial as smooth as possible.

### DIFF
--- a/docs/quickstart/install-and-use-a-package-in-visual-studio.md
+++ b/docs/quickstart/install-and-use-a-package-in-visual-studio.md
@@ -105,7 +105,7 @@ With the `Newtonsoft.Json` package in the project, call its `JsonConvert.Seriali
             Email = "john@microsoft.com",
             DOB = new DateTime(1980, 2, 20, 0, 0, 0, DateTimeKind.Utc),
         };
-        string json = JsonConvert.SerializeObject(account, Formatting.Indented);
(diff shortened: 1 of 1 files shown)


docs/quickstart/install-and-use-a-package-in-visual-studio.md
```diff
@@ -105,7 +105,7 @@ With the `Newtonsoft.Json` package in the project, call its `JsonConvert.Seriali
             Email = "john@microsoft.com",
             DOB = new DateTime(1980, 2, 20, 0, 0, 0, DateTimeKind.Utc),
         };
-        string json = JsonConvert.SerializeObject(account, Formatting.Indented);
+        string json = JsonConvert.SerializeObject(account, Newtonsoft.Json.Formatting.Indented);
```
erdembayar, line 108: Instead, we could add `using` statement. Either way, I'm fine with it.

erdembayar (approved):
My comment is not blocking.

jeffkl (approved):

PR #2991, opened 2023-02-02
December/January merge

nkolev92:


nkolev92:
Unfortunately https://github.com/NuGet/docs.microsoft.com-nuget/pull/2934 was a squash merge and that's not what we should be doing for the live branch. I'll try to disable that if possible.

nkolev92:
Unfortunately you can't really ban squash and merge for a single branch. 
This is just something we'll have to remember. 

Tagging folks as an FYI, @NuGet/nuget-pm

PR #3001, opened 2023-02-14
[NuGet.org FAQ] TLS/Cipher suites and Client resilience for transient failures

zhhyu:
Update NuGet.org FAQ page on TLS/Cipher suites and how to enhance client resilience for transient failures.
https://learn.microsoft.com/en-us/nuget/nuget-org/nuget-org-faq#nuget-org-not-accessible


docs/nuget-org/nuget-org-faq.yml
```diff
@@ -60,33 +60,55 @@ sections:
           It is possible to share dependencies between multiple packages, making the total download size for consumers of your NuGet packages smaller.
           
           Dependencies are mostly static and never change. When fixing a bug in code, the dependencies may not need to be updated. If you bundle dependencies, you end up reshipping larger packages every time. By splitting NuGet packages into related dependencies, upgrades are much more fine-grained for consumers of your package.
-          
+
   - name: NuGet.org not accessible
     questions:
-      - question: Why can't I download packages from or upload packages to NuGet.org?
+      - question: What TLS version and cipher suites does NuGet.org support?
```
joelverhagen, line 66:
```suggestion
      - question: What SSL/TLS version and cipher suites does NuGet.org support?
```

docs/nuget-org/nuget-org-faq.yml
```diff
@@ -60,33 +60,55 @@ sections:
...
         answer: |
-          First, make sure you're using the latest versions of NuGet. If that version continues to fail, [contact support](https://www.nuget.org/policies/Contact) and provide additional connection troubleshooting information including:
-          
-          - The version of NuGet you're using
+          NuGet.org supports TLS 1.2 and the following cipher suites:
+          - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+          - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+          - TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384
+          - TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
```
joelverhagen, line 72: Would it be possible to link to some of the OS-specific guidance? Like "If you are using Windows XP or Windows Server 2003, TLS 1.2 support is possible with some extra steps..."

docs/nuget-org/nuget-org-faq.yml
```diff
@@ -60,33 +60,55 @@ sections:
...
+
+      - question: Why can't I restore packages from NuGet.org?
+        answer: |
+          If you meet transient failures when restoring packages from NuGet.org, we suggest you:
+          - use [V3 API endpoint](https://learn.microsoft.com/nuget/nuget-org/overview-nuget-org#api-endpoint-for-nugetorg) to restore packages from NuGet.org.
```
joelverhagen, line 77: Do you think, at the very end, we can suggest using V2 as a workaround? I'm not happy about the idea, but we've had situations in the past where V2 worked for folks when V3 did not.

joelverhagen (approved):
🥇

JonDouglas:
@zhhyu can we close the other PR #2884 in favor of this one?

docs/nuget-org/nuget-org-faq.yml
JonDouglas: Probably should just keep TLS given SSL is deprecated?

docs/nuget-org/nuget-org-faq.yml
JonDouglas: This seems fine for now? Both these OS are deprecated anyway.

JonDouglas (approved):
Looks great. Thanks for following up on this!

docs/nuget-org/nuget-org-faq.yml
joelverhagen: I think the heading can shouldSSL but we mention SSL 1.0, 2.0, 3.0, TLS 1.0, and TLS 1.1 are disabled. Generally folks refer to TLS connections as SSL as a generalization. Include the term SSL might help Ctrl + F or Google searches.

docs/nuget-org/nuget-org-faq.yml
joelverhagen: Sure, sounds fine to me. I know that over support requests we often give more tailored guidance based on OS version.

zhhyu: I decide not to list the details of the OS-specific guidance because there are already many official documents online talking about how to configure cipher suites on different systems. Once we list the supported cipher suites, customers should be able to find documents on how to configure it and it is not our job to document the detailed steps.

The cipher suites that CDN supported before are not common ones so that's why we put some instructions when addressing support tickets.

docs/nuget-org/nuget-org-faq.yml
zhhyu: I will update the support handbook soon. How about we put the suggestion of trying V2 feed there?

The reason why I mention V3 feed here is that we have a much better control on V3 traffic. If we say that the outage scope is very limited (like ISP related) and we would not like to run the failover, we can suggest that customers use V2 feed by emails.

docs/nuget-org/nuget-org-faq.yml
```diff
@@ -60,33 +60,55 @@ sections:
...
+          - use environment variables of NuGet clients to enhance the retry policy on CI machines.
+
+          | Environment Variable | Description | Remarks |
+          | --- | --- | --- |
+          | NUGET_ENABLE_ENHANCED_HTTP_RETRY | Enables or disables enhanced HTTP retry in NuGet. | Possible values are `true` (default since NuGet CLI 6.2.0) or `false`. |
+          | NUGET_ENHANCED_MAX_NETWORK_TRY_COUNT | Configures the maximum number of times an HTTP connection should be retried when enhanced retries are enabled. | A number representing how many retries to perform, the default value is `6`. |
+          | NUGET_ENHANCED_NETWORK_RETRY_DELAY_MILLISECONDS | Configures the amount of time to wait in milliseconds before retrying an HTTP connection when enhanced retries are enabled. | Number of millseconds to wait, the default value is `1000`. |
+
+          > [!Note]
+          > These environment variables are available since [.NET CLI](https://learn.microsoft.com/nuget/reference/dotnet-commands) (.NET SDK) 6.0.100, [NuGet CLI](https://learn.microsoft.com/nuget/reference/nuget-exe-cli-reference) 6.0.0, Visual Studio 2022 version 17.0 and corresponding MSBuild version.
```
zhhyu, line 88: @zivkan Could you help me verify that these version numbers and the description here are accurate?

zivkan: It was added in NuGet 6.0: https://github.com/NuGet/NuGet.Client/commit/0a66a9486c69f95ddd5ea57158c86dfb2c9e877f

According to the [NuGet 6.0 release notes](https://learn.microsoft.com/en-us/nuget/release-notes/nuget-6.0), it was available in .NET SDK 6.0 and Visual Studio (msbuild) 17.0.

docs/nuget-org/nuget-org-faq.yml
zhhyu: I have removed the suggestion of using V3 feed in this document and will update the support handbook to mention about using V2 feed when V3 feed is under the degradation.

docs/nuget-org/nuget-org-faq.yml
zhhyu: Thanks! @zivkan Could I get an approval from you?

I have checked and followed the release notes. The description has been updated.

zhhyu:
> @zhhyu can we close the other PR #2884 in favor of this one?

Yes, I have closed that PR.

PR #3003, opened 2023-02-15
Replacement tokens apply to <files> too

ChrisGuzak:


docs/reference/nuspec.md
```diff
@@ -332,7 +332,9 @@ Specifies the minimum version of the NuGet client that can install this package,
 
 ## Replacement tokens
 
-When creating a package, the [`nuget pack` command](../reference/cli-reference/cli-ref-pack.md) replaces $-delimited tokens in the `.nuspec` file's `<metadata>` node with values that come from either a project file or the `pack` command's `-properties` switch.
+When creating a package, the [`nuget pack` command](../reference/cli-reference/cli-ref-pack.md) replaces 
+$-delimited tokens in the `.nuspec` file's `<metadata>` and `<files>` nodes with values that come from either a
+project file or the `pack` command's `-properties` switch.
```
zivkan, line 337: Can you please keep the sentence on a single line. This way if it gets modified in the future, we don't end up with unnecessary changes just to keep line lengths shorter than some arbitrary limit. It looks like you're aiming for a max line length around 115 characters.

Additionally, the new line changes break the diff engine, making it hard to see what your actual changes are. I think you added a "and `<files>`", but the diff will make it much more clear if you don't also add the new line characters.

ChrisGuzak: done.

jebriede:
Team review: Adding @heng-liu. Would you please take a look at the PR? Thanks!

ChrisGuzak:
@heng-liu ping!

nkolev92 (approved):
Did a quick sanity test. This works.

Issue #2949, opened 2022-11-30
Mention that nuget.exe string resources are not 100% localized


dominoFire:
We only ship nuget.exe with some strings localized, to avoid having a large binary file size.

It's estimated that including all localized strings in nuget.exe will increase its file size to 14 MB, more than 2x (times) the actual nuget.exe file size.

An alternative, would be guiding customers to use NuGet.Localization to have fully localized strings when using nuget.exe; and add a how-to for this workaround.

dominoFire:
This was discovered while working on https://github.com/NuGet/Home/issues/12097

PR #3013, opened 2023-02-23
Add details about nuget.exe localizaiton

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2949

I wasn't able to find a way to reference to the localization of Visual Studio. 
If someone else can find it, I'd be happy to add it.

### DIFF
--- a/docs/reference/nuget-exe-CLI-Reference.md
+++ b/docs/reference/nuget-exe-CLI-Reference.md
@@ -73,3 +73,8 @@ Developers concerned only with consuming packages, for example, need only unders
 
 > [!Note]
 > Command option names are case-insensitive. Options that are deprecated are not included in this reference, such as `NoPrompt` (replaced by `NonInteractive`) and `Verbose` (replaced by `Verbosity`).
+
+## Localization
+
+NuGet.exe's progress, warning and error messages are  translated in the same locales as Visual Studio.
+NuGet.exe ships as a single exe, and due to size considerations, only the most commonly surface messages are translated in all languages.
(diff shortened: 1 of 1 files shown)


docs/reference/nuget-exe-CLI-Reference.md
```diff
@@ -73,3 +73,8 @@ Developers concerned only with consuming packages, for example, need only unders
 
 > [!Note]
 > Command option names are case-insensitive. Options that are deprecated are not included in this reference, such as `NoPrompt` (replaced by `NonInteractive`) and `Verbose` (replaced by `Verbosity`).
+
+## Localization
+
+NuGet.exe's progress, warning and error messages are  translated in the same locales as Visual Studio.
```
erdembayar, line 79: Should we mention about environmental var which controls it?
Maybe someone using non-english OS still wants to use English error message then we need to inform them how to do.

docs/reference/nuget-exe-CLI-Reference.md
```diff
@@ -73,3 +73,8 @@ Developers concerned only with consuming packages, for example, need only unders
...
+NuGet.exe ships as a single exe, and due to size considerations, only the most commonly surface messages are translated in all languages.
```
donnie-msft, line 80:
```suggestion
NuGet.exe ships as a single exe, and due to size considerations, only the most commonly surfaced messages are translated in all languages.
```

docs/reference/nuget-exe-CLI-Reference.md
donnie-msft, line 79:
```suggestion
NuGet.exe's progress, warning and error messages are translated into the same locales as Visual Studio.
```

nkolev92: The environment variable is documented in https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-environment-variables. 

I don't expect this to be a super common scenario, so I don't think we need to add it.

jeffkl (approved):

PR #3031, opened 2023-03-10
Update NU1011 to indicate that CPM is an enterprise level feature and why floating version ranges are not recommended or allowed

jeffkl:
Add information about why floating version ranges are not recommended for a feature like CPM and a pointer to automation like Dependabot


docs/reference/errors-and-warnings/NU1011.md
```diff
@@ -15,7 +15,9 @@ f1_keywords:
 
 ### Issue
 
-PackageVersion items cannot contain floating versions.
+`<PackageVersion />` items cannot contain floating versions.  NuGet's central package management (CPM) is considered an enterprise-level feature which provides easier version management at scale as well as deterministic and secure restores.  The use of floating versions introduces the possibility for a bad package to be introduced into your build after it has been pushed to a feed.  This can lead to a situation where you made no changes in your repository but suddenly something is broken and there is no way for you to get back into a good state without removing the floating version or pushing a newer version of the package which is fixed.  Using non-floating versions means that every upgrade to a package is backed by a commit in your repository making it easy to determine what change caused the break and to revert a commit to get back into a good state.
```
nkolev92, line 18: Should we point out that transitive pinning also makes no sense with floating versions? 
That's also a pretty big factor, as with floating versions only, there's the lock files argument.

zivkan, line 18: Consider putting each sentence on a separate line. Markdown only adds paragraphs where there is a line that is entirely whitespace, and HTML ignores whitespace. By putting one sentence per line, when something gets edited, the diff is shorter. Personally, I find it easier to review as well.

erdembayar (approved):

docs/reference/errors-and-warnings/NU1011.md
```diff
@@ -15,7 +15,18 @@ f1_keywords:
...
+after it has been pushed to a feed.  This can lead to a situation where you made no changes in your repository but suddenly something is broken and there is no way for you to
+get back into a good state without removing the floating version or pushing a newer version of the package which is fixed.  Using non-floating versions means that every upgrade
+to a package is backed by a commit in your repository making it easy to determine what change caused the break and to revert a commit to get back into a good state.
+
+Features of central package management like [Transitive Pinning](../../consume-packages/Central-Package-Management.md#transitive-pinning) is designed to allow you to explicitly override the transitive versions in your graph for more control. Using a floating version as an override would lead to very undeterministic restores.  [NuGet Lock Files](../../consume-packages/package-references-in-project-files.md#enabling-the-lock-file) are another enterprise-level feature that does not work well with floating versions since the lock files will be constantly need to be regenerated.
```
nkolev92, line 24: This isn't fully correct as we discussed in the review meeting. I'll write a proposal.

nkolev92:
```suggestion
The [Transitive Pinning](../../consume-packages/Central-Package-Management.md#transitive-pinning) feature is designed to allow you to explicitly override the transitive versions in your graph for more control. Using a floating version as an override could make restores of different projects end up with different versions for the package that is supposed to be pinned, thus going against the promise of using the central version.
```

Sorry for the delay. How's this? Thinking about it more, maybe we don't need to mention lock files at all.

jeffkl: Looks good, thanks!

nkolev92 (approved):

PR #3034, opened 2023-03-23
Document new warning NU3042

dtivel:
**Please do not merge yet.  There are corresponding code changes that are still in review.**

This change adds documentation for a new warning NU3042.

CC @JonDouglas, can you please review?


nkolev92:
Where is NU3042 implemented?

It'll be great if all the warnings follow the prescribed methodology for warnings vs errors. 
https://github.com/NuGet/NuGet.Client/blob/80c3c74ca5c6d67912a19833e8cc2b10f875dc95/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs#LL21

I know a lot of the signing stuff doesn't, as lots of things are warnings in some cases, errors in others, but for true warnings we should consider it.

dtivel:
@nkolev92, NU3042 is implemented [here](https://github.com/NuGet/NuGet.Client/pull/5106/files#diff-b25a8cbd8bacaa127c1fdd8352e1165a73c121908bb35498cfc20b1588975f83R558).

From day 0, signing codes took a different approach with respect to warning vs. error, since the same NU code can be a warning or error depending on signature verification settings.  Instead, the codes are just assigned monotonically from NU3000.

docs/reference/errors-and-warnings/NU3042.md
```diff
@@ -0,0 +1,37 @@
...
+ms.reviewer: 
+f1_keywords: 
+  - "NU3042"
+---
+
+# NuGet Warning NU3042
+
+*NuGet 6.6.0+ on Linux and macOS only*
+
+<pre>The following X.509 root certificate is untrusted because it is not present in the certificate bundle at &lt;file-path&gt;.  For more information, visit https://aka.ms/nuget/NU3042.
```
nkolev92, line 17:
```suggestion
<pre>The following X.509 root certificate is untrusted because it is not present in the certificate bundle at &lt;file-path&gt;.
```

The idea with error codes would be that the warning message wouldn't need the extra link. 
NU3042 should be discoverable.

dtivel: @nkolev92, I think I resolved this satisfactorily.  Please verify.

Note that I will update the resource string in the NuGet.Client PR separately.

nkolev92 (approved):

PR #3038, opened 2023-03-30
Update the error and warnings doc to say that NoWarn can be applied on a project reference.

nkolev92:


nkolev92:
There's a client bug: https://github.com/NuGet/Home/issues/8093, which is making this annoying.

jeffkl (approved):

Issue #3041, opened 2023-04-04
Document Service API additions for restore vulnerability feature


zivkan:
As described in the design spec:
* https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md

PR #3077, opened 2023-06-16
Document server API VulnerabilityInfo resource

zivkan:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3041

### DIFF
--- a/docs/api/vulnerability-info.md
+++ b/docs/api/vulnerability-info.md
@@ -3,7 +3,7 @@ title: Vulnerability Info, NuGet API
 description: The data that allows clients to check packages for known vulnerabilities.
 author: zivkan
 ms.author: zivkan
-ms.date: 16/06/2023
+ms.date: 06/16/2023
 ms.topic: reference
 ---
 
--- a/docs/api/vulnerability-info.md
+++ b/docs/api/vulnerability-info.md
@@ -13,7 +13,7 @@ NuGet client, starting from version 6.7, can download known package vulnerabilit
 While the [package metadata resource](./registration-base-url-resource.md) also contains known vulnerability information, if an app needs to check a large number of packages for known vulnerabilities, it's much faster to download a file of known vulnerabilities and lookup locally, rather than making a large number of HTTP requests.
 For example, this enables NuGet Restore to quickly check restored packages for known vulnerabilities, which historically never downloaded package details from the package metadata resource.
 
-The API consists of at least two files, the [vulnerability index](#vulnerability-index) and one or more [vulnerability page files](#vulnerability-data).
+The API consists of at least two files, the [vulnerability index](#vulnerability-index) and one or more [vulnerability page files](#vulnerability-page).
 Known vulnerability data can be partitioned into multiple files, and the vulnerability index provides clients with information needed to cache files, and update the cache, efficiently.
 
 The resource used for building this URL is the `VulnerabilityInfo` resource found in the [service index](service-index.md).
--- a/docs/api/vulnerability-info.md
+++ b/docs/api/vulnerability-info.md
@@ -10,11 +10,28 @@ ms.topic: reference
 # Vulnerability information
 
 NuGet client, starting from version 6.7, can download known package vulnerability information to use in scenarios such as checking packages during restore operations.
+While the [package metadata resource](./registration-base-url-resource.md) also contains known vulnerability information, if an app needs to check a large number of packages for known vulnerabilities, it's much faster to download a file of known vulnerabilities and lookup locally, rather than making a large number of HTTP requests.
+For example, this enables NuGet Restore to quickly check restored packages for known vulnerabilities, which historically never downloaded package details from the package metadata resource.
 
-The API consists of at least two files, the [vulnerability index](#vulnerability-index) and one or more [vulnerability data files](#vulnerability-data).
+The API consists of at least two files, the [vulnerability index](#vulnerability-index) and one or more [vulnerability page files](#vulnerability-data).
 Known vulnerability data can be partitioned into multiple files, and the vulnerability index provides clients with information needed to cache files, and update the cache, efficiently.
 
 The resource used for building this URL is the `VulnerabilityInfo` resource found in the [service index](service-index.md).
+
+## Suggested partitioning strategy
+
+The pages listed in the vulnerability index should ideally be optimized to maximize caching, and therefore minimize updates to large files.
+This will allow clients to minimize the frequency it needs to download updates.
+
+A suggested strategy for vulnerability data partitioning is to have two pages, `base.json` and `updates.json`.
+The `base.json` file is updated periodically (for example once a month), and contains all known vulnerabilities at the time the file is regenerated.
+The `updates.json` file should contain any new advisories published since `base.json` was last regenerated.
+This will allow clients to download the large `base.json` infrequently, while the frequently changing `updates.json` file is always relatively small.
+
+NuGet clients combine the data from multiple files additively, and may load the files in any order.
+The data file schema does not allow for modification or redaction of known vulnerability from another file.
+Therefore if a server's vulnerability data source (for example the [GitHub Advisories Database](https://github.com/advisories)) modifies an existing advisory, the NuGet server must modify the page that the vulnerability information was previously reported.
+One way to achieve this with the suggested partition scheme is to treat all vulnerability modifications and deletions as a trigger to regenerate the complete `base.json` file, and empty `updates.json`.
 
 ## Versioning
 
@@ -24,18 +41,6 @@ The following `@type` values are used:
 --------------------------------- | -----
 VulnerabilityInfo/6.7.0 | The initial release
 
-## Suggested partitioning strategy
-
-The files listed in the vulnerability index should ideally be optimized to maximize caching, and therefore minimize updates to large files.
-This will allow clients to minimize the frequency it needs to download updates.
-
-A suggested strategy for vulnerability data partitioning is to have two files, `base.json` and `updates.json`.
-The `base.json` file is updated periodically (for example once a month), and contains all known vulnerabilities at the time the file is regenerated.
-The `updates.json` file should any new advisories published since `base.json` was last regenerated.
-This will allow clients to download the large `base.json` infrequently, while the frequently changing `updates.json` file is always relatively small.
-
-NuGet client does not provide a way for one file to update or remove entries from another file, so if any modifications are needed to any advisory in `base.json`, then `base.json` can be regenerated and all entries in `updates.json` merged.
-
 ## Vulnerability index
 
 The vulnerability index is a JSON array of objects with the following properties:
@@ -44,27 +49,31 @@ Name     | Type     | Required | Notes
 -------- | ---------| -------- | -----
 @name    | string   | yes      | A short name for the file, used as a cache key.
 @id      | string   | yes      | The full (absolute) URL to a vulnerability data file.
-@updated | DateTime | yes      | An ISO 8601 string representing the date and time the file was last updated, ideally with the UTC timezone.
+@updated | string   | yes      | An ISO 8601 string representing the date and time the file was last updated, ideally with the UTC timezone.
 comment  | string   | no       | A optional descriptive string.
 
 The following restrictions apply:
 
 * The index must be an array of objects with between 1 and 16 items.
+   If the server does not have any vulnerability data (zero pages), then must remove the `VulnerabilityInfo` resource from the `ServiceIndex`.
 * `@name` must be unique within the index, must be between 1 and 32 characters long, and can only use the characters `A` to `Z`, `a` to `z`, `0` to `9`, `-`, or `_`.
 * `@id` must be an absolute URL, not a relative URL.
 
-## Vulnerability data
+## Vulnerability page
 
-Vulnerability data files are a JSON object used as a dictionary.
-Property keys are package identifiers and values are an array of the following object with the following properties:
+Vulnerability page files are a JSON object used as a dictionary.
+Property keys are the lower-case package id and property values are an array of the following object with the following properties:
 
-Name     | Type   | Required | Notes
--------- | -------| -------- | -----
-severity | number | yes      | 0 means low, 1 means medium, 2 means high, 3 means critical.
-url      | string | yes      | URL where users can get more information about the vulnerability.
-versions | string | yes      | The version range that is vulnerable, using [NuGet's version range syntax](../concepts/package-versioning.md#version-ranges).
+Name     | Type    | Required | Notes
+-------- | ------- | -------- | -----
+severity | integer | yes      | 0 means low, 1 means medium, 2 means high, 3 means critical.
+url      | string  | yes      | URL where users can get more information about the vulnerability.
+versions | string  | yes      | The version range that is vulnerable, using [NuGet's version range syntax](../concepts/package-versioning.md#version-ranges).
 
-Package identifiers (the root object's keys) should be lowercase with [`String.ToLowerInvariant`](/dotnet/api/system.string.tolowerinvariant).
+Package IDs (the root object's keys) must be lowercased with [`String.ToLowerInvariant`](/dotnet/api/system.string.tolowerinvariant).
+
+The list of known vulnerabilities for a package should be sorted in descending order of the max version of the version range, followed by descending version of the min version, followed by the ascending order of the URL.
+Ranges with null min or max versions (unbounded) in a version range should be sorted to before non-null (bounded) versions.
 
 ## Samples
 
@@ -74,14 +83,14 @@ Here is a sample of a vulnerability index:
 [
     {
         "@name": "base",
-        "@id": "https://nuget.contoso.com/v3/vulnerabilities/vulnerability.base.json",
+        "@id": "https://nuget.contoso.com/v3/vulnerabilities/3bb6b300-2f74-45bc-af06-746fd21c024b.json",
         "@updated": "2023-06-01T06:14:58.4159909Z",
         "comment": "The base data for vulnerability update periodically"
     },
     {
         "@name": "update",
-        "@id": "https://nuget.contoso.com/v3/vulnerabilities/vulnerability.update.json",
-        "@updated": "2023-06-01T06:14:58.4159909Z",
+        "@id": "https://nuget.contoso.com/v3/vulnerabilities/ffd572cd-33f3-4372-8714-a9cab2e86b45.json",
+        "@updated": "2023-06-14T11:35:30.3155764Z",
         "comment": "The patch data for the vulnerability. Contains all the vulnerabilities since base was last updated."
     }
 ]
--- a/docs/api/vulnerability-info.md
+++ b/docs/api/vulnerability-info.md
@@ -1,0 +1,114 @@
+---
+title: Vulnerability Info, NuGet API
+description: The data that allows clients to check packages for known vulnerabilities.
+author: zivkan
+ms.author: zivkan
+ms.date: 16/06/2023
+ms.topic: reference
+---
+
+# Vulnerability information
+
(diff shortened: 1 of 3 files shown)


docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Versioning
+
+The following `@type` values are used:
+
+@type value                       | Notes
+--------------------------------- | -----
+VulnerabilityInfo/6.7.0 | The initial release
+
+## Suggested partitioning strategy
```
nkolev92, line 27: Can we use some of the language from https://github.com/NuGet/Home/blob/dev/proposed/2022/vulnerabilities-in-restore.md#dedicated-vulnerability-information-resource. 

In particular, I think we can be more explicit about each resource requiring at minimum 1 page and that the page content *has* to be exclusive.

nkolev92 (commented):
Looks great. I think we need a few more clarifying things.

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+
+The files listed in the vulnerability index should ideally be optimized to maximize caching, and therefore minimize updates to large files.
+This will allow clients to minimize the frequency it needs to download updates.
+
+A suggested strategy for vulnerability data partitioning is to have two files, `base.json` and `updates.json`.
```
zhhyu, line 32: Should we mention about whether the files should be compressed or not? What kind of compressions does NuGet clients support?

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+The `base.json` file is updated periodically (for example once a month), and contains all known vulnerabilities at the time the file is regenerated.
+The `updates.json` file should any new advisories published since `base.json` was last regenerated.
+This will allow clients to download the large `base.json` infrequently, while the frequently changing `updates.json` file is always relatively small.
+
+NuGet client does not provide a way for one file to update or remove entries from another file, so if any modifications are needed to any advisory in `base.json`, then `base.json` can be regenerated and all entries in `updates.json` merged.
+
+## Vulnerability index
+
+The vulnerability index is a JSON array of objects with the following properties:
```
zhhyu, line 41: I am sorry that I haven't checked what the final design is. My understanding is that one day we will have to purge very old vulnerability data as the file size will grow forever. Will it be better that we have a property to say what the time range is for vulnerability data we provide? Otherwise, customers may complain that we miss a vulnerable package as it was marked as vulnerable long time ago and we just assume that no one is using that.

docs/api/vulnerability-info.md
zivkan: As long as the server respects the the `Accept-Encoding` header on requests, then client will "Just Work". This also allows servers to adapt as new versions of .NET's runtime support newer compression formats

Note, I wasn't around when the registration resource added support for gzip, so I don't know why the gzipped versions are in a different URL to the non-gzipped versions. The URL scheme is the same (client doesn't append `.gz` to the request), so as long as the server handles requests without the `Accept-Encoding` header to request the uncompressed file.

So, I see this as a generic HTTP compliance issue, not something specific to the VulnerabilityInfo resource.

docs/api/vulnerability-info.md
zivkan: Currently new vulnerabilities in the NuGet ecosystem are growing slowly, so I hope that improvements in internet connectivity, HTTP compression algorithms, CPU speed (or at least cores, so we can do more work in parallel), and disk speed&capacity will outpace the growth of the file due to new vulnerabilities. So, I'm not sure that we will need to purge very old vulnerabilities. I think we'll be able to maintain perf for years to come, but it does depend on how quickly the file size ends up growing.

But if we want to, we can change the protocol design to cater for it. NuGetAudit is currently only available in preview versions of NuGet, but we have a very, very short time to make changes before NuGet 6.7 is locked in, at which time the protocol design will be used by GA versions of NuGet.

Another change which I wish we added initially, is to wrap all arrays (and the data file root object) in an object with a count property. This would let clients pre-allocate arrays of the correct size, reducing the number of resize events, and therefore improve performance. Something to keep in mind if we ever make other breaking changes (and therefore a new resource version), but when I do performance tracing with the current file(s), I don't see this being significant in the context of the overall restore performance.

zhhyu (approved):

docs/api/vulnerability-info.md
zhhyu: We can enrich that "comment" property with a better visibility for customers, if we have to do so one day. No protocol needs to be broken.

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+
+The files listed in the vulnerability index should ideally be optimized to maximize caching, and therefore minimize updates to large files.
+This will allow clients to minimize the frequency it needs to download updates.
+
+A suggested strategy for vulnerability data partitioning is to have two files, `base.json` and `updates.json`.
+The `base.json` file is updated periodically (for example once a month), and contains all known vulnerabilities at the time the file is regenerated.
+The `updates.json` file should any new advisories published since `base.json` was last regenerated.
+This will allow clients to download the large `base.json` infrequently, while the frequently changing `updates.json` file is always relatively small.
+
+NuGet client does not provide a way for one file to update or remove entries from another file, so if any modifications are needed to any advisory in `base.json`, then `base.json` can be regenerated and all entries in `updates.json` merged.
```
donnie-msft, line 37: You mean client would update the base.json file? I don't understand why client couldn't do this. Can you elaborate on "does not provide a way"?

zivkan: Say `base.json` has an advisory for `GHA-ABC-123`, but that advisory is edited on github after `base.json` is generated. There's no way for `update.json` to tell client "hey, remember `base.json`'s GHA-ABC-123? Delete it, or overwrite it with the contents from this file".

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Vulnerability index
+
+The vulnerability index is a JSON array of objects with the following properties:
+
+Name     | Type     | Required | Notes
+-------- | ---------| -------- | -----
+@name    | string   | yes      | A short name for the file, used as a cache key.
+@id      | string   | yes      | The full (absolute) URL to a vulnerability data file.
+@updated | DateTime | yes      | An ISO 8601 string representing the date and time the file was last updated, ideally with the UTC timezone.
```
joelverhagen, line 47: To be consistent with other APIs, the `Type` should be `string` not `DateTime`. I like the specificity here but there is no DateTime type in JSON.

Example: `published` in https://review.learn.microsoft.com/en-us/nuget/api/registration-base-url-resource?branch=pr-en-us-3077#catalog-entry

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+comment  | string   | no       | A optional descriptive string.
+
+The following restrictions apply:
+
+* The index must be an array of objects with between 1 and 16 items.
```
joelverhagen, line 52: Where is this restriction on max 16 enforced. This seems pretty weird to include in the protocol specification since it's not clear to me what goes wrong if this rule is violated (either from the lower bound or upper bound).

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+* `@name` must be unique within the index, must be between 1 and 32 characters long, and can only use the characters `A` to `Z`, `a` to `z`, `0` to `9`, `-`, or `_`.
+* `@id` must be an absolute URL, not a relative URL.
+
+## Vulnerability data
+
+Vulnerability data files are a JSON object used as a dictionary.
+Property keys are package identifiers and values are an array of the following object with the following properties:
```
joelverhagen, line 59: Using the term `identifier` is inconsistent with the rest of the docs. I recommend `package ID`.

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+
+Name     | Type   | Required | Notes
+-------- | -------| -------- | -----
+severity | number | yes      | 0 means low, 1 means medium, 2 means high, 3 means critical.
+url      | string | yes      | URL where users can get more information about the vulnerability.
+versions | string | yes      | The version range that is vulnerable, using [NuGet's version range syntax](../concepts/package-versioning.md#version-ranges).
+
+Package identifiers (the root object's keys) should be lowercase with [`String.ToLowerInvariant`](/dotnet/api/system.string.tolowerinvariant).
```
joelverhagen, line 67:
```suggestion
Package identifiers (the root object's keys) must be lowercase with [`String.ToLowerInvariant`](/dotnet/api/system.string.tolowerinvariant).
```

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+[
+    {
+        "@name": "base",
+        "@id": "https://nuget.contoso.com/v3/vulnerabilities/vulnerability.base.json",
+        "@updated": "2023-06-01T06:14:58.4159909Z",
+        "comment": "The base data for vulnerability update periodically"
+    },
+    {
+        "@name": "update",
+        "@id": "https://nuget.contoso.com/v3/vulnerabilities/vulnerability.update.json",
```
joelverhagen, line 83: I recommend putting a random component in the URL to hint to the reader that this URL should not be hard coded. We have been bitten by this false assumption in the past.

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+* `@id` must be an absolute URL, not a relative URL.
+
+## Vulnerability data
+
+Vulnerability data files are a JSON object used as a dictionary.
+Property keys are package identifiers and values are an array of the following object with the following properties:
+
+Name     | Type   | Required | Notes
+-------- | -------| -------- | -----
+severity | number | yes      | 0 means low, 1 means medium, 2 means high, 3 means critical.
```
joelverhagen, line 63:
```suggestion
severity | integer | yes      | 0 means low, 1 means medium, 2 means high, 3 means critical.
```

You are right about the type name w.r.t. to JSON naming. The type is indeed number. But in our docs we say integer. This contradicts my `DateTime` comment above in that we use a more specific type name that the official JSON term but I think it is best to be "consistently inconsistent" with the rest of the docs.

docs/api/vulnerability-info.md
joelverhagen: In the package metadata endpoint, we have some resources returning gzip and others not. Notably, `Accept-Encoding` is NOT guaranteed to be respected by the server.

On the other hand, the package content resource does not talk about gzip at all.

I think this is more about client flexibility than server respecting request headers. The server can return gzipped JSON or not and official client will handle it because it is configured to do so.

We can choose to remain vague here, knowing that server will never do anything that breaks client (hopefully, it's our job after all) and that 3rd party client implementations can figure this out pretty easily. Or we can be explicit and go into what is happening on NuGet.org in practice. I don't feel strongly personally.

docs/api/vulnerability-info.md
joelverhagen: It might be worth lightly describing how clients are meant to merge the 1-16 vuln. files. I think it means any vulnerability data present about a given package in the later dated file replaces the earlier dated file. 

This does beg the question of how a vulnerability is redacted. If `base.json` contains "Newtonsoft 13.0.1 is vulnerable", what would `update.json` contain if we want to state "oopsie, Newtonsoft 13.0.1 is not vulnerable after all", i.e. what does a "non-vulnerable" entry look like? I guess the range no longer includes that non-vulnerable version and the range is either shrunk or split into two disjoint ranges?

joelverhagen: The more I think about it, it seems to only make sense to replace the version ranges per ID as the ID level rather than merging the list of ranges per ID, but again it might be worth calling out in the doc.

docs/api/vulnerability-info.md
joelverhagen: Something like the Unix timestamp in the file name part or some random string to drive the point home

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+    {
+        "@name": "base",
+        "@id": "https://nuget.contoso.com/v3/vulnerabilities/vulnerability.base.json",
+        "@updated": "2023-06-01T06:14:58.4159909Z",
+        "comment": "The base data for vulnerability update periodically"
+    },
+    {
+        "@name": "update",
+        "@id": "https://nuget.contoso.com/v3/vulnerabilities/vulnerability.update.json",
+        "@updated": "2023-06-01T06:14:58.4159909Z",
```
joelverhagen, line 84: Maybe provide a later timestamp to better hint at what we plan on doing in the northstar?

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,114 @@
...
+description: The data that allows clients to check packages for known vulnerabilities.
+author: zivkan
+ms.author: zivkan
+ms.date: 06/16/2023
+ms.topic: reference
+---
+
+# Vulnerability information
+
+NuGet client, starting from version 6.7, can download known package vulnerability information to use in scenarios such as checking packages during restore operations.
```
joelverhagen, line 12: Do you think it would be worth mentioning why this resource would be used over vulnerability info in the package metadata resource?

joelverhagen (changes requested):
I left some minor comments about inclusion of details, typos, and samples.

docs/api/vulnerability-info.md
zivkan: I feel like the most intuitive place to document restrictions is where the file schemas are defined, which already says "The index must be an array of objects with between 1 and 16 items". Hence, it's already documented that there's a minimum of 1. About the minimum of 1 page, it's also mentioned in the introduction in the first section ("one or more").

docs/api/vulnerability-info.md
zivkan: > In the package metadata endpoint, we have some resources returning gzip and others not. Notably, Accept-Encoding is NOT guaranteed to be respected by the server.

👀 😱 

This is completely unintuitive to me, but I wasn't on the NuGet team at the time this was implemented, so I don't know why this decision was made.

docs/api/vulnerability-info.md
zivkan: If there are zero pages, client won't warn or error, but it's bad for customer performance, hence the protocol spec doesn't allow it. If there are more than 16 pages, client will show the customer a warning that their server doesn't implement the protocol correctly, and then will ignore all pages above index 16 (well, 15, using zero indexing).

I don't agree it's weird to specify expected protocol implementations without explaining what the error is. In other documents we don't say "the package ID must be lowercase, and if you don't you're going to get an HTTP 404". Clients can choose to be strict and error out, or can choose to be lenient and allow server implementations that don't follow the spec. But the client does error out, especially if the client starts off lenient and becomes stricter in a future version, at least the protocol was defined precisely from the beginning, so server implementations could have had enough information to be compliant from the beginning.

docs/api/vulnerability-info.md
joelverhagen: It was done this way because Azure Blob Storage doesn't respect the Accept-Encoding request header. It just serves whatever `Content-Encoding` is baked into the blob.

docs/api/vulnerability-info.md
joelverhagen: > I don't understand what this is trying to convey.

Sorry, I see how that was confusing since I had a typo.

If one data file has `"nuget.core": [ "[2.0.0, )" ]` as vulnerable (single range, array with one element) and another has `"nuget.core": [ "[3.0.0, )" ]` as vulnerable (single range, array with one element), the resulting set of vulnerable ranges for `NuGet.Core` would be `[ "[2.0.0, )" ]` or `[ "[3.0.0, )" ] ` depending on the precedence defined by the client (I am guessing timestamp) rather than the resulting set being both `[ "[2.0.0, )", "[3.0.0, )" ]`. That is, _replace_ rather than _merge_.

docs/api/vulnerability-info.md
joelverhagen: Sure, seems fine!

joelverhagen: Suppose there is a package source that has a small set of packages released by the feed owners (e.g. a Telerik or some other library vendor). How would they explicitly state to the client that there are no vulnerabilities (yet)? Would it be a single empty page? Or would they opt to not have the resource in the service index at all? The latter option doesn't seem right since that's indistinguishable by the client from the "we don't support that resource" case. It seems to me there may be a scenario for 0 pages if we're talking about a generic protocol, not just what NuGet.org needs (where 0 pages will never be a thing).

docs/api/vulnerability-info.md
joelverhagen: @zhhyu, I think it would be surprising to client implementors based on this current doc to ever have the comment overloaded to be used for anything more than an informational string. I agree with @zivkan's point that we can just release a new resource version (e.g. `VulnerabilityInfo/8.0.0`) with a radically different shape if needed. Or some tactical incremental change to the current protocol that may or may not be backwards compatible. The resource version concept allows us to be very creative in the future without much worry.

docs/api/vulnerability-info.md
zivkan: Ok, that makes more sense. In that case, client will see them as two different vulnerabilities, even if they have the same advisory URL.

Since we're communicating asynchronously, I'm just going to add this in case it's not apparent enough already, but this PR is adding public documentation for what is already implemented. This PR is not a design proposal for future feature work.

What do you think about my updated commit?  I tried to explain better that there is no way for one file to "update" or "revoke" a vulnerability defined in another file. All the files are additive, or concatenated, or whatever language you prefer. I'm not sure how to better describe this.

docs/api/vulnerability-info.md
joelverhagen: Update looks good. Thanks.

docs/api/vulnerability-info.md
joelverhagen: The updated commit looks good. It is clear now. 

For other, @zivkan and I chatted over Teams and the data suggests that, although the number of vulnerability changes and removals are higher than expected, it's not something to change now. We can observe other metrics as we ship this version to assess if any further protocol changes are needed.

For reference, there are 342 IDs with vulnerabilities, 24 of which have had at least one vuln. removal and 189 of which had a vulnerability added to an existing version (a lower bound on the "vuln. change" count). In short, this means that vulnerability additions to IDs that already have vulnerabilities or vulnerability changes may indeed be a significant source of data file updates in both the base and the updates blob.

docs/api/vulnerability-info.md
```diff
@@ -0,0 +1,123 @@
...
+
+A suggested strategy for vulnerability data partitioning is to have two pages, `base.json` and `updates.json`.
+The `base.json` file is updated periodically (for example once a month), and contains all known vulnerabilities at the time the file is regenerated.
+The `updates.json` file should contain any new advisories published since `base.json` was last regenerated.
+This will allow clients to download the large `base.json` infrequently, while the frequently changing `updates.json` file is always relatively small.
+
+NuGet clients combine the data from multiple files additively, and may load the files in any order.
+The data file schema does not allow for modification or redaction of known vulnerability from another file.
+Therefore if a server's vulnerability data source (for example the [GitHub Advisories Database](https://github.com/advisories)) modifies an existing advisory, the NuGet server must modify the page that the vulnerability information was previously reported.
+One way to achieve this with the suggested partition scheme is to treat all vulnerability modifications and deletions as a trigger to regenerate the complete `base.json` file, and empty `updates.json`.
```
nkolev92, line 34: this is really cool :D

docs/api/vulnerability-info.md
nkolev92: Does an empty page show more intentionality than a resource with no pages?

nkolev92 (approved):
Looks great.

joelverhagen (approved):

nkolev92 (approved):

Issue #2799, opened 2022-07-01
Describe new VS Options page for Package Source Mappings


donnie-msft:
A new VS Options UI is being added for package source mappings in https://github.com/NuGet/Home/issues/11362.
We need to highlight that and describe it in this document.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 15617693-33df-025b-26c3-2d80e4df68d3
* Version Independent ID: 15617693-33df-025b-26c3-2d80e4df68d3
* Content: [Package Source Mapping](https://docs.microsoft.com/en-us/nuget/consume-packages/package-source-mapping)
* Content Source: [docs/consume-packages/Package-Source-Mapping.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/Package-Source-Mapping.md)
* Product: **nuget**
* GitHub Login: @nkolev92
* Microsoft Alias: **nikolev**

PR #3079, opened 2023-06-20
Document the Package Source Mappings workflow in VS Options & PM UI

donnie-msft:
- A new VS Options UI is being added for package source mappings in https://github.com/NuGet/Home/issues/11362.
We need to highlight that and describe it in this document.

- The PM UI Details Pane was changed to show Package Source Mapping status in: https://github.com/NuGet/Home/issues/12586
We need to highlight that and describe it in this document.
Related to: https://github.com/NuGet/docs.microsoft.com-nuget/issues/2799


docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -32,6 +41,41 @@ For a video-based overview of the Package Source Mapping feature, consider watch
...
+2. Navigate to the `Package Source Mappings` Options Dialog.
+
+_From the Package Manager UI_
+
+- Select a package from the list to show it in the Details Pane.
+- Press the `Configure` button to open the Package Source Mappings options page.
+
+![The NuGet Package Manager window in Visual Studio showing a selected package, and a highlight around the "Package source mapping is off" status with a `Configure` button.](media/packageSourceMapping_PMUI_Status_Off_Annotated.png)
+
+_From the Visual Studio Options Dialog_
```
JonDouglas, line 56: This seems a bit confusing. Should we show this selection of Tools? Or could someone just open generic VS Options too?

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -14,6 +14,15 @@ Use Package Source Mapping along side other [best practices](..\concepts\Securit
...
 Starting with [NuGet 6.0](..\release-notes\NuGet-6.0.md), you can centrally declare which source each package in your solution should restore from in your nuget.config file.
 
+Starting with Visual Studio 17.5, you can add and remove Package Source Mappings with the Visual Studio Options Dialog
+
+### Visual Studio support
+| Visual Studio | Package Source Mapping | Support in Tools -> Options | Support in Package Manager UI |
+|-----|---------------------|---------------------|---------------------|
+| 17.0 - 17.4 | ✅ Available | ❌ Not available | ❌ Not available |
+| 17.5 | ✅ Available | ✅ Available | ❌ Not available |
+| 17.7 Preview 3| ✅ Available | ✅ Available | ✅ Status displayed |
```
JonDouglas, line 24: Great table, but I think we should simplify with a note to be consistent with all other docs. See https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#global-package-references for an example

Fine to keep it for now too

JonDouglas (approved):
Great docs additions! Added some comments for consideration.

docs/consume-packages/Package-Source-Mapping.md
donnie-msft: Yes, my point was that you can either use Tools menu or you can use PM UI with the `Configure` shortcut.
I can show it... but I thought selecting Tools was convered in another docs.

docs/consume-packages/Package-Source-Mapping.md
donnie-msft: Chatted offline - may reduce the table into a couple notes once more features have shipped

PR #3083, opened 2023-06-22
Update F1 Keywords for VS Options to point to new Package Source Mappings page

donnie-msft:


donnie-msft:
Wrong doc, moved to https://github.com/NuGet/docs.microsoft.com-nuget/pull/3084

Issue #3059, opened 2023-05-03
Add recommendations for NuGet server implementations


zivkan:
* How to support embedded icons, licenses and readmes
* How to support HttpClient's PreAuthenticate to reduce HTTP 401's.

PR #3088, opened 2023-06-29
Add NuGet Server Implementation Guide

zivkan:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3059

### DIFF
--- a/docs/api/implementation-guide.md
+++ b/docs/api/implementation-guide.md
@@ -12,7 +12,7 @@ ms.topic: conceptual
 Prior to 2018, NuGet package repositories could copy the package metadata from the `nupkg`'s [`nuspec` file](../reference/nuspec.md), and return the results unmodified in [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) requests.
 Since then, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value to developers.
 
-# Chronology
+## Chronology
 
 To assist authors of existing NuGet repositories keep up to date with NuGet's newest features, here is the chronology of the relevant features mentioned in the remainder of the document.
 
@@ -20,10 +20,10 @@ To assist authors of existing NuGet repositories keep up to date with NuGet's ne
 |--|--|
 |N/A (2013)|[A blog post explaining how to manage package owners on nuget.org](https://devblogs.microsoft.com/nuget/managing-package-owners/) clarified the owners shown on the website are the accounts that have permission to upload new versions, and therefore [the `owners` metadata in the package is ignored](#owner-field)|
 |[4.9](../release-notes/NuGet-4.9-RTM.md)|[Embedded licenses](#embedded-files)|
-|[5.3](../release-notes/NuGet-5.3.md)|[Embedded icons](#embedded-files), [Package deprecation in `RegistrationBaseUrl` (package metadata resource)](#known-vulnerability-and-deprecation-data)|
+|[5.3](../release-notes/NuGet-5.3.md)|[Embedded icons](#embedded-files) <br/> [Package deprecation in `RegistrationBaseUrl` (package metadata resource)](#known-vulnerability-and-deprecation-data)|
 |[5.8](../release-notes/NuGet-5.8.md)|[Package vulnerability information in `RegistrationsBaseUrl` (package metadata resource)](#known-vulnerability-and-deprecation-data)|
 |[5.10](../release-notes/NuGet-5.10.md)|[Embedded readme](#embedded-files)|
-|6.7|[PreAuthenticate authenticated requests](#url-structure-for-authenticated-feeds)|
+|6.7|[PreAuthenticate authenticated requests](#url-structure-for-authenticated-feeds) <br/> [`VulnerabilityInfo` resource](#known-vulnerabilities-database-vulnerabilityinfo)|
 
 ## Owner field
 
--- a/docs/api/implementation-guide.md
+++ b/docs/api/implementation-guide.md
@@ -12,24 +12,37 @@ ms.topic: conceptual
 Prior to 2019, NuGet package repositories could copy the package metadata from the `nupkg`'s [`nuspec` file](../reference/nuspec.md), and return the results unmodified in [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) requests.
 Since then, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value to developers.
 
-## URL structure for private feeds
+## URL structure for authenticated feeds
 
 As described in [the overview of the NuGet API](./overview.md), the starting URL for all NuGet server communication is [the service index](./service-index.md).
 This document contains the URLs for all other resources that NuGet clients will query.
-As of NuGet 6.7 (Visual Studio & MSBuild 17.7, and .NET SDK 7.0.400), NuGet uses [.NET's `HttpClientHandler.PreAuthenticate](/dotnet/api/system.net.http.httpclienthandler.preauthenticate), which only works when subsequent URLs are in the same virtual directory, or a subdirectory, of a URL that has previously been authenticated.
+As of NuGet 6.7 (Visual Studio & MSBuild 17.7, and .NET SDK 7.0.400), NuGet uses [.NET's `HttpClientHandler.PreAuthenticate`](/dotnet/api/system.net.http.httpclienthandler.preauthenticate), which only avoids anonymous HTTP requests  when subsequent URLs are in the same virtual directory, or a subdirectory, of a URL that has previously been authenticated.
 This will dramatically reduce the number of unauthenticated HTTP requests sent to the server, and therefore will reduce your server workload.
 
-For example, imagine the service index is located at https://pkgs.contoso.com/nuget/v3/service/index.json, but the search resource is at https://pkgs.contoso.com/nuget/v3/search.
-Since the search URL is not in the same directory as the service index, or a subdirectory, `HttpClientHandler`'s credential cache won't use the credential on the first request, forcing the server to respond with an HTTP 401 response, and then `HttpClientHandler` will re-request the URL with credentials.
-If the search resource was instead at https://pkgs.contoso.com/nuget/v3/service/search, the first request to the search resource will be authenticated since the search endpoint is in the same directory as the service index.
-Similarly, a `RegistrationBaseUrl` of https://pkgs.contoso.com/nuget/v3/registration/ will always try an unauthenticated request first, since the URL is not a subdirectory of the service index's location, but https://pkgs.contoso.com/nuget/v3/service/registration/ will be pre-authenticated since it's a subdirectory of the service index's location.
+Here are some examples:
 
-This means if you wish to host different NuGet server resources on different servers, you will need a reverse proxy to ensure all customer facing URLs meet `HttpClientHandler`'s `PreAuthenticate` requirements.
+|URL|Will PreAuthenticate?|
+|--|--|
+|https://pkgs.contoso.com/nuget/v3/feed/index.json|N/A, this is the service index.|
+|https://pkgs.contoso.com/nuget/v3/search|No, not in the same or sub-directory as the service index.|
+|https://search.pkgs.contoso.com/nuget/v3/feed/|No, not on the same host name as the service index.|
+|https://pkgs.contoso.com/nuget/v3/feed/search|Yes, in the same directory as the service index.|
+|https://pkgs.contoso.com/nuget/v3/registration/|No, not in a subdirectory of the service index.|
+|https://pkgs.contoso.com/nuget/v3/feed/registration/|Yes, in a subdirectory of the service index.|
+|https://pkgs.contoso.com/nuget/v3/{guid}/registration/|See below|
+
+In the last example, the server might have a canonical (in this example a guid) name, and have one or more aliases. 
+If the service index request was authenticated on a non-canonical URL (the "friendly" name, in our example `feed`), then no, any requests to resources under the canonical URL will not match `HttpClientHandler`'s rules for `PreAuthenticate`.
+However, if the non-canonical URL is an HTTP redirection to the canonical URL, https://pkgs.contoso.com/nuget/v3/{guid}/index.json, then this URL will be used in `HttpClientHandler`'s credential cache.
+In this case, every request to the service index will have additional latency, due to the redirection.
+
+While NuGet's V3 API was designed to work on a static file server, the search resource is the exception that always requires a dynamic web service to process requests.
+If you wish to host search, or indeed any other NuGet API resource, on different servers, in order to benefit from `HttpClientHandler`'s `PreAuthenticate`, you will need to use a reverse proxy to ensure all customer facing URLs in the service index meet the "same or subdirectory" rule.
 
 ## Embedded files
 
-Package [icons](../reference/nuspec.md#icon), [license](../reference/nuspec.md#license), and [readme](../reference/nuspec.md#readme) files can be (and is recommended to be) embedded in the package.
-These files need a URL endpoint, either extracted and put on a static file, or a URL that dynamically extracts the files from the `.nupkg` on request, so that they can be viewed without downloading the entire `nupkg`.
+Package [icons](../reference/nuspec.md#icon), [license](../reference/nuspec.md#license), and [readme](../reference/nuspec.md#readme) files can be (and are recommended to be) embedded in the package.
+These files need a URL endpoint, either extracted and put on a static file server, or a URL that dynamically extracts the files from the `.nupkg` on request, so that they can be viewed without downloading the entire `nupkg`.
 If your package repository provides package browsing and viewing package details, you can use the URLs to show customers the embedded content on your website.
 
 Finally, the [package metadata resource](./registration-base-url-resource.md) and [search resource](./search-query-service-resource.md) must contain the hosted URL in the `iconUrl`, `licenseUrl`, and/or `readmeUrl` properties of the JSON response.
@@ -62,6 +75,6 @@ If your package repository is up-sourcing from nuget.org specifically, by keepin
 
 For the `VulnerabilityInfo` resource, nuget.org is providing vulnerability data for all GitHub reviewed advisories from the [GitHub Advisories database](https://github.com/advisories), even packages that are not hosted on nuget.org.
 Therefore, if your package repository provides upsourcing from nuget.org, or is otherwise intended on being a feed that customers host all their packages on, not just their private packages, your [service index](./service-index.md) could contain a link to nuget.org's [`VulnerabilityInfo` index](./vulnerability-info.md#vulnerability-index).
-However, if your repository is intended to be available on the customer's premises, and therefore not require a connection to the public internet, the vulnerability information should be mirrored.
+However, if your repository is intended to be available on the customer's premises, and therefore not require a connection to the public internet, the vulnerability pages should be mirrored.
 
 If your package repository is hosting first-party packages, and you would like to provide vulnerability information to customers using your own feed, but don't yet have any disclosed package vulnerabilities, you should provide a [vulnerability index](./vulnerability-info.md#vulnerability-index) with one or more [vulnerability pages](./vulnerability-info.md#vulnerability-page) whose contents are an empty JSON array (`[]`).
--- a/docs/api/implementation-guide.md
+++ b/docs/api/implementation-guide.md
@@ -1,0 +1,53 @@
+---
+title: NuGet Server Implementation Guide
+description: Guidelines and recommendations to anyone implementing the NuGet Server API in their own package repository
+author: zivkan
+ms.author: zivkan
+ms.date: 07/29/2023
+ms.topic: conceptual
+---
+
+# NuGet Server Implementation Guide
+
+Prior to 2019, NuGet package repositories could copy the package metadata from the `nupkg`'s [`nuspec` file](../reference/nuspec.md), and return the results unmodified in [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) requests.
+Since then, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value to developers.
+
+## Embedded files
+
+Package [icons](../reference/nuspec.md#icon), [license](../reference/nuspec.md#license), and [readme](../reference/nuspec.md#readme) files can be (and is recommended to be) embedded in the package.
+These files need a URL endpoint, either extracted and put on a static file, or a URL that dynamically extracts the files from the `.nupkg` on request, so that they can be viewed without downloading the entire `nupkg`.
+If your package repository provides package browsing and viewing package details, you can use the URLs to show customers the embedded content on your website.
+
+Finally, the [package metadata resource](./registration-base-url-resource.md) and [search resource](./search-query-service-resource.md) must contain the hosted URL in the `iconUrl`, `licenseUrl`, and/or `readmeUrl` properties of the JSON response.
+Packages (`.nupkg` files) are immutable and must not be modified.
+
+Note that the license could be an SPDX expression, or an embedded file (but not both).
+Packages that use a license expression, when represented in search and package metadata results, can have the `licenseUrl` set to the license expression, URL encoded, and appended to the end of https://licenses.nuget.org/.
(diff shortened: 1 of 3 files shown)


docs/api/implementation-guide.md
```diff
@@ -0,0 +1,53 @@
...
+The `<owners>` field is intended to denote who published the package on a repository, and therefore who should be contacted in case of packing issues or questions.
+
+The `<owners>` field has been deprecated from the `.nuspec` file, so if packages contain this metadata, it should be ignored.
+Do not return the value of the `.nuspec` file's `<owners>` field in the `owners` property in the search resource or package metadata resource JSON response.
+
+If your repository has per-package permissions, it is recommended to report the accounts that have permissions to publish new versions in the search and package metadata resources JSON responses.
+
+## Known vulnerability and deprecation data
+
+The [Package Metadata Resource](./registration-base-url-resource.md) can contain [deprecation](./registration-base-url-resource.md#package-deprecation) and [vulnerability](./registration-base-url-resource.md#vulnerabilities) information.
```
zivkan, line 41: @theot12 looking at the search resource docs, it doesn't contain deprecation or vulnerability info. Was this an oversight or is it intentional?

@donnie-msft @martinrrm I'm not sure what the status of PM UI no longer hitting registration to get deprecation and/or vulnerability info is, when the search endpoint returns the data. I'm not even sure if either of you are the right person to talk to about the client side of this feature.  Should we be documenting anything related to the search resource here, regarding deprecation and vulnerability information?

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,53 @@
...
+This allows customers browsing packages in Visual Studio's Package Manager User Interface, or equivalent in other IDEs, to be notified of important security or maintenance issues.
+
+Additionally, in order to provide high-performance vulnerability scanning during package restore, NuGet also downloads the full list of known vulnerabilities from [the `VulnerabilityInfo` resource](./vulnerability-info.md).
+
+If your package repository is "up-sourcing" packages from another repository, in order to mirror packages in your own feed, we recommend periodically checking the original source if there is deprecation or vulnerability data, and mirror that metadata in your own repository.
+If your package repository is up-sourcing from nuget.org specifically, by keeping state of the last time you checked (a "cursor"), you can use [the `Catalog` resource](./catalog-resource.md) to efficiently check if there are any package updates for packages you're mirroring, without having to download a large number of package metadata JSON files from the upstream feed.
```
zivkan, line 47: @joelverhagen is this right?  Does nuget.org create a new catalog leaf when deprecation and/or vulnerability info changes?

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,67 @@
...
+
+For example, imagine the service index is located at https://pkgs.contoso.com/nuget/v3/service/index.json, but the search resource is at https://pkgs.contoso.com/nuget/v3/search.
+Since the search URL is not in the same directory as the service index, or a subdirectory, `HttpClientHandler`'s credential cache won't use the credential on the first request, forcing the server to respond with an HTTP 401 response, and then `HttpClientHandler` will re-request the URL with credentials.
+If the search resource was instead at https://pkgs.contoso.com/nuget/v3/service/search, the first request to the search resource will be authenticated since the search endpoint is in the same directory as the service index.
+Similarly, a `RegistrationBaseUrl` of https://pkgs.contoso.com/nuget/v3/registration/ will always try an unauthenticated request first, since the URL is not a subdirectory of the service index's location, but https://pkgs.contoso.com/nuget/v3/service/registration/ will be pre-authenticated since it's a subdirectory of the service index's location.
```
zivkan, line 25: This is very wordy, and I'm not sure it's clear.  What do you all think if I replace it with this table?

|URL|Will PreAuthenticate?|
|--|--|
|https://pkgs.contoso.com/nuget/v3/service/index.json|n/a, this is the service index|
|https://pkgs.contoso.com/nuget/v3/search|no, not in the same directory, or subdirectory, of the service index.|
|https://pkgs.contoso.com/nuget/v3/registration/|no, not in the same directory, or subdirectory, of the service index.|
|https://pkgs.contoso.com/nuget/v3/service/search|Yes, in the same directory as the service index.|
|https://pkgs.contoso.com/nuget/v3/service/registration|yes, in a subdirectory of the service index.|
|https://pkgs.contoso.com/nuget/v3/{guid}/search|Only if the service index URL redirected to https://pkgs.contoso.com/nuget/v3/{guid}/index.json. Otherwise, even if the GUID is the canonical name of the feed and all other names are aliases, since it's not the URL that `HttpClientHandler` used in its credential cache, these requests will not be pre-authenticated.|

nkolev92: I like the able better.

zivkan:
Hey @loic-sharma !  I know you've implemented a NuGet server from scratch. If you have a bit of time to review and provide feedback, that'd be great.

docs/api/implementation-guide.md
martinrrm: From what I know, we are still using registration API to get vulnerabilities and deprecation information for all the versions. Search returns this information for the latest version only.

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,67 @@
...
+# NuGet Server Implementation Guide
+
+Prior to 2019, NuGet package repositories could copy the package metadata from the `nupkg`'s [`nuspec` file](../reference/nuspec.md), and return the results unmodified in [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) requests.
+Since then, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value to developers.
+
+## URL structure for private feeds
+
+As described in [the overview of the NuGet API](./overview.md), the starting URL for all NuGet server communication is [the service index](./service-index.md).
+This document contains the URLs for all other resources that NuGet clients will query.
+As of NuGet 6.7 (Visual Studio & MSBuild 17.7, and .NET SDK 7.0.400), NuGet uses [.NET's `HttpClientHandler.PreAuthenticate](/dotnet/api/system.net.http.httpclienthandler.preauthenticate), which only works when subsequent URLs are in the same virtual directory, or a subdirectory, of a URL that has previously been authenticated.
```
loic-sharma, line 19:
```suggestion
As of NuGet 6.7 (Visual Studio & MSBuild 17.7, and .NET SDK 7.0.400), NuGet uses [.NET's `HttpClientHandler.PreAuthenticate`](/dotnet/api/system.net.http.httpclienthandler.preauthenticate), which only works when subsequent URLs are in the same virtual directory, or a subdirectory, of a URL that has previously been authenticated.
```

docs/api/implementation-guide.md
loic-sharma: FWIW, I found the text perfectly readable but a table does add visual variety. Both seem like good options :)

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,67 @@
...
+Since the search URL is not in the same directory as the service index, or a subdirectory, `HttpClientHandler`'s credential cache won't use the credential on the first request, forcing the server to respond with an HTTP 401 response, and then `HttpClientHandler` will re-request the URL with credentials.
+If the search resource was instead at https://pkgs.contoso.com/nuget/v3/service/search, the first request to the search resource will be authenticated since the search endpoint is in the same directory as the service index.
+Similarly, a `RegistrationBaseUrl` of https://pkgs.contoso.com/nuget/v3/registration/ will always try an unauthenticated request first, since the URL is not a subdirectory of the service index's location, but https://pkgs.contoso.com/nuget/v3/service/registration/ will be pre-authenticated since it's a subdirectory of the service index's location.
+
+This means if you wish to host different NuGet server resources on different servers, you will need a reverse proxy to ensure all customer facing URLs meet `HttpClientHandler`'s `PreAuthenticate` requirements.
```
loic-sharma, line 27: Is this a strict requirement? Or is this a recommendation to improve performance?

The line above indicated this was an optional performance improvement:

> This will dramatically reduce the number of unauthenticated HTTP requests sent to the server, and therefore will reduce your server workload.

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Embedded files
+
+Package [icons](../reference/nuspec.md#icon), [license](../reference/nuspec.md#license), and [readme](../reference/nuspec.md#readme) files can be (and is recommended to be) embedded in the package.
```
loic-sharma, line 31: Should this be `are`? Not sure 😅

```suggestion
Package [icons](../reference/nuspec.md#icon), [license](../reference/nuspec.md#license), and [readme](../reference/nuspec.md#readme) files can be (and are recommended to be) embedded in the package.
```

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,67 @@
...
+ms.date: 07/29/2023
+ms.topic: conceptual
+---
+
+# NuGet Server Implementation Guide
+
+Prior to 2019, NuGet package repositories could copy the package metadata from the `nupkg`'s [`nuspec` file](../reference/nuspec.md), and return the results unmodified in [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) requests.
+Since then, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value to developers.
+
+## URL structure for private feeds
```
loic-sharma, line 15: One may misunderstand "private feed" as a "custom feed that's not nuget.org" instead of "feed that requires authentication". To clear up this confusion, I'd consider starting this paragraph with something like "For feeds that require authentication, ..."

docs/api/implementation-guide.md
zivkan: It's just for performance. I've changed the wording, so hopefully it's clearer now.

docs/api/implementation-guide.md
zivkan: English is hard. Can we write docs in a formal mathematical language instead?  `&& Recommended()`

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,80 @@
...
+
+Note that the license could be an SPDX expression, or an embedded file (but not both).
+Packages that use a license expression, when represented in search and package metadata results, can have the `licenseUrl` set to the license expression, URL encoded, and appended to the end of https://licenses.nuget.org/.
+For example, https://licenses.nuget.org/Apache-2.0.
+
+## Owner field
+
+The [package manifest file (`.nuspec`)](../reference/nuspec.md) have two fields, `<authors>` and `<owners>`.
+Package authors who are packaging third-party content often put the third-party name in the `<authors>` field.
+The `<owners>` field is intended to denote who published the package on a repository, and therefore who should be contacted in case of packing issues or questions.
```
loic-sharma, line 59: Super nitpick since the field is deprecated:

```suggestion
The `<owners>` field was intended to denote who published the package on a repository, and therefore who should be contacted in case of packing issues or questions.
```

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,80 @@
...
+
+Additionally, in order to provide high-performance vulnerability scanning during package restore, NuGet also downloads the full list of known vulnerabilities from [the `VulnerabilityInfo` resource](./vulnerability-info.md).
+
+If your package repository is "up-sourcing" packages from another repository, in order to mirror packages in your own feed, we recommend periodically checking the original source if there is deprecation or vulnerability data, and mirror that metadata in your own repository.
+If your package repository is up-sourcing from nuget.org specifically, by keeping state of the last time you checked (a "cursor"), you can use [the `Catalog` resource](./catalog-resource.md) to efficiently check if there are any package updates for packages you're mirroring, without having to download a large number of package metadata JSON files from the upstream feed.
+
+For the `VulnerabilityInfo` resource, nuget.org is providing vulnerability data for all GitHub reviewed advisories from the [GitHub Advisories database](https://github.com/advisories), even packages that are not hosted on nuget.org.
```
loic-sharma, line 76:
```suggestion
For the `VulnerabilityInfo` resource, nuget.org provides vulnerability data for all GitHub reviewed advisories from the [GitHub Advisories database](https://github.com/advisories), which includes packages that are not hosted on nuget.org.
```

loic-sharma:
Is the goal of this document to make it easier for server implementors to stay up-to-date with the best practices? Will we keep adding more information to this document over time?

If yes, it may be useful to structure this document to be time-oriented like a changelog. This would make it easier for server implementors to check what's new without having to read the entire document. That said, the document's length is very reasonable so I don't feel strongly about this!

Otherwise, this doc looks great! Thanks for writing this!

docs/api/implementation-guide.md
loic-sharma: (Apparently I can't resolve my own comments?) Looks good!

docs/api/implementation-guide.md
joelverhagen: Why not put the resource name instead of the URL in the first column? (and link the resource name to the respective doc)

docs/api/implementation-guide.md
joelverhagen: Yes, it does. You can consider linking to this how-to in addition to the catalog resource: https://learn.microsoft.com/en-us/nuget/guides/api/query-for-all-published-packages

docs/api/implementation-guide.md
zivkan: @joelverhagen I don't understand the suggestion. Can you show an example?

I'm not sure if this answers your question, but one of the issues is that github limits the width it renders markdown, so if there are too many columns, or if there's too much text, everything gets wrapped. For URLs in particular, this gets harder to read.

nkolev92 (approved):

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,100 @@
...
+description: Guidelines and recommendations to anyone implementing the NuGet Server API in their own package repository
+author: zivkan
+ms.author: zivkan
+ms.date: 07/29/2023
+ms.topic: conceptual
+---
+
+# NuGet Server Implementation Guide
+
+Prior to 2018, NuGet package repositories could copy the package metadata from the `nupkg`'s [`nuspec` file](../reference/nuspec.md), and return the results unmodified in [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) requests.
```
joelverhagen, line 12: This is a pretty strong statement. The shapes are quite different and in practice data manipulation occurs even for the fields that overlap between the XML .nuspec and the JSON search response. For example, the tags returned is an array, version normalization occurs (.nuspec is the input, but is not guaranteed to be normalize), the version list in search results is not available in the .nuspec (a nuspec is a version specific thing), etc etc

Perhaps I am missing some intent of the document but it reads a bit oddly to me.

What about an opening that starts a bit broader, excludes the nuance of history, while still alluding to the fact the server data mostly comes from package metadata. How about:

> In some cases, you may want to implement your own NuGet package feed. [Many existing implementations](https://learn.microsoft.com/en-us/nuget/hosting-packages/overview) exist which allow you to host your own feed in a variety way, but the protocol between the official NuGet client software and a package feed is documented allowing you to build your own feed implementation from scratch.
>
> The protocol does evolve over time and this guide is aimed at those that wish to or already have implemented a NuGet package server.
>
> Since the initial release of the NuGet V3 protocol in 2015, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value their package consumers, beyond simply exacting metadata from hosted packages and returning the metadata in various forms. For example, the [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) endpoints contain more than just metadata found in the `nupkg`'s [`nuspec` file](../reference/nuspec.md).
> 
> Note that this guide is focused on the NuGet V3 protocol since the V2 protocol is essentially undocumented and, since 2015, the recommended protocol for NuGet client and server communication is the V3 protocol. For more information read about [protocol versioning](https://learn.microsoft.com/en-us/nuget/api/overview#versioning).

Feel free to scrap my wording but the current intro focuses on one very specific confusion point which I think is too specific for such a broad guide.

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,100 @@
...
+Since then, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value to developers.
+
+## Chronology
+
+To assist authors of existing NuGet repositories keep up to date with NuGet's newest features, here is the chronology of the relevant features mentioned in the remainder of the document.
+
+|Version|Feature|
```
joelverhagen, line 19: I think this should say "NuGet client version" since we don't really have a protocol version aside from "V3" and some resource version strings which are not guaranteed to be exactly matching the respective client version that first shipped the feature.

I think a date column would be best for the first column. That would be easiest for people to remember or refer against their own effort (e.g. "oh yeah I first implemented my Go-based server in 2021 based on Git commits, so I should look at the stuff after 2021").

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,100 @@
...
+|[5.10](../release-notes/NuGet-5.10.md)|[Embedded readme](#embedded-files)|
+|6.7|[PreAuthenticate authenticated requests](#url-structure-for-authenticated-feeds) <br/> [`VulnerabilityInfo` resource](#known-vulnerabilities-database-vulnerabilityinfo)|
+
+## Owner field
+
+Consider two of the [package manifest file (`.nuspec`)](../reference/nuspec.md) fields, `<authors>` and `<owners>`.
+Package authors who are packaging third-party content often put the third-party name in the `<authors>` field.
+The `<owners>` field was intended to denote who published the package on a repository, and therefore who should be contacted in case of packing issues or questions.
+
+This was explained in [a blog post from 2013](https://devblogs.microsoft.com/nuget/managing-package-owners/), so the `<owners>` field is considered deprecated the `.nuspec` file.
```
joelverhagen, line 34:
```suggestion
This was explained in [a blog post from 2013](https://devblogs.microsoft.com/nuget/managing-package-owners/), so the `<owners>` field is considered deprecated in the `.nuspec` file.
```

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,100 @@
...
+If your repository has per-package permissions, it is recommended to report the accounts that have permissions to publish new versions in the `owner` metadata for search and package metadata resources JSON responses.
+
+## Embedded files
+
+Package [icons](../reference/nuspec.md#icon), [license](../reference/nuspec.md#license), and [readme](../reference/nuspec.md#readme) files can be (and are recommended to be) embedded in the package.
+These files need a URL endpoint, either extracted and put on a static file server, or a URL that dynamically extracts the files from the `.nupkg` on request, so that they can be viewed without downloading the entire `nupkg`.
+If your package repository provides package browsing and viewing package details, you can use the URLs to show customers the embedded content on your website.
+
+Finally, the [package metadata resource](./registration-base-url-resource.md) and [search resource](./search-query-service-resource.md) must contain the hosted URL in the `iconUrl`, `licenseUrl`, and/or `readmeUrl` properties of the JSON response.
+Packages (`.nupkg` files) are immutable and must not be modified.
```
joelverhagen, line 47: This is not strictly true per protocol. Several feeds allow mutability today and it's more that client caches things in a couple tricky ways making package mutability problematic or at least harder to navigate for the package consumer. 

I'm okay with keeping it simple in this guide since suggesting they CAN be mutable would make it even worse, but such a blanket statement might give people the wrong impression. Maybe something like:

> NuGet.org and NuGet client software are designed around a concept of package immutability where a .nupkg for a given package version will never change. The embedded files mentioned in this section support that concept. There are however third-party package feeds which allow varying degrees of package immutability. As such, there is now hard requirement of package immutability but due to client-side caching and other factors, it is recommended that authors and feed maintainers treat packages as immutable. 

Again, feel free to scrap it but this statement might confuse users or feeds implementors which have evidence to the contrary.

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,100 @@
...
+
+Note that the license could be an SPDX expression, or an embedded file (but not both).
+Packages that use a license expression, when represented in search and package metadata results, can have the `licenseUrl` set to the license expression, URL encoded, and appended to the end of https://licenses.nuget.org/.
```
joelverhagen, line 50:
```suggestion
Packages that use a license expression, when represented in search and package metadata results, can have the `licenseUrl` set to the license expression, URL encoded, and appended to the end of https://licenses.nuget.org/. For more about NuGet.org's license content host and its protocol, see [https://learn.microsoft.com/en-us/nuget/nuget-org/licenses.nuget.org].
```

docs/api/implementation-guide.md
joelverhagen, line 19: Package types (returning them in the JSON and filtering by them) were added to search in 2019. Client does not use this but it was around 5.5.0 timeframe. .NET CLI does it it.

ID prefix reservation was implemented in mid to late 2017 and the "verified" boolean was added. 

SemVer 2.0.0 filtering was added in early 2017.

Not sure if you wanted those mentioned.

docs/api/implementation-guide.md
```diff
@@ -0,0 +1,100 @@
...
+ms.date: 07/29/2023
+ms.topic: conceptual
+---
+
+# NuGet Server Implementation Guide
+
+Prior to 2018, NuGet package repositories could copy the package metadata from the `nupkg`'s [`nuspec` file](../reference/nuspec.md), and return the results unmodified in [search](./search-query-service-resource.md) and [package metadata](./registration-base-url-resource.md) requests.
+Since then, NuGet has evolved to provide developers with a richer experience, and this requires package repositories to do additional work in order to provide the additional value to developers.
+
+## Chronology
```
joelverhagen, line 15: Maybe have a section which is "Required endpoints" where you simply link to this section: https://learn.microsoft.com/en-us/nuget/api/overview#resources-and-schema and tell the reader to refer to the "required" column.

docs/api/implementation-guide.md
joelverhagen: We were holding off on documented depr+vuln in search until client actually does the implementation. We have flexibility to modify the protocol right now on NuGet.org since client doesn't use it (well, kind of... but yeah we have flexibility right now we would lose if we document the shape as is).

docs/api/implementation-guide.md
zivkan: Other than semver2, nothing seems essential (and given how old NuGet 3.6 is, supporting semver1 only clients seems like a non-issue any longer). Client can't assume that just because nuget.org implemented something that other servers have as well, so I really find it hard to imagine any change in the future that is truly required. 🤷

docs/api/implementation-guide.md
zivkan: This is utterly perfect, and exactly why I wanted as many people as possible to review.

> SemVer 2.0.0 filtering was added in early 2017.

I wasn't familiar with this before. While all my updates should be reviewed, please pay close attention to this one, because I largely made up everything I wrote based on my best guess and searching the existing resources docs for semver.

zivkan: @joelverhagen should we also add symbols packages info?

This doc has taken me so long, I'm losing motivation. Can I ask someone from the server team to add a section on symbols after I merge this? Especially given client's knowledge of symbols is basically limited to pdbs go in a different nupkg.

docs/api/implementation-guide.md
zivkan: I made a different change that hopefully explains it well enough.  I'm particularly concerned about too much information and being technically pedantic over practical, which will make it harder for developers of multi-ecosystem package repositories (nuget + npm + pip, etc) that aren't interested in NuGet's nuances.

docs/api/implementation-guide.md
zivkan: I've spent so many hours on this document now, I really need to focus on my other work, so I copy-pasted your text. To be perfectly honest, I think this is too long of an introduction. We all know that the more we write, the sooner we lose the attention of people who need to read it, so shorter is better. But I really want to get this merged and move on to other work, so maybe someone can open a PR to improve it after this PR is merged.

docs/api/implementation-guide.md
joelverhagen: Sounds good.

joelverhagen (approved):
Sorry for dragging this out with my verbosity. The doc is looking 100% shippable. :shipit:

docs/api/implementation-guide.md
joelverhagen: Probably, yes. We need to find some good links for the symserv HTTP protocol since the NuGet protocol is just ingress. Feel free to open a docs bug and put it on my plate.

PR #3104, opened 2023-07-12
Add WarningsNotAsErrors and WarningAsErrors info to NuGetAudit docs

zivkan:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3106


nkolev92 (approved):

donnie-msft (approved):

JonDouglas (commented):
Wow you are thinking on the same wavelength! Should we add the inversion here too to address #3106 ? Basically how people can break their build on purpose.

PR #3110, opened 2023-07-24
Add NuGet Audit Mode blurb

JonDouglas:


docs/concepts/Auditing-Packages.md
```diff
@@ -74,6 +74,14 @@ If you do not want to fix the vulnerability or are unable to update or replace t
 
 If no security vulnerabilities are found, this means that packages with known vulnerabilities were not found in your package graph at the present moment of time you checked. Since the advisory database can be updated at any time, we recommend regularly checking your `dotnet restore` output and ensuring the same in your continuous integration process.
 
+### Setting a security audit mode
+
+By default, a security audit is done for top-level dependencies. In the case that you'd like to audit both top-level and transitive dependencies, you can set the `<NuGetAuditMode>` MSBuild property to the desired mode in which auditing will run. Possible values are `direct` and `all`. For example if you wanted to audit all dependencies for security advisories, you can set the following:
```
zivkan, line 79:
```suggestion
By default, a security audit is done for top-level dependencies.
In the case that you'd like to audit both top-level and transitive dependencies, you can set the `<NuGetAuditMode>` MSBuild property to the desired mode in which auditing will run.
Possible values are `direct` and `all`. For example if you wanted to audit all dependencies for security advisories, you can set the following:
```

each sentence on a new line improves PR diffs when people in the future make changes, but the rendered docs are the same (sentences in one paragraph)

zivkan (approved):

PR #3126, opened 2023-08-17
Update creating-a-package-msbuild.md for NuGet.Build.Tasks.Pack version

### LINKED ISSUE #3124: Update version number (or remove) for NuGet.Build.Tasks.Pack
The example points to `NuGet.Build.Tasks.Pack` version **5.2.0** [here in this section](https://learn.microsoft.com/en-us/nuget/create-packages/creating-a-package-msbuild#add-the-nugetbuildtaskspack-package):

```xml
<PackageReference Include="NuGet.Build.Tasks.Pack" Version="5.2.0"/>
```

However, this version is years old (and even unlisted on NuGet now), and as I was bashing my head against an issue for the last few weeks as it doesn't properly support the `PackageIcon` property recommended for use for current NuGet guidelines...

The doc here should be updated to not reference a specific version or be included with updates to the base package, which is now at **6.7.0**.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 31da7408-981b-cc54-1791-ca39dbdbc091
* Version Independent ID: 2060c9e8-7f2e-5da1-ee19-d196f38339fe
* Content: [Create a NuGet package using MSBuild](https://learn.microsoft.com/en-us/nuget/create-packages/creating-a-package-msbuild)
* Content Source: [docs/create-packages/creating-a-package-msbuild.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/create-packages/creating-a-package-msbuild.md)
* Service: **nuget**
* GitHub Login: @JonDouglas
* Microsoft Alias: **jodou**

michael-hawker:
Fixes #3124 

Update reference of including NuGet package to current version and private assets.

Page would now look like:

image

Added Note based on PR feedback to point to package on NuGet.org

### DIFF
--- a/docs/create-packages/creating-a-package-msbuild.md
+++ b/docs/create-packages/creating-a-package-msbuild.md
@@ -3,7 +3,7 @@ title: Create a NuGet package using MSBuild
 description: A detailed guide to the process of designing and creating a NuGet package using MSBuild, including key decision points like files and versioning.
 author: JonDouglas
 ms.author: jodou
-ms.date: 07/14/2022
+ms.date: 08/17/2023
 ms.topic: conceptual
 ---
 
@@ -86,7 +86,7 @@ If you are using MSBuild with a non-SDK-style project and PackageReference, add 
    ```xml
    <ItemGroup>
      <!-- ... -->
-     <PackageReference Include="NuGet.Build.Tasks.Pack" Version="5.2.0"/>
+     <PackageReference Include="NuGet.Build.Tasks.Pack" Version="6.7.0" PrivateAssets="all" Pack="false" />
      <!-- ... -->
    </ItemGroup>
    ```
--- a/docs/create-packages/creating-a-package-msbuild.md
+++ b/docs/create-packages/creating-a-package-msbuild.md
@@ -90,6 +90,9 @@ If you are using MSBuild with a non-SDK-style project and PackageReference, add 
      <!-- ... -->
    </ItemGroup>
    ```
+   
+   > [!NOTE]
+   > [Consider using the latest version of this package available.](https://www.nuget.org/packages/nuget.build.tasks.pack)
 
 2. Open a Developer command prompt (In the **Search** box, type **Developer command prompt**).
 
--- a/docs/create-packages/creating-a-package-msbuild.md
+++ b/docs/create-packages/creating-a-package-msbuild.md
@@ -86,7 +86,7 @@ If you are using MSBuild with a non-SDK-style project and PackageReference, add 
    ```xml
    <ItemGroup>
      <!-- ... -->
-     <PackageReference Include="NuGet.Build.Tasks.Pack" Version="6.7.0" PrivateAssets="all" Pack="false" />
+     <PackageReference Include="NuGet.Build.Tasks.Pack" Version="6.7.0" PrivateAssets="all" />
      <!-- ... -->
    </ItemGroup>
    ```


docs/create-packages/creating-a-package-msbuild.md
```diff
@@ -86,7 +86,7 @@ If you are using MSBuild with a non-SDK-style project and PackageReference, add
    ```xml
    <ItemGroup>
      <!-- ... -->
-     <PackageReference Include="NuGet.Build.Tasks.Pack" Version="5.2.0"/>
+     <PackageReference Include="NuGet.Build.Tasks.Pack" Version="6.7.0" PrivateAssets="all" Pack="false" />
```
nkolev92, line 89: I don't think you need `Pack="false"`. I'm not really sure what it does anyways. 

It's not really a supported PackageReference metadata. 
https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files

Also, maybe the comment on 88 can point to https://learn.microsoft.com/en-us/nuget/release-notes/ (via a relative link), and pointing out to reference the *latest* package version.

zivkan: `Pack="false"` is read by items that correspond to files, and therefore can be a way to prevent that file from being included in the package (although from memory it ALSO needs `PackagePath="??" metadata). But for PackageReference, it has no effect.

zivkan, line 89: In another 3 years, this package version is going to be very old and won't support features that the current NuGet at the time will support.  I know people love examples that can be copy-pasted. I like deterministic tools, so I don't like floating versions as a feature, so I don't recommend `Version="*"`. Are there any other alternatives?

`Version="get latest version from nuget.org"`? `Version="$(LatestNuGetVersion)"`?  Links don't work in code blocks, so before or after the code block, we could have a link to https://www.nuget.org/packages/NuGet.Build.Tasks.Pack#versions-body-tab

There's no way we're going to remember to update this doc every release. Even if we do, people will get frustrated when pack hasn't had any meaningful updates, but we have to update this version anyway. 

I can't think of any obvious best solution.

nkolev92: The docs can reference "code blocks", like here: https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk#list-package-versions 

Can they reference csproj blocks too?

JonDouglas (approved):
Seems fine to me. Not sure about the pack attribute. Could also just consider a small note such as https://learn.microsoft.com/en-us/contribute/content/markdown-reference#alerts-note-tip-important-caution-warning for the next reader who may get thrown off by the specific package version. Such as note can be as simple as "Consider using the latest version of this package available -> link to NuGet.org's listing of the package"

docs/create-packages/creating-a-package-msbuild.md
nkolev92: > But for PackageReference, it has no effect.

That's my understanding too. I've just seen it...a lot.

michael-hawker: Remove pack suggestion

```suggestion
     <PackageReference Include="NuGet.Build.Tasks.Pack" Version="6.7.0" PrivateAssets="all" />
```

michael-hawker: Added a note here as suggested by @JonDouglas

michael-hawker:
Final result:
image

nkolev92 (approved):

PR #3127, opened 2023-08-18
fix: username allows env var

lonix1:


nkolev92 (approved):

nkolev92:
@lonix1 Have you looked if there's a tracking issue for this? 
Title says fix which made me think there's an issue somewhere.

JonDouglas (approved):
While this has been done in other key/value pairs, it would probably be more read-able if there was a single note that included these cases such as:

https://learn.microsoft.com/en-us/contribute/content/markdown-reference#alerts-note-tip-important-caution-warning

Something to the effect of:

> Use an environment variable to improve your security best practices for fields like `username` and `cleartextpassword`.

PR #3138, opened 2023-09-07
Resolve OPS build warnings for insecure links

Mikejo5000:


Mikejo5000:
@JonDouglas I believe an build suggestion on insecure links (http) was switched to a warning a while ago. I switched these to https to reduce the warnings that are showing up in the Build Validation score. If we could merge these live soon, that would be great. Thanks!

docs/consume-packages/Team-Foundation-Build.md
```diff
@@ -81,7 +81,7 @@ The source code is under the `src` folder. Although our demo only uses a single
 ### Ignore files
 
 > [!Note]
-> There is currently a [known bug in the NuGet client](https://nuget.codeplex.com/workitem/4072) that causes the client to still add the `packages` folder to version control. A workaround is to disable the source control integration. In order to do that, you need a `Nuget.Config ` file in the  `.nuget` folder that is parallel to your solution. If this folder doesn't exist yet, you need to create it. In [`Nuget.Config`](../consume-packages/configuring-nuget-behavior.md), add the following content:
+> There is currently a [known bug in the NuGet client](https://blog.nuget.org/workitem/4072) that causes the client to still add the `packages` folder to version control. A workaround is to disable the source control integration. In order to do that, you need a `Nuget.Config ` file in the  `.nuget` folder that is parallel to your solution. If this folder doesn't exist yet, you need to create it. In [`Nuget.Config`](../consume-packages/configuring-nuget-behavior.md), add the following content:
```
zivkan, line 84: https://blog.nuget.org/workitem/4072 doesn't exist. It redirects to the blog's root page.

Internet archive's wayback machine has a copy of this issue: http://web.archive.org/web/20170421035254/https://nuget.codeplex.com/workitem/4072

Should we attempt to actually retain useful links?

I didn't check the other 307 modifications in this PR.  From memory there are some codeplex issues that the wayback machine hasn't indexed. Some issues were duplicated on github, but they aren't always easy to find.

While it's unlikely that anyone's going to look at most of these links, especially the release notes for OLD versions of NuGet, it still feels a shame to destroy the discoerability. Wouldn't it be better to just remove the link, rather than point it to something that still doesn't exist?

Mikejo5000: @zivkan Yes I think all the codeplex links should be removed or converted to non-link text, especially since no one is likely clicking on them anymore. That would remove the build suggestions for those links. I can plan to do this in a PR later on, but the regex replacement for that is trickier. The insecure links are build warnings, so it would be good to make this fix first.

Mikejo5000: @zivkan @JonDouglas OK to merge? Warnings are fixed. I can take a look at broken links in a later PR.

Mikejo5000: @zivkan @JonDouglas We are checking with a vendor to remove the links, since I will be on vacation the 2nd half of September. It would be good if we merged this first, if possible. Thanks.

Mikejo5000:
Closing this in favor of https://github.com/NuGet/docs.microsoft.com-nuget/pull/3145, which removes the broken links.

PR #3145, opened 2023-09-14
remove the broken links which go to codeplex

huypub:


Mikejo5000:
@zivkan @JonDouglas Closed the Insecure Link fix PR in favor of this PR, which removes the broken links that go to the decommissioned codeplex site, but retains the link text.

j-martens:
@zivkan @JonDouglas  - can you please review and approve this PR? Thank you!

JonDouglas:
Can I get some context as to why these links are being removed? Can we just update them from http to https to appease the tooling so we can hold onto the link for historical reasons? i.e. internet archive searching?

Mikejo5000:
@JonDouglas The main issue is that if we just switch to https we still have broken links that give customers 404s (aside from tooling suggestions). An alternative might be to put the entire link into code style instead of a markdown link. Would that be preferred to the solution in this PR?

nkolev92:
Given that these changes are in the release notes only, the value is primarily historical. 

I like changing the entire link into code style.

JonDouglas:
Let's just move them to code style links then. I find it a bit unnecessary to remove historical links in the first place, even if they 404. Release notes need to be left alone as much as possible.

huypub:
As suggested, I will use the code style for the broken links instead of removing them. So I will close this PR.

Issue #3175, opened 2023-11-06
Add missing packages to NuGet Client SDK docs


kartheekp-ms:
[NuGet Client SDK docs](https://learn.microsoft.com/nuget/reference/nuget-client-sdk) docs do not list all the packages that are published by the NuGet team.

The list of packages that are currently published by the NuGet Client team can be found in the [source code](https://github.com/NuGet/NuGet.Client/blob/dev/scripts/utils/EnsureAllPackagesExist.ps1) or at https://www.nuget.org/profiles/nuget. The list of packages in the source code will not match with the list of packages that have already been published by the NuGet account on nuget.org. This is because a few of the packages were either deprecated or published by the server team.

PR #3176, opened 2023-11-06
Update NuGet Client SDK docs to include all published packages

kartheekp-ms:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3175

[NuGet Client SDK docs](https://learn.microsoft.com/nuget/reference/nuget-client-sdk) docs do not list all the packages that are published by the NuGet team.

The list of packages that are currently published by the NuGet Client team can be found in the [source code](https://github.com/NuGet/NuGet.Client/blob/dev/scripts/utils/EnsureAllPackagesExist.ps1) or at https://www.nuget.org/profiles/nuget. The list of packages in the source code will not match with the list of packages that have already been published by the NuGet account on nuget.org. This is because a few of the packages were either deprecated or published by the server team.

Credit to @zivkan for raising awareness about this internally.

### DIFF
--- a/docs/reference/NuGet-Client-SDK.md
+++ b/docs/reference/NuGet-Client-SDK.md
@@ -25,7 +25,7 @@ The *NuGet Client SDK* refers to a group of NuGet packages:
 * [`NuGet.Credentials`](https://www.nuget.org/packages/NuGet.Credentials) - NuGet client's authentication models.
 * [`NuGet.DependencyResolver.Core`](https://www.nuget.org/packages/NuGet.DependencyResolver.Core) - NuGet's PackageReference dependency resolver implementation.
 * [`NuGet.Frameworks`](https://www.nuget.org/packages/NuGet.Frameworks) - NuGet's understanding of target frameworks.
-* [`NuGet.LibraryModel`](https://www.nuget.org/packages/NuGet.Frameworks) - NuGet's types and interfaces for understanding dependencies.
+* [`NuGet.LibraryModel`](https://www.nuget.org/packages/NuGet.LibraryModel) - NuGet's types and interfaces for understanding dependencies.
 * [`NuGet.Localization`](https://www.nuget.org/packages/NuGet.Localization) - NuGet localization package.
 * [`NuGet.PackageManagement`](https://www.nuget.org/packages/NuGet.PackageManagement) - NuGet Package Management functionality for Visual Studio installation flow.
 * [`NuGet.Packaging.Core`](https://www.nuget.org/packages/NuGet.Packaging.Core) - The (former home to) core data structures for `NuGet.Packaging`.
--- a/docs/reference/NuGet-Client-SDK.md
+++ b/docs/reference/NuGet-Client-SDK.md
@@ -11,13 +11,6 @@ ms.topic: conceptual
 
 The *NuGet Client SDK* refers to a group of NuGet packages:
 * [`NuGet.Indexing`](https://www.nuget.org/packages/NuGet.Indexing) - NuGet's indexing library for the Visual Studio client search functionality.
-* [`NuGet.VisualStudio.Contracts`](https://www.nuget.org/packages/NuGet.VisualStudio.Contracts) - This package contains RPC contracts for NuGet’s Visual Studio Service Broker extensibility APIs.
-* [`NuGet.VisualStudio`](https://www.nuget.org/packages/NuGet.VisualStudio) - This package provides APIs for invoking NuGet services in Visual Studio.
-* [`Microsoft.Build.NuGetSdkResolver`](https://www.nuget.org/packages/Microsoft.Build.NuGetSdkResolver) - MSBuild SDK resolver for NuGet packages.
-* [`NuGet.Build.Tasks.Console`](https://www.nuget.org/packages/NuGet.Build.Tasks.Console) - NuGet Build tasks for MSBuild and dotnet restore.
-* [`NuGet.Build.Tasks.Pack`](https://www.nuget.org/packages/NuGet.Build.Tasks.Pack) - NuGet tasks for MSBuild and dotnet pack.
-* [`NuGet.Build.Tasks`](https://www.nuget.org/packages/NuGet.Build.Tasks) - NuGet tasks for MSBuild and dotnet restore.
-* [`NuGet.CommandLine.XPlat`](https://www.nuget.org/packages/NuGet.CommandLine.XPlat) - NuGet executable wrapper for the dotnet CLI nuget functionality.
 * [`NuGet.Commands`](https://www.nuget.org/packages/NuGet.Commands) - Complete commands common to command-line and GUI NuGet clients.
 * [`NuGet.Common`](https://www.nuget.org/packages/NuGet.Common) - Common utilities and interfaces for all NuGet libraries.
 * [`NuGet.Configuration`](https://www.nuget.org/packages/NuGet.Configuration) - NuGet's configuration settings implementation.
@@ -27,7 +20,6 @@ The *NuGet Client SDK* refers to a group of NuGet packages:
 * [`NuGet.LibraryModel`](https://www.nuget.org/packages/NuGet.LibraryModel) - NuGet's types and interfaces for understanding dependencies.
 * [`NuGet.Localization`](https://www.nuget.org/packages/NuGet.Localization) - NuGet localization package.
 * [`NuGet.PackageManagement`](https://www.nuget.org/packages/NuGet.PackageManagement) - NuGet Package Management functionality for Visual Studio installation flow.
-* [`NuGet.Packaging.Core`](https://www.nuget.org/packages/NuGet.Packaging.Core) - The (former home to) core data structures for `NuGet.Packaging`.
 * [`NuGet.Packaging`](https://www.nuget.org/packages/NuGet.Packaging) - Provides a set of APIs to interact with `.nupkg` and `.nuspec` files from a stream. `NuGet.Protocol` depends on this package.
 * [`NuGet.ProjectModel`](https://www.nuget.org/packages/NuGet.ProjectModel) - NuGet's core types and interfaces for PackageReference-based restore, such as lock files, assets file and internal restore models.
 * [`NuGet.Protocol`](https://www.nuget.org/packages/NuGet.Protocol) - Provides a set of APIs interact with HTTP and file-based NuGet feeds.
--- a/docs/reference/NuGet-Client-SDK.md
+++ b/docs/reference/NuGet-Client-SDK.md
@@ -10,6 +10,7 @@ ms.topic: conceptual
 # NuGet Client SDK
 
 The *NuGet Client SDK* refers to a group of NuGet packages:
+
 * [`NuGet.Indexing`](https://www.nuget.org/packages/NuGet.Indexing) - NuGet's indexing library for the Visual Studio client search functionality.
 * [`NuGet.Commands`](https://www.nuget.org/packages/NuGet.Commands) - Complete commands common to command-line and GUI NuGet clients.
 * [`NuGet.Common`](https://www.nuget.org/packages/NuGet.Common) - Common utilities and interfaces for all NuGet libraries.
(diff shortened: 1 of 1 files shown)


docs/reference/NuGet-Client-SDK.md
```diff
@@ -10,9 +10,30 @@ ms.topic: conceptual
 # NuGet Client SDK
 
 The *NuGet Client SDK* refers to a group of NuGet packages:
-
-* [`NuGet.Protocol`](https://www.nuget.org/packages/NuGet.Protocol) - Used to interact with HTTP and file-based NuGet feeds
-* [`NuGet.Packaging`](https://www.nuget.org/packages/NuGet.Packaging) - Used to interact with NuGet packages. `NuGet.Protocol` depends on this package
+* [`NuGet.CommandLine`](https://www.nuget.org/packages/NuGet.CommandLine) - NuGet Command Line Interface.
```
zivkan, line 13: This isn't part of the SDK. It's the package that contains an exe, and if you reference the package in a project, it will either fail restore, or it won't provide any APIs.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -10,9 +10,30 @@ ms.topic: conceptual
...
+* [`NuGet.Indexing`](https://www.nuget.org/packages/NuGet.Indexing) - NuGet's indexing library for the Visual Studio client search functionality.
+* [`NuGet.VisualStudio.Contracts`](https://www.nuget.org/packages/NuGet.VisualStudio.Contracts) - This package contains RPC contracts for NuGet’s Visual Studio Service Broker extensibility APIs.
+* [`NuGet.VisualStudio`](https://www.nuget.org/packages/NuGet.VisualStudio) - This package provides APIs for invoking NuGet services in Visual Studio.
```
zivkan, line 16: These two are documented on another docs page. They're not part of the client SDK.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -10,9 +10,30 @@ ms.topic: conceptual
...
+* [`Microsoft.Build.NuGetSdkResolver`](https://www.nuget.org/packages/Microsoft.Build.NuGetSdkResolver) - MSBuild SDK resolver for NuGet packages.
+* [`NuGet.Build.Tasks.Console`](https://www.nuget.org/packages/NuGet.Build.Tasks.Console) - NuGet Build tasks for MSBuild and dotnet restore.
+* [`NuGet.Build.Tasks.Pack`](https://www.nuget.org/packages/NuGet.Build.Tasks.Pack) - NuGet tasks for MSBuild and dotnet pack.
+* [`NuGet.Build.Tasks`](https://www.nuget.org/packages/NuGet.Build.Tasks) - NuGet tasks for MSBuild and dotnet restore.
+* [`NuGet.CommandLine.XPlat`](https://www.nuget.org/packages/NuGet.CommandLine.XPlat) - NuGet executable wrapper for the dotnet CLI nuget functionality.
```
zivkan, line 21: These are not library packages, they do not provide APIs for projects to use in their apps. With the expect of the pack tasks, they're only packages due to the way that the .NET SDK is composed.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -10,9 +10,30 @@ ms.topic: conceptual
...
+* [`NuGet.Configuration`](https://www.nuget.org/packages/NuGet.Configuration) - NuGet's configuration settings implementation.
+* [`NuGet.Credentials`](https://www.nuget.org/packages/NuGet.Credentials) - NuGet client's authentication models.
+* [`NuGet.DependencyResolver.Core`](https://www.nuget.org/packages/NuGet.DependencyResolver.Core) - NuGet's PackageReference dependency resolver implementation.
+* [`NuGet.Frameworks`](https://www.nuget.org/packages/NuGet.Frameworks) - NuGet's understanding of target frameworks.
+* [`NuGet.LibraryModel`](https://www.nuget.org/packages/NuGet.LibraryModel) - NuGet's types and interfaces for understanding dependencies.
+* [`NuGet.Localization`](https://www.nuget.org/packages/NuGet.Localization) - NuGet localization package.
+* [`NuGet.PackageManagement`](https://www.nuget.org/packages/NuGet.PackageManagement) - NuGet Package Management functionality for Visual Studio installation flow.
+* [`NuGet.Packaging.Core`](https://www.nuget.org/packages/NuGet.Packaging.Core) - The (former home to) core data structures for `NuGet.Packaging`.
```
zivkan, line 31: As the description points out, this package no longer contains anything. It's entirely superceded by NuGet.Packaging. Unrelated to this PR, we've had the type forwarding for so long now, we can probably stop creating and publishing this package

zivkan (changes requested):
A bunch of these packages do not provide APIs, and therefore are not part of the NuGet Client SDK.

docs/reference/NuGet-Client-SDK.md
nkolev92: Yep, I want us to remove that package: https://github.com/NuGet/Home/issues/12495. 

I have it as a breaking change.

kartheekp-ms:
> A bunch of these packages do not provide APIs, and therefore are not part of the NuGet Client SDK.

Thanks for the feedback, Andy. For the blog post I am working on internally, I need a document that lists all the packages published by the NuGet Client team. Unfortunately, I haven't found the right place. Perhaps we need a new documentation page to list all the packages our team publishes.

zivkan:
> Perhaps we need a new documentation page to list all the packages our team publishes.

Why?  What benefit does it provide customers?

Even if you proposed to put it in a blog post, I'd say it's not beneficial, but at least blog posts don't need to be kept up to date. But a docs page creates a maintenance burden, we need to remember to update the docs when we change which packages get published, so unless it provides customer value, I don't see the point.

I'm thinking specifically about packages like NuGet.CommandLine.XPlat and NuGet.Build.Tasks, where there is no use case whatsoever that I can think of where downloading or using the package provides any benefit to customers. Honestly, there's no good reason for us to publish those packages to nuget.org either. I think it's just that our release scripts are simple (push NuGet.*.nupkg to nuget.org, without filtering out these "useless" packages that only exist because this is how we insert into the .NET SDK, but that uses a different feed, not nuget.org), and nobody found it valuable to add more logic into the release scripts to filter the packages to only the ones useful for customers.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -111,14 +125,14 @@ Read a package from a file stream using [`NuGet.Packaging`](https://www.nuget.or
 
 You can find examples and documentation for some of the API in the following blog series by Dave Glick, published 2016:
 
-- [Exploring the NuGet v3 Libraries, Part 1: Introduction and concepts](http://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-1)
-- [Exploring the NuGet v3 Libraries, Part 2: Searching for packages](http://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-2)
-- [Exploring the NuGet v3 Libraries, Part 3: Installing packages](http://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-3)
+* [Exploring the NuGet v3 Libraries, Part 1: Introduction and concepts](http://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-1)
+* [Exploring the NuGet v3 Libraries, Part 2: Searching for packages](http://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-2)
+* [Exploring the NuGet v3 Libraries, Part 3: Installing packages](http://daveaglick.com/posts/exploring-the-nuget-v3-libraries-part-3)
```
kartheekp-ms, line 130: I clicked on "Fix All Known Formatting Problems" in VS Code, which resulted in this change. I am happy to revert it if the team thinks this change should not be part of this PR.

I didn't see any difference in the rendered content between this change in the preview URL and the production URL.

[Preview URL for this section](https://review.learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk?branch=pr-en-us-3176#third-party-documentation)

[Production URL for this section](https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk#third-party-documentation)

docs/reference/NuGet-Client-SDK.md
```diff
@@ -111,14 +125,14 @@ Read a package from a file stream using [`NuGet.Packaging`](https://www.nuget.or
...
 
 > [!Note]
 > These blog posts were written shortly after the **3.4.3** version of the NuGet client SDK packages were released.
 > Newer versions of the packages may be incompatible with the information in the blog posts.
 
 Martin Björkström did a follow-up blog post to Dave Glick's blog series where he introduces a different approach on using the NuGet Client SDK to install NuGet packages:
 
-- [Revisiting the NuGet v3 Libraries](https://martinbjorkstrom.com/posts/2018-09-19-revisiting-nuget-client-libraries)
+* [Revisiting the NuGet v3 Libraries](https://martinbjorkstrom.com/posts/2018-09-19-revisiting-nuget-client-libraries)
```
kartheekp-ms, line 138: I clicked on "Fix All Known Formatting Problems" in VS Code, which resulted in this change. I am happy to revert it if the team thinks this change should not be part of this PR.

I didn't see any difference in the rendered content between this change in the preview URL and the production URL.

[Preview URL for this section](https://review.learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk?branch=pr-en-us-3176#third-party-documentation)

[Production URL for this section](https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk#third-party-documentation)

kartheekp-ms:
> Why? What benefit does it provide customers?

> Even if you proposed to put it in a blog post, I'd say it's not beneficial, but at least blog posts don't need to be kept up to date. But a docs page creates a maintenance burden, we need to remember to update the docs when we change which packages get published, so unless it provides customer value, I don't see the point.

Thank you for reviewing this PR. Your questions were valid and led me to realize that my understanding of the NuGet SDK was incorrect. I had assumed that any package published to nuget.org was part of the NuGet SDK, but this is not the case. I agree that creating a new docs page would unnecessarily increase maintenance without benefiting customers. I have accepted all your commit suggestions, and the PR is now ready for review.

 > I think it's just that our release scripts are simple (push NuGet.*.nupkg to nuget.org, without filtering out these "useless" packages that only exist because this is how we insert into the .NET SDK, but that uses a different feed, not nuget.org), and nobody found it valuable to add more logic into the release scripts to filter the packages to only the ones useful for customers.

I agree. Filed https://github.com/NuGet/Client.Engineering/issues/2582 issue.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -11,11 +11,23 @@ ms.topic: conceptual
 
 The *NuGet Client SDK* refers to a group of NuGet packages:
 
-* [`NuGet.Protocol`](https://www.nuget.org/packages/NuGet.Protocol) - Used to interact with HTTP and file-based NuGet feeds
-* [`NuGet.Packaging`](https://www.nuget.org/packages/NuGet.Packaging) - Used to interact with NuGet packages. `NuGet.Protocol` depends on this package
+* [`NuGet.Indexing`](https://www.nuget.org/packages/NuGet.Indexing) - NuGet's indexing library for the Visual Studio client search functionality.
```
nkolev92, line 14: Why do we ship this again?
Is it VS4Mac?

kartheekp-ms: Great question. [NuGet.org used by tab](https://www.nuget.org/packages/NuGet.Indexing#usedby-body-tab) says https://github.com/mono/monodevelop repo is using that package which is marked as public archive.

kartheekp-ms: Added `NuGet.Indexing` package to https://github.com/NuGet/Client.Engineering/issues/2582.

nkolev92: monodevelop is what eventually became vs 4 mac, so sounds like yes is the answer.

nkolev92 (approved):

docs/reference/NuGet-Client-SDK.md
zivkan: We have at least one person/partner, who I'm confident is not VS4M, using NuGet.Indexing: https://github.com/NuGet/Home/issues/11388

Although given the version number they reported, it sounds like they're getting the package though Arcade/Maestro, not nuget.org. Anyway, if we want to effectively remove this package as part of the client SDK, then we should be aware that it will break some people who are using it.

zivkan (approved):

docs/reference/NuGet-Client-SDK.md
nkolev92: On the contrary, I'm pretty certain those 2 work on VS4M :)

nkolev92:
Consider merging this soon so that https://github.com/NuGet/docs.microsoft.com-nuget/pull/3178 makes it live.

kartheekp-ms:
> Consider merging this soon so that #3178 makes it live.

Thank you for being aware of this PR. I was about to merge it so that the changes can go live before we publish the blog post.

PR #3179, opened 2023-11-11
Fix build validation issues

Mikejo5000:


Mikejo5000:
@JonDouglas @nkolev92 Follow up PR to October build validation fixes. OK to merge?

docs/nuget-org/Publish-a-package.md
```diff
@@ -87,7 +87,7 @@ Nuget.org has a package size limit of about 250 MB. When a package exceeding tha
 
 If such package is pushed from the command line, the following output is produced:
 
->   RequestEntityTooLarge https://www.nuget.org/api/v2/package/ 13903ms
+>   RequestEntityTooLarge https://api.nuget.org/v3/index.json 13903ms
 > error: Response status code does not indicate success: 413 (The package file exceeds the size limit. Please try again.).
```
zivkan, line 91: If you open https://api.nuget.org/v3/index.json in your browser, you'll see that https://www.nuget.org/api/v2/package is the URL for the "package publish" resource. The error message that this documentation is demonstrating will not show the service index URL (index.json), but will show the package publish URL.  So, I think we should find a different solution to fixing the doc validation issue.

````suggestion
```text
  RequestEntityTooLargehttps://www.nuget.org/api/v2/package/ 13903ms
error: Response status code does not indicate success: 413 (The package file exceeds the size limit. Please try again.).
```
````

nkolev92: I agree with Andy. 
The rest of this looks good. 

Maybe the validation should allowlist the `https://www.nuget.org/api/v2/package/ ` url, since it's not expected to be used directly. 

Right now it's probably just string matching `https://www.nuget.org/api/v2`

zivkan (approved):

Mikejo5000:
Thanks, @zivkan !

Issue #2414, opened 2021-04-29
The version doc may be misleading if read strictly top down - https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics


nkolev92:
In particular https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-basics may lead someone to believe that prerelease versions with leading `0`s are ok. 

For example `1.0.0-0123` is not valid.

StingyJack:
It's not valid for *you* maybe. It's nice that you want us to use semver, but don't force it. Sometimes we can't, or it's cheaper to not use it than to try to fix it (which is what happened when you released this breaking change in a minor version nuget update, customer had to reprint and revalidate part of a system because the component package version we told them was not the exact version  they got. Pharma companies are sticklers for this kind of thing and for good reason)

nkolev92:
@StingyJack 

`1.0.0-0123` has not been a valid version in any version of NuGet. 
There are no changes to any behavior here, just better clarifying the doc. 

- 5.10.0-preview.2 - https://nugettools.azurewebsites.net/5.10.0-preview.2.7185/parse-version?version=1.0.0-0123
- 4.9.4 - https://nugettools.azurewebsites.net/4.9.4/parse-version?version=1.0.0-0123
- 3.5.0 - https://nugettools.azurewebsites.net/3.5.0/parse-version?version=1.0.0-0123
- 2.14.0 - https://nugettools.azurewebsites.net/2.14.0/parse-version?version=1.0.0-0123
- 2.7.0 - https://nugettools.azurewebsites.net/2.7.0/parse-version?version=1.0.0-0123

PR #3190, opened 2023-12-05
Refactoring Package Versioning, add tabs for SemVer 1.0 vs SemVer 2.0.

nkolev92:
Fixes https://github.com/NuGet/docs.microsoft.com-nuget/issues/2414

- Moved the floating version section to dependency resolution and PackageReference. It is not necessarily versioning basics. It's specific to the PackageReference feature and as such was making the document busy.
- Rework the page to be Semver 2.0 first. Semver 2.0 has been supported completely for more than 6 years now.
- Remove the PR selection parts. I don't think it belongs here, plus it's documented in the dependency resolution and PR part.
- Moved the SemVer 2.0 section to the bottom, in particular to focus on what's new and how using it could make packages not available for really old clients.
- From the prerelease packages section, link to the package versioning basics part instead of copying 90% of the content. 
- Add tabs for the sorting discrepancies.

Couple of things I considered: 

- Do an include, where we just reference a file from the prerelease and versioning basics docs. 
- Removing the examples/making them tabulated similar as the sorting differences. IMO, they're taking way too much space and should arguably be documented in the specific parts instead of the generic doc.

### DIFF
--- a/docs/concepts/Package-Versioning.md
+++ b/docs/concepts/Package-Versioning.md
@@ -55,8 +55,7 @@ That said, package developers generally follow recognized naming conventions:
 - `-beta`: Beta release, typically one that is feature complete for the next planned release, but may contain known bugs.
 - `-rc`: Release candidate, typically a release that's potentially final (stable) unless significant bugs emerge.
 
-When ordering versions by precedence, NuGet chooses a version without a suffix first, then applies precedence to pre-release versions in reverse alphabetical order and treats dot notation numbers with numerical order.
-An example ordering:
+When ordering versions by precedence, NuGet follows the SemVer standard and chooses a version without a suffix first, then applies precedence to pre-release versions in reverse alphabetical order and treats dot notation numbers with numerical order.
 
 > [!Note]
 > Prerelease numbers with dot notation, as in *1.0.1-build.23*, are considered are part of the [SemVer 2.0.0](https://semver.org/spec/v2.0.0.html) standard, and as such are [only supported with NuGet 4.3.0+](#semantic-versioning-200).
--- a/docs/concepts/Package-Versioning.md
+++ b/docs/concepts/Package-Versioning.md
@@ -51,10 +51,12 @@ That said, package developers generally follow recognized naming conventions:
 - `-beta`: Beta release, typically one that is feature complete for the next planned release, but may contain known bugs.
 - `-rc`: Release candidate, typically a release that's potentially final (stable) unless significant bugs emerge.
 
-> [!Note]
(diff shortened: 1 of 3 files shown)


docs/concepts/Package-Versioning.md
```diff
@@ -51,10 +55,12 @@ That said, package developers generally follow recognized naming conventions:
 - `-beta`: Beta release, typically one that is feature complete for the next planned release, but may contain known bugs.
 - `-rc`: Release candidate, typically a release that's potentially final (stable) unless significant bugs emerge.
 
+When resolving package references and multiple package versions differ only by suffix, NuGet chooses a version without a suffix first, then applies precedence to pre-release versions in reverse alphabetical order and treats dot notation numbers with numerical order. For example, the following versions would be chosen in the exact order shown:
```
zivkan, line 58: As you probably know, I like markdown with one sentence per line.

```suggestion
When resolving package references and multiple package versions differ only by suffix, NuGet chooses a version without a suffix first, then applies precedence to pre-release versions in reverse alphabetical order and treats dot notation numbers with numerical order.
For example, the following versions would be chosen in the exact order shown:
```

zivkan, line 58: To be honest, I'm having a bit of a hard time grokking the first sentence.

Is it saying that NuGet sorts by descending SemVer 2 order, and then chooses the first (highest version)?

docs/concepts/Package-Versioning.md
```diff
@@ -68,37 +74,27 @@ When resolving package references and multiple package versions differ only by s
...
+1.0.1-alpha10
+1.0.1-aaa
+```
 
-Third-party clients:
+Note that versions such as `1.0.1-rc.10` and `1.0.1-rc.2` are not parsable by older versions of the client, and such packages with those versions won't be available for download with those clients.
 
-- JetBrains Rider
-- Paket version 5.0+
+If you use numerical suffixes with pre-release tags that might use double-digit numbers (or more), use leading zeroes as in beta01 and beta05 to ensure that they sort correctly when the numbers get larger. This recommendation only applies this schema.
```
zivkan, line 95: As you probably know, I like markdown with one sentence per line.

```suggestion
If you use numerical suffixes with pre-release tags that might use double-digit numbers (or more), use leading zeroes as in beta01 and beta05 to ensure that they sort correctly when the numbers get larger.
This recommendation only applies this schema.
```

zivkan (approved):

docs/concepts/Package-Versioning.md
zivkan: Thanks. I think it's an improvement now.  Personally I'd prefer it to just say it sorts by descending SemVer2 order. I say we should get customers comfortable with understanding what SemVer2 is, whereas the change still might be interpreted as NuGet having some "custom" sort order.

But it's also fine as it is, we can wait to see if we get any customers asking questions where it appears they're confused by this.

nkolev92: I agree with that.

I'll change it to make it explicit that it's semver, and not nuget order.

Issue #3216, opened 2024-01-20
docs on package source mapping doesn't explain what package source mapping actually is


zivkan:
Putting myself in the position of a customer who doesn't already know what PSM is, I don't believe this doc page actually explains what the feature is or why it's useful

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 15617693-33df-025b-26c3-2d80e4df68d3
* Version Independent ID: 15617693-33df-025b-26c3-2d80e4df68d3
* Content: [Package Source Mapping](https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping)
* Content Source: [docs/consume-packages/Package-Source-Mapping.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/Package-Source-Mapping.md)
* Service: **nuget**
* GitHub Login: @nkolev92
* Microsoft Alias: **nikolev**

PR #3218, opened 2024-01-22
Explain what Package Source Mapping is

zivkan:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3216

### DIFF
--- a/docs/consume-packages/Package-Source-Mapping.md
+++ b/docs/consume-packages/Package-Source-Mapping.md
@@ -11,12 +11,12 @@ f1_keywords:
 
 # Package Source Mapping
 
-By default, NuGet will search all package sources when it needs to download a package.
+Package Source Mapping is a tool that can be used to improve your supply chain security, especially if you use a mix of public and private package sources.
+
+By default, NuGet will search all configured package sources when it needs to download a package.
 When a package exists on multiple sources, it may not be deterministic which source the package will be downloaded from.
 With Package Source Mapping, you can filter, per package, which source(s) NuGet will search.
-This ensures that packages you only want to download from a specific source, or a subset of sources, will not accidentally be downloaded from a source considered untrusted for that package id.
 
-Package Source Mapping is a tool that can be used to improve your supply chain security, especially if you use a mix of public and private package sources.
 We also have suggestions for other [best practices](..\concepts\Security-Best-Practices.md) to help you fortify your supply chain against attacks.
 
 Package Source Mapping was added in [NuGet 6.0](..\release-notes\NuGet-6.0.md).
--- a/docs/consume-packages/Package-Source-Mapping.md
+++ b/docs/consume-packages/Package-Source-Mapping.md
@@ -11,6 +11,11 @@ f1_keywords:
 
 # Package Source Mapping
 
+By default, NuGet will search all package sources when it needs to download a package.
+When a package exists on multiple sources, it may not be deterministic which source the package will be downloaded from.
+With Package Source Mapping, you can filter, per package, which source(s) NuGet will search.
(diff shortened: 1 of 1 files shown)


docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -11,6 +11,11 @@ f1_keywords:
 
 # Package Source Mapping
 
+By default, NuGet will search all package sources when it needs to download a package.
+When a package exists on multiple sources, it may not be deterministic which source the package will be downloaded from.
+With Package Source Mapping, you can filter, per package, which source(s) NuGet will search.
+This ensures that packages you only want to download from a specific source, or a subset of sources, will not accidentally be downloaded from a source considered untrusted for that package id.
```
nkolev92, line 17: I think the below 2-3 paragraphs might need tweaked so that it doesn't read like we have 2-3 intros.

Please take a look and let me know if you share the same concern.

nkolev92 (approved):

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -11,14 +11,19 @@ f1_keywords:
 
 # Package Source Mapping
 
-Safeguarding your software supply chain is crucial if you use a mix of public and private package sources.
-Use Package Source Mapping along side other [best practices](..\concepts\Security-Best-Practices.md) to help you fortify your supply chain against attacks.
+By default, NuGet will search all package sources when it needs to download a package.
+When a package exists on multiple sources, it may not be deterministic which source the package will be downloaded from.
+With Package Source Mapping, you can filter, per package, which source(s) NuGet will search.
+This ensures that packages you only want to download from a specific source, or a subset of sources, will not accidentally be downloaded from a source considered untrusted for that package id.
```
donnie-msft, line 17: To me, this line is a bit redundant and uses the term "trust" which I was asked not to use in another spec.
 
```suggestion
```

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -11,14 +11,19 @@ f1_keywords:
 
 # Package Source Mapping
 
-Safeguarding your software supply chain is crucial if you use a mix of public and private package sources.
-Use Package Source Mapping along side other [best practices](..\concepts\Security-Best-Practices.md) to help you fortify your supply chain against attacks.
+By default, NuGet will search all package sources when it needs to download a package.
```
donnie-msft, line 14:
```suggestion
By default, NuGet will search all configured package sources when it needs to download a package.
```

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -11,14 +11,19 @@ f1_keywords:
...
 
-Safeguarding your software supply chain is crucial if you use a mix of public and private package sources.
-Use Package Source Mapping along side other [best practices](..\concepts\Security-Best-Practices.md) to help you fortify your supply chain against attacks.
+By default, NuGet will search all package sources when it needs to download a package.
+When a package exists on multiple sources, it may not be deterministic which source the package will be downloaded from.
+With Package Source Mapping, you can filter, per package, which source(s) NuGet will search.
+This ensures that packages you only want to download from a specific source, or a subset of sources, will not accidentally be downloaded from a source considered untrusted for that package id.
 
-Starting with [NuGet 6.0](..\release-notes\NuGet-6.0.md), you can centrally declare which source each package in your solution should restore from in your nuget.config file.
+Package Source Mapping is a tool that can be used to improve your supply chain security, especially if you use a mix of public and private package sources.
```
donnie-msft, line 19: I think moving this to the 2nd sentence would be a better intro. The 3rd sentence would then elaborate: 

> With Package Source Mapping, you can filter, per package, which source(s) NuGet will search.

donnie-msft (approved):
🎉 Thank you!

Issue #3214, opened 2024-01-18
Add NuGetSolver into Dependency resolution page


erdembayar:
https://learn.microsoft.com/en-us/nuget/concepts/dependency-resolution

PR #3219, opened 2024-01-22
Add NuGetSolver for dependency resolution

erdembayar:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3214

NuGetSolver handles all these calculations behind the scenes and resolves dependency conflicts, specifically addresses this problem.


JonDouglas:
Can you fix the template so it includes the tip on the same line? It renders wrong today.

JonDouglas (approved):
Approved so long as you fix the included file to render appropriately

PR #3229, opened 2024-02-08
Create Support Policy sections for NuGet.exe & NuGet SDK packages

donnie-msft:
Fixes: https://github.com/NuGet/Client.Engineering/issues/2660

Adapted the verbiage from the relevant blog post: https://devblogs.microsoft.com/nuget/announcing-nuget-exe-and-nuget-client-sdk-packages-support-policy-keeping-you-informed-and-secure/

### DIFF
--- a/docs/consume-packages/install-use-packages-nuget-cli.md
+++ b/docs/consume-packages/install-use-packages-nuget-cli.md
@@ -20,30 +20,6 @@ For most commands, the NuGet CLI tool uses the current directory, unless you spe
 
 For a complete list of commands and their arguments, see the [NuGet CLI reference](../reference/nuget-exe-cli-reference.md).
 
-## Support policy
-
-We are fully committed to supporting the most recent version of NuGet.exe.
-This means you can rely on us for bug fixes, updates, and enhancements exclusive to the version currently under development.
-
-NuGet.exe follows the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).
-
-Here are steps you can take to leverage the NuGet Support Policy effectively:
-
-* Use the latest version of NuGet.exe.
-* Note that we will release patches when critical security fixes are required for an LTS version of either Visual Studio or the .NET SDK.
-* Watch for unlisted NuGet.exe versions in [tools.json](../api/tools-json.md).
-
-### Security Patch Releases
-
-We will release patched versions of NuGet.exe exclusively when critical security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.
-
-All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [https://aka.ms/opensource/security/create-report].
-Also see the [security policy in the NuGet.Client repo](https://github.com/NuGet/NuGet.Client/blob/dev/SECURITY.md).
-
-### NuGet.exe unlisting
-
-We will begin to remove links to deprecated and vulnerable versions of NuGet.exe from [tools.json](../api/tools-json.md) by March 31st, 2024.
-
 ## Prerequisites
 
 Download the NuGet CLI from [nuget.org](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe). Save the *nuget.exe* file to a suitable directory, and make sure the directory is in your PATH environment variable.
--- a/docs/consume-packages/install-use-packages-nuget-cli.md
+++ b/docs/consume-packages/install-use-packages-nuget-cli.md
@@ -25,13 +25,13 @@ For a complete list of commands and their arguments, see the [NuGet CLI referenc
 We are fully committed to supporting the most recent version of NuGet.exe.
 This means you can rely on us for bug fixes, updates, and enhancements exclusive to the version currently under development.
 
-NuGet.exe follows the [Microsoft Modern Lifecycle Policy](https://learn.microsoft.com/lifecycle/policies/modern).
+NuGet.exe follows the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).
 
 Here are steps you can take to leverage the NuGet Support Policy effectively:
 
 * Use the latest version of NuGet.exe.
 * Note that we will release patches when critical security fixes are required for an LTS version of either Visual Studio or the .NET SDK.
-* Watch for unlisted NuGet.exe versions in [tool.json](https://learn.microsoft.com/nuget/api/tools-json).
+* Watch for unlisted NuGet.exe versions in [tool.json](../api/tools-json.md).
 
 ### Security Patch Releases
 
@@ -42,7 +42,7 @@ Also see the [security policy in the NuGet.Client repo](https://github.com/NuGet
 
 ### NuGet.exe unlisting
 
-We will begin to remove links to deprecated and vulnerable versions of NuGet.exe from [tool.json](https://learn.microsoft.com/nuget/api/tools-json) by March 31st, 2024.
+We will begin to remove links to deprecated and vulnerable versions of NuGet.exe from [tool.json](../api/tools-json.md) by March 31st, 2024.
 
 ## Prerequisites
 
--- a/docs/consume-packages/install-use-packages-nuget-cli.md
+++ b/docs/consume-packages/install-use-packages-nuget-cli.md
@@ -31,7 +31,7 @@ Here are steps you can take to leverage the NuGet Support Policy effectively:
 
 * Use the latest version of NuGet.exe.
 * Note that we will release patches when critical security fixes are required for an LTS version of either Visual Studio or the .NET SDK.
-* Watch for unlisted NuGet.exe versions in [tool.json](../api/tools-json.md).
+* Watch for unlisted NuGet.exe versions in [tools.json](../api/tools-json.md).
 
 ### Security Patch Releases
 
@@ -42,7 +42,7 @@ Also see the [security policy in the NuGet.Client repo](https://github.com/NuGet
 
 ### NuGet.exe unlisting
 
-We will begin to remove links to deprecated and vulnerable versions of NuGet.exe from [tool.json](../api/tools-json.md) by March 31st, 2024.
+We will begin to remove links to deprecated and vulnerable versions of NuGet.exe from [tools.json](../api/tools-json.md) by March 31st, 2024.
 
 ## Prerequisites
 
--- a/docs/consume-packages/install-use-packages-nuget-cli.md
+++ b/docs/consume-packages/install-use-packages-nuget-cli.md
@@ -20,6 +20,30 @@ For most commands, the NuGet CLI tool uses the current directory, unless you spe
 
 For a complete list of commands and their arguments, see the [NuGet CLI reference](../reference/nuget-exe-cli-reference.md).
 
+## Support policy
+
+We are fully committed to supporting the most recent version of NuGet.exe.
+This means you can rely on us for bug fixes, updates, and enhancements exclusive to the version currently under development.
+
+NuGet.exe follows the [Microsoft Modern Lifecycle Policy](https://learn.microsoft.com/lifecycle/policies/modern).
+
+Here are steps you can take to leverage the NuGet Support Policy effectively:
+
+* Use the latest version of NuGet.exe.
+* Note that we will release patches when critical security fixes are required for an LTS version of either Visual Studio or the .NET SDK.
+* Watch for unlisted NuGet.exe versions in [tool.json](https://learn.microsoft.com/nuget/api/tools-json).
+
+### Security Patch Releases
+
+We will release patched versions of NuGet.exe exclusively when critical security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.
+
+All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [https://aka.ms/opensource/security/create-report].
+Also see the [security policy in the NuGet.Client repo](https://github.com/NuGet/NuGet.Client/blob/dev/SECURITY.md).
+
+### NuGet.exe unlisting
+
+We will begin to remove links to deprecated and vulnerable versions of NuGet.exe from [tool.json](https://learn.microsoft.com/nuget/api/tools-json) by March 31st, 2024.
+
 ## Prerequisites
 
 Download the NuGet CLI from [nuget.org](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe). Save the *nuget.exe* file to a suitable directory, and make sure the directory is in your PATH environment variable.
--- a/docs/install-nuget-client-tools.md
+++ b/docs/install-nuget-client-tools.md
@@ -40,7 +40,7 @@ Package Manager Console commands work only within Visual Studio on Windows and d
 
 ## Support policy
 
-Visual Studio for Windows support policy can be found at [Visual Studio Product Lifecycle and Servicing](https://learn.microsoft.com/visualstudio/productinfo/vs-servicing).
+The Visual Studio for Windows support policy can be found at [Visual Studio Product Lifecycle and Servicing](https://learn.microsoft.com/visualstudio/productinfo/vs-servicing).
 
 
 The most recent version of NuGet.exe is fully supported and can be relied on for bug fixes, updates, and enhancements.
--- a/docs/install-nuget-client-tools.md
+++ b/docs/install-nuget-client-tools.md
@@ -42,8 +42,6 @@ Package Manager Console commands work only within Visual Studio on Windows and d
 
 The most recent version of NuGet.exe is fully supported and can be relied on for bug fixes, updates, and enhancements.
 
-Out-of-support NuGet.exe versions will be unlisted from [tools.json](./api/tools-json.md).
-
 For more information on NuGet.exe's support policy, see the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).
 
 ### Patch Releases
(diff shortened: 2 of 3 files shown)


donnie-msft:
I've replaced the Microsoft Support Policy link with https://aka.ms/lifecycle. I'm not able to find a direct link to the Modern Lifecycle page (https://learn.microsoft.com/en-us/lifecycle/policies/modern). If folks believe this is necessary, let me know and I'll track down how I can create a relative link directly there.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -34,12 +34,33 @@ You can find the source code for these packages in the [NuGet/NuGet.Client](http
 
 ## Support policy
 
-All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [https://aka.ms/opensource/security/create-report].
-Also see the [security policy in the NuGet.Client repo](https://github.com/NuGet/NuGet.Client/blob/dev/SECURITY.md).
+We are fully committed to supporting the most recent versions of the NuGet Client SDK packages.
```
nkolev92, line 37: Given our `we don't guarantee API` stability disclaimer, I'm wondering if we should tone down the enthusiasm in the SDK section? :D 
wyt @donnie-msft?

JonDouglas: Agreed. Docs aren't the place to be excited. NOR ARE GITHUB REVIEWS! jk

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -20,6 +20,30 @@ For most commands, the NuGet CLI tool uses the current directory, unless you spe
 
 For a complete list of commands and their arguments, see the [NuGet CLI reference](../reference/nuget-exe-cli-reference.md).
 
+## Support policy
+
+We are fully committed to supporting the most recent version of NuGet.exe.
```
JonDouglas, line 25: We need to rewrite these for docs. They need to be more direct and sincere. This is more blog-speak to me. That means avoiding pronouns where possible like "We are committed".

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -20,6 +20,30 @@ For most commands, the NuGet CLI tool uses the current directory, unless you spe
...
+This means you can rely on us for bug fixes, updates, and enhancements exclusive to the version currently under development.
+
+NuGet.exe follows the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).
+
+Here are steps you can take to leverage the NuGet Support Policy effectively:
+
+* Use the latest version of NuGet.exe.
+* Note that we will release patches when critical security fixes are required for an LTS version of either Visual Studio or the .NET SDK.
```
JonDouglas, line 33: simplify and make direct

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -20,6 +20,30 @@ For most commands, the NuGet CLI tool uses the current directory, unless you spe
...
 For a complete list of commands and their arguments, see the [NuGet CLI reference](../reference/nuget-exe-cli-reference.md).
 
+## Support policy
+
+We are fully committed to supporting the most recent version of NuGet.exe.
+This means you can rely on us for bug fixes, updates, and enhancements exclusive to the version currently under development.
+
+NuGet.exe follows the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).
+
+Here are steps you can take to leverage the NuGet Support Policy effectively:
```
JonDouglas, line 30: Nobody leverages a support policy per-say. It just is.

JonDouglas (commented):
We may need to rewrite some of this to be "docs-esq". Today it doesn't translate to docs yet due to needing to be a little more direct and topical rather than lofty. Remove anything that hasn't manifested yet and simplify where possible if you can! 

(Use AI if you need to make this simple)

docs/reference/NuGet-Client-SDK.md
donnie-msft: @JonDouglas & @nkolev92 - so which one are you suggesting to remove (the existing comment or the blog post comment)? The API stability is more about breaking changes, I'd assume. @zivkan would know more.

nkolev92: We don't want to remove the API stability comment, it's about the blog one.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -34,12 +34,33 @@ You can find the source code for these packages in the [NuGet/NuGet.Client](http
...
+This means you can rely on us for bug fixes, updates, and enhancements exclusive to the version currently under development.
+
+NuGet Client SDK packages follow the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).
```
zivkan, line 40: I disagree that the NuGet Client SDK packages follow the modern lifecycle policy. Our VS, .NET SDK, and nuget.exe experiences do, but I don't think the client SDK packages do.

Specifically, when I follow the link and look at the modern policy, it says "Microsoft's policy is to provide a minimum 30 days' notification when customers are required to take action in order to avoid significant degradation to the normal use of the product or service."

Since we're open source, does public commits on github constitute notifications with regards to the NuGet Client SDK?  Or would we have to put announcements in the annoncements repo and blog posts as well?

Also, what does "significant degradation" mean in the context of a package? My best guess is it means breaking changes to public APIs, since calling APIs is the only "normal use" of a package proving APIs.

But we will not provide 30 days notice of changing when we have to break an API for a security fix. Following the policy would also prevent us from making late changes (within 30 days of shipping) that require breaking changes to a public API for a non-security bug fix or new feature.

So, I think you should delete this line, to the lifecycle policy on the client SDK page.

donnie-msft: I can delete it here, but that was already stated in the Blog post which I copied it from. @JonDouglas & @kartheekp-ms do we need to follow-up with that blog post as well?

docs/consume-packages/install-use-packages-nuget-cli.md
```diff
@@ -20,6 +20,30 @@ For most commands, the NuGet CLI tool uses the current directory, unless you spe
 
 For a complete list of commands and their arguments, see the [NuGet CLI reference](../reference/nuget-exe-cli-reference.md).
 
+## Support policy
```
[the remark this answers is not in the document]
kartheekp-ms: I think https://learn.microsoft.com/nuget/install-nuget-client-tools is a better place to refer to the support policy of all the NuGet distribution vehicles.

docs/reference/NuGet-Client-SDK.md
kartheekp-ms: Thanks for bringing this up Andy. I agree with your feedback. I guess we should also update the blog post based on discussion on this thread. 

Given that NuGet SDK docs has already support policy section in the docs, how about just adding `We will release patched versions of NuGet Client SDK packages only when critical security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.` as per the blog post, https://devblogs.microsoft.com/nuget/announcing-nuget-exe-and-nuget-client-sdk-packages-support-policy-keeping-you-informed-and-secure/#security-patch-releases.

I think we should also mention about NuGet SDK package deprecation in the support policy as mentioned in the blog post https://devblogs.microsoft.com/nuget/announcing-nuget-exe-and-nuget-client-sdk-packages-support-policy-keeping-you-informed-and-secure/#package-deprecation

zivkan: > We will release patched versions of NuGet Client SDK packages only when critical security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.

We often publish packages any time we service an already GA version of NuGet, even if it doesn't have security fixes. For example, the P0 bug on VS17.9 is going to create new packages, which I guess we might publish packages for, but it's not a security fix.

I don't know what good wording is, but the end result is that our packages are a byproduct of the work that we do support. We'll fix security bug, sure. But anything else, we're not going to prioritize. If a customer can show us a VS, .NET SDK or nuget.exe scenario that hits the bug, it has a chance of being worked on. But if it only reproduces by writing a custom app that uses the NuGet SDK packages directly, then it's not going to be scheduled in a sprint.

docs/consume-packages/install-use-packages-nuget-cli.md
donnie-msft: This line is directly from the blog post. I'll rewrite it.

docs/consume-packages/install-use-packages-nuget-cli.md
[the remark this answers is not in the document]
donnie-msft: Moved to install-nuget-client-tools.md

docs/reference/NuGet-Client-SDK.md
```diff
@@ -33,12 +33,24 @@ You can find the source code for these packages in the [NuGet/NuGet.Client](http
...
+Out-of-support NuGet Client SDK packages will be [deprecated on nuget.org](../nuget-org/Deprecate-packages.md#client-experience-for-deprecated-packages).
+
+### Patch Releases
+
+Patched versions of NuGet Client SDK will be released exclusively when critical bugs or security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.
+
+All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [https://aka.ms/opensource/security/create-report].
 Also, see the [security policy in the NuGet.Client repo](https://github.com/NuGet/NuGet.Client/blob/dev/SECURITY.md).
 
-We do not guarantee API stability, as our team's responsibility is tooling, not libraries.
```
nkolev92, line 40: I think we should keep this. 
While we'd make bug fixes, we still stand behind that long term we don't guarantee API stability.

nkolev92 (commented):
Thanks for digesting the discussion + blog into something a lot more straightforward. 

Only feedback I have is around adding back the don't guarantee API stability comment.

docs/install-nuget-client-tools.md
```diff
@@ -38,6 +38,25 @@ The [MSBuild CLI](reference/msbuild-targets.md) also restores and creates packag
...
+
+Out-of-support NuGet.exe versions will be unlisted from [tools.json](./api/tools-json.md).
+
+For more information on NuGet.exe's support policy, see the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).
+
+### Patch Releases
+
+Patched versions of NuGet.exe will be released exclusively when critical security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.
+
+All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [https://aka.ms/opensource/security/create-report].
```
kartheekp-ms, line 53:
```suggestion
All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [MSRC's report page](https://aka.ms/opensource/security/create-report).
```

Inspired by https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk#support-policy

docs/install-nuget-client-tools.md
```diff
@@ -38,6 +38,25 @@ The [MSBuild CLI](reference/msbuild-targets.md) also restores and creates packag
 
 Package Manager Console commands work only within Visual Studio on Windows and don't work within other PowerShell environments.
 
+## Support policy
```
kartheekp-ms, line 41: I think the support policy should cover other NuGet distribution vehicles such as VS, .NET SDK, NuGet Client SDK (pointing to it's the dedicated support policy link) also in addition to `nuget.exe`. Please refer to https://devblogs.microsoft.com/nuget/announcing-nuget-exe-and-nuget-client-sdk-packages-support-policy-keeping-you-informed-and-secure/#what-you-can-expect

docs/reference/NuGet-Client-SDK.md
```diff
@@ -33,12 +33,24 @@ You can find the source code for these packages in the [NuGet/NuGet.Client](http
...
-See the [NuGet SDK documentation in the NuGet.Client repo](https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md) for more information.
+### Package Deprecation
+
+As of January 31, 2024, older versions of NuGet Client SDK packages that are not tied to an LTS version of either Visual Studio or .NET have been deprecated.
```
kartheekp-ms, line 51: Unfortunately, this work was partially done. We can certainly complete package deprecation by end of May 2024 at the latest.

docs/reference/NuGet-Client-SDK.md
kartheekp-ms: Donnie added `Patched versions of NuGet Client SDK will be released exclusively when critical bugs or security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.` statement in the latest commit. I think it addresses this comment. Thanks for resolving this comment, Donnie.

docs/reference/NuGet-Client-SDK.md
nkolev92: Hmm...maybe we remove this? 
Since this would become a JDI.

donnie-msft: Removed the date entirely

docs/install-nuget-client-tools.md
donnie-msft: Could you specify what you want said? According to Andy, the blog is wrong about the Modern Lifecycle Policy applying to the SDK. I already removed that.

docs/reference/NuGet-Client-SDK.md
```diff
@@ -33,12 +33,24 @@ You can find the source code for these packages in the [NuGet/NuGet.Client](http
...
-All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [MSRC's report page](https://aka.ms/opensource/security/create-report).
+The recommendation is to use the latest versions of NuGet Client SDK packages, and to examine your project for dependencies on deprecated NuGet Client SDK packages.
+
+Out-of-support NuGet Client SDK packages will be [deprecated on nuget.org](../nuget-org/Deprecate-packages.md#client-experience-for-deprecated-packages).
+
+### Patch Releases
+
+Patched versions of NuGet Client SDK will be released exclusively when critical bugs or security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.
+
+All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [https://aka.ms/opensource/security/create-report].
```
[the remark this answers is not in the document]
donnie-msft: I accidentally copied over it. Restored now.

docs/install-nuget-client-tools.md
donnie-msft: I accidentally copied over it. Restored now.

docs/reference/NuGet-Client-SDK.md
donnie-msft: Restored

docs/install-nuget-client-tools.md
```diff
@@ -38,6 +38,25 @@ The [MSBuild CLI](reference/msbuild-targets.md) also restores and creates packag
...
+### Patch Releases
+
+Patched versions of NuGet.exe will be released exclusively when critical security fixes are required for a long-term support (LTS) version of Visual Studio or .NET SDK.
+
+All security bugs should be reported to the Microsoft Security Response Center (MSRC) at [MSRC's report page](https://aka.ms/opensource/security/create-report).
+Also, see the [security policy in the NuGet.Client repo](https://github.com/NuGet/NuGet.Client/blob/dev/SECURITY.md).
+
+### NuGet.exe unlisting
+
+Deprecated and vulnerable versions of NuGet.exe will be removed from [tools.json](./api/tools-json.md) in 2024.
```
nkolev92, line 58: This and like 45 are saying similar things.
Should they be combined?

docs/reference/NuGet-Client-SDK.md
```diff
@@ -33,13 +33,28 @@ You can find the source code for these packages in the [NuGet/NuGet.Client](http
 > For documentation on the NuGet server protocol, please refer to the [NuGet Server API](~/api/overview.md).
 
 ## Support policy
+The most recent version of NuGet Client SDK is fully supported and can be relied on for bug fixes, updates, and enhancements.
+
+The recommendation is to use the latest versions of NuGet Client SDK packages, and to examine your project for dependencies on deprecated NuGet Client SDK packages.
+
+Out-of-support NuGet Client SDK packages will be [deprecated on nuget.org](../nuget-org/Deprecate-packages.md#client-experience-for-deprecated-packages).
+
+### Patch Releases
```
nkolev92, line 42: Does this paragraph need to be moved to line 51?

I think the API stability stuff should be in the support policy.

docs/install-nuget-client-tools.md
donnie-msft: Doesn't this sentence explain that for nuget.exe?

> For more information on NuGet.exe's support policy, see the [Microsoft Modern Lifecycle Policy](https://aka.ms/lifecycle).

docs/reference/NuGet-Client-SDK.md
nkolev92: I see this was resolved, but I didn't see the paragraph moved, did you miss a push?

docs/install-nuget-client-tools.md
kartheekp-ms: My suggestion is to cover all the distribution vehicles here not only NuGet.exe. NuGet Client tools are distributed through the VS, .NET SDK, NuGet.exe and NuGet SDK. Instead of covering NuGet.exe we should also add a link to support policy of VS, .NET SDK, NuGet Client SDK along with NuGet.exe. Please refer to the table in the blog post https://devblogs.microsoft.com/nuget/announcing-nuget-exe-and-nuget-client-sdk-packages-support-policy-keeping-you-informed-and-secure/#what-you-can-expect

donnie-msft: Please see what you think of my current change.

JonDouglas:
Should this be its own separate doc? Seems like it shouldn't distract this core doc about installing.

nkolev92 (approved):

PR #3240, opened 2024-02-16
document security best practices for managing credentials.

kartheekp-ms:
Fixes: https://github.com/NuGet/Client.Engineering/issues/2583

Updated the documentation to include a warning about setting credentials in plain text and raise awareness about the lack of encrypted password support in non-Windows platforms.

After merging this PR, I will create a follow-up PR in the dotnet docs repo, adding similar warnings to the [dotnet nuget add source](https://learn.microsoft.com/dotnet/core/tools/dotnet-nuget-add-source) & [dotnet nuget update source](https://learn.microsoft.com/dotnet/core/tools/dotnet-nuget-update-source) docs.

### DIFF
--- a/docs/concepts/Security-Best-Practices.md
+++ b/docs/concepts/Security-Best-Practices.md
@@ -121,6 +121,8 @@ When using multiple public & private NuGet source feeds, a package can be downlo
 
 For more information to secure your package feeds, see [3 Ways to Mitigate Risk When Using Private Package Feeds](https://azure.microsoft.com/resources/3-ways-to-mitigate-risk-using-private-package-feeds/).
 
+When using a private feed, refer to the [security best practices for managing credentials](../consume-packages/consuming-packages-authenticated-feeds.md#security-best-practices-for-managing-credentials).
+
 ### Client trust policies
 
 **📦 Package Consumer**
--- a/docs/concepts/Security-Best-Practices.md
+++ b/docs/concepts/Security-Best-Practices.md
@@ -121,7 +121,7 @@ When using multiple public & private NuGet source feeds, a package can be downlo
 
 For more information to secure your package feeds, see [3 Ways to Mitigate Risk When Using Private Package Feeds](https://azure.microsoft.com/resources/3-ways-to-mitigate-risk-using-private-package-feeds/).
 
-When using a private feed, refer to the [security best practices for managing credentials](../consume-packages/consuming-packages-authenticated-feeds.md#security-best-practices-for-managing-credentials) for guidance.
+When using a private feed, refer to the [security best practices for managing credentials](../consume-packages/consuming-packages-authenticated-feeds.md#security-best-practices-for-managing-credentials).
 
 ### Client trust policies
 
--- a/docs/concepts/Security-Best-Practices.md
+++ b/docs/concepts/Security-Best-Practices.md
@@ -119,7 +119,7 @@ For more information about Dependabot alerts & security updates, [see the follow
 
 When using multiple public & private NuGet source feeds, a package can be downloaded from any of the feeds. To ensure your build is predictable and secure from known attacks such as [Dependency Confusion](https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610), knowing what specific feed(s) your packages are coming from is a best practice. You can use a single feed or private feed with upstreaming capabilities for protection.
 
-For more information to secure your package feeds, see [3 Ways to Mitigate Risk When Using Private Package Feeds](https://azure.microsoft.com/resources/3-ways-to-mitigate-risk-using-private-package-feeds/).
+For more information to secure your package feeds, see [3 Ways to Mitigate Risk When Using Private Package Feeds](https://azure.microsoft.com/resources/3-ways-to-mitigate-risk-using-private-package-feeds/en-us/).
 
 When using a private feed, refer to the [security best practices for managing credentials](../consume-packages/consuming-packages-authenticated-feeds.md#security-best-practices-for-managing-credentials).
 
--- a/docs/concepts/Security-Best-Practices.md
+++ b/docs/concepts/Security-Best-Practices.md
@@ -121,7 +121,7 @@ When using multiple public & private NuGet source feeds, a package can be downlo
 
 For more information to secure your package feeds, see [3 Ways to Mitigate Risk When Using Private Package Feeds](https://azure.microsoft.com/resources/3-ways-to-mitigate-risk-using-private-package-feeds/).
 
-When using a private feed, refer to the [security best practices for managing credentials](../consume-packages/consuming-packages-authenticated-feeds.md#security-best-practices-for-managing-credentials).
+When using a private feed, refer to the [security best practices for managing credentials](../consume-packages/consuming-packages-authenticated-feeds.md#security-best-practices-for-managing-credentials) for guidance.
 
 ### Client trust policies
 
--- a/docs/consume-packages/consuming-packages-authenticated-feeds.md
+++ b/docs/consume-packages/consuming-packages-authenticated-feeds.md
@@ -100,7 +100,7 @@ For more information about valid authentication types, see [the docs on package 
 See [the *nuget.config* file reference doc section on package source credentials](../reference/nuget-config-file.md#packagesourcecredentials) for more information, including syntax.
 However, it's easier to use [`dotnet nuget update source`](/dotnet/core/tools/dotnet-nuget-update-source) on the command line to set the credentials.
 
-> [!CAUTION]
+> [!WARNING]
 > Take care when setting credentials in *nuget.config* files, especially when saving the credential as plain text.
 > If the credential is written to a *nuget.config* file that is in source control, there is an increased risk of accidentally leaking the secret.
 >
--- a/docs/consume-packages/consuming-packages-authenticated-feeds.md
+++ b/docs/consume-packages/consuming-packages-authenticated-feeds.md
@@ -54,7 +54,7 @@ If these options are not feasible, you can store the credentials in the *nuget.c
 However, this option should only be used in environments where no other secure option is available.
 For more information, refer to the section on [credentials in *nuget.config* files](#credentials-in-nugetconfig-files).
 
-    > [!CAUTION]
+    > [!WARNING]
     > Storing credentials in clear text in the *nuget.config* file, especially when saving the file in source control, is risky as it increases the chances of accidental credential leaks.
     > If you must store credentials in the *nuget.config* file, consider using one of the more secure options mentioned above.
 
--- a/docs/consume-packages/consuming-packages-authenticated-feeds.md
+++ b/docs/consume-packages/consuming-packages-authenticated-feeds.md
@@ -18,39 +18,38 @@ For HTTP feeds, NuGet will make an unauthenticated request, and if the server re
 
 ## Security best practices for managing credentials
 
-Although NuGet searches for credentials in the above-mentioned order, we recommend the following order to manage credentials for authenticating with private feeds in a secure manner:
+Although NuGet searches for credentials in the order mentioned above, we recommend the following sequence for securely managing credentials when authenticating with private feeds:
 
 1. **Credential Provider**: It is highly recommended to use a credential provider whenever possible.
-This avoids storing secrets in plain text and reduces the risk of accidentally leaking secrets via source control.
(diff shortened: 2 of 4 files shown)


docs/consume-packages/consuming-packages-authenticated-feeds.md
```diff
@@ -16,11 +16,40 @@ For HTTP feeds, NuGet will make an unauthenticated request, and if the server re
...
+If the credential provider supports single sign-on, it may decrease the frequency of logins or the number of places where credentials need to be saved. Refer to the [credential providers](#credential-providers) section for more information.
+
+1. **Encrypted Credentials in nuget.config**: If a credential provider is not available, you should consider using encrypted credentials.
+This method provides an extra layer of security by storing the credentials in an encrypted format.
+For more information, refer to the section on [credentials in *nuget.config* files](#credentials-in-nugetconfig-files).
+
+    > [!NOTE]
+    > :warning: **WARNING** :warning: Be aware that encrypted passwords are only supported on Windows. Moreover, they can only be decrypted on the same machine and by the same user who originally encrypted them.
+
+1. **Using Environment Variable Macros in nuget.config**: If using encrypted credentials is not possible, consider storing the credentials in the *nuget.config* file with environment variable macros.
```
nkolev92, line 35: Worth noting what this is preferred over the environment variable approach. 

In particular, this approach provides better transparency for the end user to figure out how their credentials are configured.

nkolev92 (approved):
This is great @kartheekp-ms!

1 suggestion that I think would better clarify why the env var + config is preferred over the env var only one.

docs/concepts/Security-Best-Practices.md
```diff
@@ -121,6 +121,8 @@ When using multiple public & private NuGet source feeds, a package can be downlo
 
 For more information to secure your package feeds, see [3 Ways to Mitigate Risk When Using Private Package Feeds](https://azure.microsoft.com/resources/3-ways-to-mitigate-risk-using-private-package-feeds/).
```
zivkan, line 122: I know my comment is out of scope for the intent of this PR, but it occurs to me that this whitepaper was written before Package Source Mapping was written, and therefore doesn't propose PSM as a potential mitigation.

Also, the link no longer takes us to the whitepaper, so maybe this entire paragraph should be delete 🤷

docs/consume-packages/consuming-packages-authenticated-feeds.md
```diff
@@ -16,11 +16,40 @@ For HTTP feeds, NuGet will make an unauthenticated request, and if the server re
...
-> Additionally, it typically reduces the number of places you need to update when a credential expires or changes.
-> If the credential provider supports single sign-on, it may reduce the number of times you need to login, or the number of places that credentials need to be saved.
+## Security best practices for managing credentials
+
+Although NuGet searches for credentials in the order mentioned above, we recommend the following sequence for securely managing credentials when authenticating with private feeds:
+
+1. **Credential Provider**: It is highly recommended to use a credential provider whenever possible.
+This approach avoids storing secrets in plain text and minimizes the risk of accidentally exposing secrets through source control.
+Moreover, it generally reduces the number of places you need to update when a credential expires or changes.
+If the credential provider supports single sign-on, it may decrease the frequency of logins or the number of places where credentials need to be saved. Refer to the [credential providers](#credential-providers) section for more information.
```
zivkan, line 26:
```suggestion
If the credential provider supports single sign-on, it may decrease the frequency of logins or the number of places where credentials need to be saved.
Refer to the [credential providers](#credential-providers) section for more information.
```

Once sentence per line please

docs/consume-packages/consuming-packages-authenticated-feeds.md
```diff
@@ -16,11 +16,40 @@ For HTTP feeds, NuGet will make an unauthenticated request, and if the server re
...
+
+1. **Encrypted Credentials in nuget.config**: If a credential provider is not available, you should consider using encrypted credentials.
+This method provides an extra layer of security by storing the credentials in an encrypted format.
+For more information, refer to the section on [credentials in *nuget.config* files](#credentials-in-nugetconfig-files).
+
+    > [!NOTE]
+    > :warning: **WARNING** :warning: Be aware that encrypted passwords are only supported on Windows. Moreover, they can only be decrypted on the same machine and by the same user who originally encrypted them.
```
zivkan, line 33:
```suggestion
    > :warning: **WARNING** :warning: Be aware that encrypted passwords are only supported on Windows. 
    > Moreover, they can only be decrypted on the same machine and by the same user who originally encrypted them.
```

One sentence per line, please.

docs/consume-packages/consuming-packages-authenticated-feeds.md
```diff
@@ -16,11 +16,40 @@ For HTTP feeds, NuGet will make an unauthenticated request, and if the server re
...
+For more information, refer to the section on [credentials in *nuget.config* files](#credentials-in-nugetconfig-files).
+
+    > [!NOTE]
+    > :warning: **WARNING** :warning: Storing credentials in clear text in the *nuget.config* file, especially when saving the file in source control, is risky as it increases the chances of accidental credential leaks.
+    > If you must store credentials in the *nuget.config* file, consider using one of the more secure options mentioned above.
+
+By adhering to these best practices, you can securely authenticate private feeds while minimizing the risk of sensitive information exposure.
 
 The credentials you need to use are determined by the package source.
 Therefore, unless you're using a credential provider, you should check with your package source for what credentials to use.
```
zivkan, line 55: This paragraph (package sources define what the correct password/PAT is) doesn't make sense under the security recommendations section. It should be moved to the introduction, as it was before you introduced the new heading.

docs/consume-packages/consuming-packages-authenticated-feeds.md
```diff
@@ -68,7 +97,8 @@ For more information about valid authentication types, see [the docs on package
 See [the *nuget.config* file reference doc section on package source credentials](../reference/nuget-config-file.md#packagesourcecredentials) for more information, including syntax.
 However, it's easier to use [`dotnet nuget update source`](/dotnet/core/tools/dotnet-nuget-update-source) on the command line to set the credentials.
 
-> ![Warning]
+> [!NOTE]
+> :warning: **WARNING** :warning:
```
zivkan, line 101: Why this change?  Does it align with Microsoft's documentation standards?

The markdown reference docs: https://learn.microsoft.com/en-us/contribute/content/markdown-reference#alerts-note-tip-important-caution-warning
already lists the warning alert as:

> Dangerous certain consequences of an action

Using `> [!CAUTION]` might be more appropriate though

> Negative potential consequences of an action

You've used this note alert with warning icons in multiple places in the PR.

docs/reference/nuget-config-file.md
```diff
@@ -150,6 +150,12 @@ Optionally, valid authentication types can be specified with the `-validauthenti
 | cleartextpassword | The unencrypted password for the source. Note: environment variables can be used for improved security. |
 | validauthenticationtypes | Comma-separated list of valid authentication types for this source. Set this to `basic` if the server advertises NTLM or Negotiate and your credentials must be sent using the Basic mechanism, for instance when using a PAT with on-premises Azure DevOps Server. Other valid values include `negotiate`, `kerberos`, `ntlm`, and `digest`, but these values are unlikely to be useful. |
 
+> [!NOTE]
+> :warning: **WARNING** :warning:
+> Storing passwords in clear text is strongly discouraged.
+> Please note that encrypted passwords are only supported on Windows. Furthermore, they can only be decrypted when used on the same machine and by the same user who originally encrypted them.
```
zivkan, line 156:
```suggestion
> Please note that encrypted passwords are only supported on Windows.
> Furthermore, they can only be decrypted when used on the same machine and by the same user who originally encrypted them.
```

one sentence per line, please.

docs/reference/cli-reference/cli-ref-sources.md
```diff
@@ -53,6 +53,11 @@ where `<operation>` is one of *List, Add, Remove, Enable, Disable,* or *Update*,
 
   Specifies the password for authenticating with the source.
 
+  > [!Note]
+  > Make sure to add the sources' password under the same user context as the nuget.exe is later used to access the package source.
+  > The password will be stored encrypted in the config file and can only be decrypted in the same user context as it was encrypted.
+  > So for example when you use a build server to restore NuGet packages the password must be encrypted with the same Windows user under which the build server task will run.
```
zivkan, line 59: Encrypted passwords only work on the same machine, in addition to the same account. Therefore, committing encrypted passwords into source control, or otherwise sharing a nuget.config with encrypted password, never makes sense.

I think the wording you used elsewhere in the PR was better. I'm not sure why build servers are being called out specifically here.

docs/consume-packages/consuming-packages-authenticated-feeds.md
zivkan, line 33: encrypted passwords being Windows only doesn't feel like a "warning" to me, at least not according to what appears to by learn.microsoft.com's recommendations: https://learn.microsoft.com/en-us/contribute/content/markdown-reference#alerts-note-tip-important-caution-warning

Also, the markdocs reference docs say:

> Avoid notes, tips, and important boxes. Readers tend to skip over them. It's better to put that info directly into the article text.

docs/concepts/Security-Best-Practices.md
kartheekp-ms: Thank you, Andy, for noticing this broken link. The link, https://azure.microsoft.com/resources/3-ways-to-mitigate-risk-using-private-package-feeds/en-us/, is now operational. Generally, we try to avoid using the `en-us` locale in our documentation. However, in this instance, appending `/en-us/` has resolved the issue. I will incorporate this change into the current PR. If we find a link in the future that functions without specifying any locale, we can update it accordingly.

docs/consume-packages/consuming-packages-authenticated-feeds.md
kartheekp-ms: Address this comment on the latest commit.

docs/consume-packages/consuming-packages-authenticated-feeds.md
kartheekp-ms: Good catch, Andy. I addressed this comment in the latest commit by moving this paragraph before the new heading.

docs/consume-packages/consuming-packages-authenticated-feeds.md
kartheekp-ms: I didn’t know that there was guidance on alerts. Thank you for taking the time to find this information. I updated all occurrences of `:warning:` to `[!CAUTION]` in the latest commit.

docs/reference/cli-reference/cli-ref-sources.md
kartheekp-ms: > I'm not sure why build servers are being called out specifically here.

I copied the entire paragraph from the existing documentation https://learn.microsoft.com/nuget/reference/cli-reference/cli-ref-sources (refer to `Note` in the doc). I will change this note to the following.

```
    > [!NOTE]
    > Be aware that encrypted passwords are only supported on Windows. 
    > Moreover, they can only be decrypted on the same machine and by the same user who originally encrypted them.
````

File: docs/consume-packages/consuming-packages-authenticated-feeds.md (hunk 4 above)
Comment (kartheekp-ms) [reply docs/consume-packages/consuming-packages-authenticated-feeds.md]: I agree with your feedback, Andy. I changed the alert to a `Note` type. The documentation states that it should provide information the user should notice even if skimming, which sounds appropriate for this case.

**Review by nkolev92 — APPROVED** [review]

File: docs/reference/nuget-config-file.md [hunk 9]
```diff
@@ -170,49 +176,51 @@ In the config file, the `<packageSourceCredentials>` element contains child node
...
     <Test_x0020_Source>
         <add key="Username" value="user" />
-        <add key="ClearTextPassword" value="hal+9ooo_da!sY" />
+        <add key="ClearTextPassword" value="%TestSourcePassword%" />
     </Test_x0020_Source>
 </packageSourceCredentials>
 ```
 
-Additionally, valid authentication methods can be supplied:
+When using unencrypted passwords.
```
JonDouglas, line 211:
```suggestion
When using unencrypted passwords:
```

JonDouglas (approved):
Great changes! These will surely help provide some further best practices. Thank you kindly!

PR #3256, opened 2024-03-07
Remove old stale bot (#3255)

JonDouglas:
* Switch from GH doc issue feedback to new feedback control

* Add GitHub stale issues bot

* removed old stale bot


nkolev92:
you only have the stale bot changes here, and since `main` is the main branch, you probably don't need these in live immediately.

PR #3260, opened 2024-03-13
Distinguish case sensitivity for NuGet.Config keys

donnie-msft:
"always" case sensitive is misleading, as it makes it sound like adding a new key with a different case should be acceptable.
My goal is to clarify that keys are referenced in a case-sensitive manner.
Adding "nuget.org" and "nuGet.org" though should be denied by NuGet tooling as they are not unique (case-insensitive).

### DIFF
--- a/docs/consume-packages/configuring-nuget-behavior.md
+++ b/docs/consume-packages/configuring-nuget-behavior.md
@@ -47,7 +47,7 @@ A `NuGet.Config` file is a simple XML text file containing key/value pairs as de
 Settings are managed using the NuGet CLI [config command](../reference/cli-reference/cli-ref-config.md):
 - By default, changes are made to the user-level config file. (On Mac/Linux, the location of user-level config file varies by tooling)
 - To change settings in a different file, use the `-configFile` switch. In this case files can use any filename.
-- Keys are always case sensitive.
+- Keys are case sensitive when being referenced. New keys must be unique regardless of casing (case insensitive) .
 - Elevation is required to change settings in the computer-level settings file.
 
 > [!Warning]


docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -47,7 +47,7 @@ A `NuGet.Config` file is a simple XML text file containing key/value pairs as de
 Settings are managed using the NuGet CLI [config command](../reference/cli-reference/cli-ref-config.md):
 - By default, changes are made to the user-level config file. (On Mac/Linux, the location of user-level config file varies by tooling)
 - To change settings in a different file, use the `-configFile` switch. In this case files can use any filename.
-- Keys are always case sensitive.
+- Keys are case sensitive when being referenced. New keys must be unique regardless of casing (case insensitive) .
```
nkolev92, line 50: This sounds like a developer guide instead of a user facing doc.

I think you can tweak the phrasing to communicate the same thing in a slightly different way. 

```suggestion
Keys are case sensitive, but they always unique regardless of casing.
```

You don't have to accept my suggestion, but just an idea that I found more in style with the rest of the doc.

donnie-msft: I don't find that more clear, so I'll let others chime in.

nkolev92: I think it's fair to question whether my proposal is good enough, but `New keys must be unique regardless of casing (case insensitive) .` currently is aimed the nuget devs since we're the only ones that can add new keys and we're not the audience for this doc.

donnie-msft: Who is the target audience for this document? I thought if a customer is consuming packages, that means they are modifying nuget.configs, which they need to know the 2 points:
1. that it matters what casing they use for their keys
2. once a casing is created, another casing for the same key cannot be used

donnie-msft: >     <packageSources>
>         <add key="Contoso Package Source" value="https://contoso.com/packages/" />
>         <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
>     </packageSources>

Here's a sample from the document. We need to say that this is invalid for a customer to add:
`<add key="CONTOSO Package Source" value="https://contoso.com/packages/" />`

nkolev92: I think the doc you are changing is talking about the `<config>` section of the nuget.config file. 

Keys such as `globalPackagesFolder` etc, not the sources.

nkolev92: Now that I read more of it, it never mentions keys such as sources/package source mapping, while it probably should, or at least link to a page that does.

donnie-msft: For "being referenced", I gave 2 examples here https://github.com/NuGet/docs.microsoft.com-nuget/pull/3260#discussion_r1525608514

As far as making product changes, that should be a NuGet/Home issue, not a docs issue. I'm just trying to make the docs point out what the product does now, not what makes sense to change (which would be another docs change once the product change is released).

Issue #2769, opened 2022-05-25
[Bug]: Hierarchical nuget.config files are ignored when `-configFile` parameter is used


yangskyboxlabs:
### NuGet Product Used

NuGet.exe

### Product Version

6.1.0, 5.3.1

### Impact

It's more difficult to complete my work

### Repro Steps & Context

When `-configFile` option is used to invoke NuGet.exe, hierarchical configuration support appears to be disabled. This contradicts the documented behaviour: https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior

```
PS D:\nuget\subdir> ..\nuget.exe
NuGet Version: 6.1.0.106
usage: NuGet <command> [args] [options]
[-- snip --]

PS D:\nuget\subdir> cat ..\NuGet.Config
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="parent" value="https://example.com/parent/nuget/v3/index.json" />
  </packageSources>
</configuration>

PS D:\nuget\subdir> cat .\NuGet.Config
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="sub" value="https://example.com/sub/nuget/v3/index.json" />
  </packageSources>
</configuration>

PS D:\nuget\subdir> ..\nuget.exe sources
Registered Sources:
  1.  sub [Enabled]
      https://example.com/sub/nuget/v3/index.json
  2.  parent [Enabled]
      https://example.com/parent/nuget/v3/index.json

PS D:\nuget\subdir> ..\nuget.exe sources -configFile .\NuGet.Config
Registered Sources:
  1.  sub [Enabled]
      https://example.com/sub/nuget/v3/index.json
```

yangskyboxlabs:
PR posted.

PR #3262, opened 2024-03-14
nuget: Correct conf loading rules with -configFile

yangskyboxlabs:
When -configFile is given, only the specified file is scanned for configuration, and the hierarchical config loading behavior is not used.

#2769

### DIFF
--- a/docs/consume-packages/configuring-nuget-behavior.md
+++ b/docs/consume-packages/configuring-nuget-behavior.md
@@ -114,13 +114,14 @@ Multiple `NuGet.Config` files allow you to store settings in different locations
 These settings collectively apply to any NuGet operation invoked from the command line or from Visual Studio, with settings that exist "closest" to a solution or the current folder taking precedence.
 If a command line tool is used on a project file, rather than a solution file, then the project directory is used as the "solution directory", which can lead to inconsistent behaviour when there is a `NuGet.Config` file in a subdirectory of the solution file.
 
-Specifically, NuGet loads settings from the different config files in the following order:
+Specifically, when a config file is not specified explicitly on the command line, NuGet loads settings from the different config files in the following order:
 
 1. The [`NuGetDefaults.Config` file](#nuget-defaults-file), which contains settings related only to package sources.
 1. The computer-level file.
 1. The user-level file.
-1. The file specified with `-configFile`.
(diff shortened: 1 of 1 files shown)


yangskyboxlabs:
@dotnet-policy-service agree [company="SkyBox Labs Inc"]

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -114,14 +114,16 @@ Multiple `NuGet.Config` files allow you to store settings in different locations
 These settings collectively apply to any NuGet operation invoked from the command line or from Visual Studio, with settings that exist "closest" to a solution or the current folder taking precedence.
 If a command line tool is used on a project file, rather than a solution file, then the project directory is used as the "solution directory", which can lead to inconsistent behaviour when there is a `NuGet.Config` file in a subdirectory of the solution file.
 
-Specifically, NuGet loads settings from the different config files in the following order:
+Specifically, when `-configFile` command line option is not specified, NuGet loads settings from the different config files in the following order:
```
nkolev92, line 117: The config file option differs across different tools, for example `-configFile` for nuget.exe, `--configfile` for dotnet.exe and `RestoreConfigFile` for msbuild. 

Maybe keep the sentence more generic rather than call out a specific option?

nkolev92 (commented):
Thanks for creating this PR!

docs/consume-packages/configuring-nuget-behavior.md
yangskyboxlabs: Hm. I wasn't even aware that `dotnet` had a `--configfile` option.

Going to remove this change, and be more explicit about this being a `nuget.exe` specific behaviour below.

nkolev92: All CLI tools have the options. 

How about saying `when a config file is not specified explicitly on the commandline`

That way it's more generic, since these docs apply to all tools.

yangskyboxlabs: Had a deeper look through the docs to find where `--configfile` is for `dotnet`.

I do agree a more generic text would be more appropriate.

How about this latest change?

nkolev92 (approved):

zivkan (approved):

PR #3268, opened 2024-03-18

bangjiehan:
https://github.com/microsoft/MSBuildSdks/pull/135


bangjiehan:
@dotnet-policy-service agree

nkolev92 (changes requested):
I don't think that update is correct: https://github.com/NuGet/NuGet.Client/blob/1845d6bd450a7453d573035371c9fec43683d1ef/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets#L122. 

cc @jeffkl

PR #3272, opened 2024-03-29
Test fix to auto-close for stale-issues.yml

Mikejo5000:
Stale issues marked as "stale" before 3/14/24, and still untouched since then, should have been auto-closed, but weren't. Testing this bug fix I found related to the bot.


Mikejo5000:
Fixed issue by manually deleting the _state cache

PR #3278, opened 2024-04-05
Update stale issues bot

Mikejo5000:
Quite a few stale issues are labeled "doc-enhancement", which at the moment are excluded from closure. We should probably close these as well.


Mikejo5000:
@nkolev92 Nikolche, were you labeling any issues "doc-enhancement" to keep them open? It seems best to auto-close stale issues labeled doc-enhancement, so I made this change to the bot. ("exempt" is the best label to keep issues open for the time being.)

nkolev92:
Maybe I'm misinterpreting what doc-enhancement means, but that's an ask for new content, usually for new or existing features. 

I think having undocumented features is not great.

Mikejo5000:
@nkolev92 Yes, that's right about what the label means. And yes, undocumented features is not good either. The main issue here is how best to close out the GitHub issues by end of June. I will forward you the e-mail in case you didn't see it. Remaining issues at the end can be converted to ADO items, although I'm not sure what resources will be available for team members to look at those. We can keep doc-enhancement open for now, and tentatively plan to convert to ADO.

nkolev92 (approved):

PR #3280, opened 2024-04-11
A string represents a single owner's username

donnie-msft:
I'd like to specify what a string represents. It could be assumed to be a comma-delimited list of owners.

### DIFF
--- a/docs/api/search-query-service-resource.md
+++ b/docs/api/search-query-service-resource.md
@@ -111,7 +111,7 @@ versions       | array of objects           | yes      | All of the versions of 
 authors        | string or array of strings | no       | 
 iconUrl        | string                     | no       | 
 licenseUrl     | string                     | no       | 
-owners         | string or array of strings | no       | 
+owners         | string or array of strings | no       | A string represents a single owner's username.
 projectUrl     | string                     | no       | 
 registration   | string                     | no       | The absolute URL to the associated [registration index](registration-base-url-resource.md#registration-index)
 summary        | string                     | no       | 
--- a/docs/api/search-query-service-resource.md
+++ b/docs/api/search-query-service-resource.md
@@ -111,7 +111,7 @@ versions       | array of objects           | yes      | All of the versions of 
 authors        | string or array of strings | no       | 
 iconUrl        | string                     | no       | 
 licenseUrl     | string                     | no       | 
-owners         | string or array of strings | no       | A string represents a single owner's username.
+owners         | string or array of strings | no       | A string represents a single owner's username
 projectUrl     | string                     | no       | 
 registration   | string                     | no       | The absolute URL to the associated [registration index](registration-base-url-resource.md#registration-index)
 summary        | string                     | no       | 


docs/api/search-query-service-resource.md
```diff
@@ -111,7 +111,7 @@ versions       | array of objects           | yes      | All of the versions of
 authors        | string or array of strings | no       | 
 iconUrl        | string                     | no       | 
 licenseUrl     | string                     | no       | 
-owners         | string or array of strings | no       | 
+owners         | string or array of strings | no       | A string represents a single owner's username.
```
donnie-msft, line 114:
```suggestion
owners         | string or array of strings | no       | A string represents a single owner's username
```

joelverhagen (approved):

docs/api/search-query-service-resource.md
```diff
@@ -111,7 +111,7 @@ versions       | array of objects           | yes      | All of the versions of
 authors        | string or array of strings | no       | 
 iconUrl        | string                     | no       | 
 licenseUrl     | string                     | no       | 
-owners         | string or array of strings | no       | 
+owners         | string or array of strings | no       | A string represents a single owner's username
```
zivkan, line 114: authors and tags also have type "string or array of strings".  If we're adding the note for owners, I think the other two should have the note too.

joelverhagen: It would be good if could know whether or not this is broadly true across package feed implementations today, but that might be hard to nail down accurately without some effort.

When I wrote the docs originally, I did some analysis of how multiple feeds handle authors and tags but I forget which implementations were quirky compared to others. I put "string or array of strings" because there was a variety in existing feed implementations (the woes of retro-specing or docs long after implementation). To illustrate what I mean:

Just on NuGet.org, `authors` is only ever an array of a single string, but that string may contain commas. It can also be an empty string. No splitting is done on the .nuspec authors field (which is a single string) to make it into an array of 0 to N elements. For `tags`, we _do_ split on some separators ([source](https://github.com/NuGet/NuGet.Jobs/blob/6fe505f86b5cd0aeae85263a0d385decfb1c1787/src/Catalog/Helpers/Utils.cs#L37-L48)). So, it's always an array and can have 0 to N elements (this is the ideal situation IMO since it tries to make the best of an under-modelled .nuspec situation).

And for owners it can be 0 to N string elements in an array and never any empty strings. Also good (IMO).

So owners and tags are consistent, but authors... is kind of not. And that's just NuGet.org. And yeah, it kind of sucks! 

IMO, this specific change (adding a note to owners) is most important of the 3 fields because the semantics of a single string matters now for using the owner template, i.e. we're explicitly not splitting the single string on comma or something like that. We would just plop that CSV value into the template as if it's a single username.

As you say, we can opt to draw a line in the sand for authors and tags now, but I think there is less urgency. If you want to add a note to these, it might be worth performing a survey of existing implementations to understand if a) we're pulling the rug out and asserting specificity where there was none before and in doing so making other package feeds "non-compliant" and b) if there are more variants worth covering, such as the silly thing we have for NuGet.org authors where there's only a single array element but it's a CSV of authors!

My personally opinion is to leave authors and tags as is (no note) since there's no need to investigate it right now, but it's indeed righteous work to go figure out the package feed quirks and try to wrangle them into one cohesive picture. We could also as a longer note about how this whole thing is 🤬.

joelverhagen: Oh, and the primary concern for the shape of the field (string or array of strings) is: **what does client accept?** Perhaps we can just document that, but that may not alone be enough to be fully clear on the model because client may not care if tags or authors are a single string containing multiple values separated with commas if for display purposes it just shows the string verbatim and doesn't linkify or templatize the values at all.

donnie-msft: > authors and tags also have type "string or array of strings". If we're adding the note for owners, I think the other two should have the note too.

Yes, they are the same type, but I'm not working on changing Client around either of those types, right now. 

> We would just plop that CSV value into the template as if it's a single username.

This sentence boils down the problem I wanted to address.

Issue #3123, opened 2023-08-16
Improve outdated examples of package sources in `NuGet.Configs`


donnie-msft:
Several examples here are outdated, using terminology that is inconsistent with our other verbiage in docs/messages/blogs, and may be confusing
eg, `"NuGet official package source"`

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 370d9d91-9f35-79aa-3ace-7b77436e988f
* Version Independent ID: 3da4591a-3d3d-c07a-6672-97964de13b14
* Content: [Common NuGet configurations](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#settings-walkthrough)
* Content Source: [docs/consume-packages/configuring-nuget-behavior.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/configuring-nuget-behavior.md)
* Service: **nuget**
* GitHub Login: @JonDouglas
* Microsoft Alias: **jodou**

kartheekp-ms:
@donnie-msft - I am not sure why I assigned this issue to myself. I have now removed my assignment. Sorry for the confusion. Please feel free to assign this issue to yourself if you are still interested.

Mikejo5000:
We apologize for the inconvenience. We are phasing out GitHub issues as the feedback mechanism on NuGet documentation content and replacing it with a new feedback system using the Feedback link within the articles. If you believe your feedback needs to be preserved and addressed, please submit a PR or open a new issue at https://github.com/NuGet/Home/issues

Issue #2645, opened 2022-01-23
Inaccurate statement regarding `globalPackagesFolder` and project restore style.


JunielKatarn:
The key `globalPackagesFolder` is said to be related to `PackageReference` only.
This is not true.

Setting/overriding this key using the `packages.config` mechanism still affects the global user cache for NuGet package downloads.

Try running `NuGet.exe -Set globalPackagesFolder=(some alternative path)` and then `NuGet.exe restore` on a project using `packages.config` to verify.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0ea1390e-db36-6759-3095-ebab9d546fdb
* Version Independent ID: ca88d8c6-0515-a267-aac3-a508c54157ab
* Content: [nuget.config File Reference](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file)
* Content Source: [docs/reference/nuget-config-file.md](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/reference/nuget-config-file.md)
* Product: **nuget**
* GitHub Login: @JonDouglas
* Microsoft Alias: **jodou**

JunielKatarn:
The documentation still contains the false statement.

https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file

JunielKatarn:
Yes, this feedback still needs to be reviewed.

PR #3290, opened 2024-05-09
Update nuget.config examples & unemphasize NuGetDefaults.Config

donnie-msft:
Fixes: https://github.com/NuGet/docs.microsoft.com-nuget/issues/3123

1. Remove "NuGet official package source" and deprecated `activePackageSource` sample and instead use a nuget.config generated by `dotnet new nugetconfig`.
2. Reference `dotnet new nugetconfig` as a way to create a new config.
3. Declare in examples using `repositoryPath` that this is a packages.config setting, and add a PackageReference sample as well.
4. Reduce mention of `NuGetDefaults.Config` and call out that it is uncommon.

### DIFF
--- a/docs/consume-packages/configuring-nuget-behavior.md
+++ b/docs/consume-packages/configuring-nuget-behavior.md
@@ -9,7 +9,7 @@ ms.topic: conceptual
 
 # Common NuGet configurations
 
-NuGet's behavior is driven by the accumulated settings in one or more `NuGet.Config` (XML) files that can exist at solution- (project if no solution is used), user-, and computer-wide levels.
+NuGet's behavior is driven by the accumulated settings in one or more config (XML) files that can exist at solution- (project if no solution is used), user-, and computer-wide levels.
 
 ## Config file locations and uses
 
@@ -58,7 +58,7 @@ Settings are managed using the NuGet CLI [config command](../reference/cli-refer
 Windows:
 
 ```cli
-# Set globalPackagesFolder (available for PackageReference only) in the user-level config file
+# Set globalPackagesFolder in the user-level config file
 nuget config -set globalPackagesFolder=c:\packages
 
 # Set repositoryPath (available for packages.config only) in the user-level config file
@@ -237,7 +237,7 @@ These files cannot be edited by the standard tooling.
 
 The `NuGetDefaults.Config` is uncommon and can only specify package sources from which packages are installed and updated, or control the default target for publishing packages with `nuget push`.
 
-Because administrators can conveniently (using Group Policy, for example) deploy consistent `NuGetDefaults.Config` files to developer and build machines, they can ensure that everyone in the organization is using the correct package sources, whether or not that includes nuget.org.
+Because administrators can conveniently (using Group Policy, for example) deploy consistent `NuGetDefaults.Config` files to developer and build machines, they can ensure that everyone in the organization is using consistent package sources, whether or not that includes nuget.org.
 
 > [!Important]
 > The `NuGetDefaults.Config` file never causes a package source to be removed from a developer's NuGet configuration. That means if the developer has already used NuGet and therefore has the nuget.org package source registered, it won't be removed after the creation of a `NuGetDefaults.Config` file.
--- a/docs/consume-packages/configuring-nuget-behavior.md
+++ b/docs/consume-packages/configuring-nuget-behavior.md
@@ -9,7 +9,7 @@ ms.topic: conceptual
 
 # Common NuGet configurations
 
-NuGet's behavior is driven by the accumulated settings in one or more `NuGet.Config` (XML) files that can exist at solution- (project if no solution is used), user-, and computer-wide levels. A global `NuGetDefaults.Config` file also specifically configures package sources. Settings apply to all commands issued in the CLI, the Package Manager Console, and the Package Manager UI.
+NuGet's behavior is driven by the accumulated settings in one or more `NuGet.Config` (XML) files that can exist at solution- (project if no solution is used), user-, and computer-wide levels.
 
 ## Config file locations and uses
 
@@ -58,7 +58,10 @@ Settings are managed using the NuGet CLI [config command](../reference/cli-refer
 Windows:
 
 ```cli
-# Set repositoryPath in the user-level config file
+# Set globalPackagesFolder (available for PackageReference only) in the user-level config file
+nuget config -set globalPackagesFolder=c:\packages
+
+# Set repositoryPath (available for packages.config only) in the user-level config file
 nuget config -set repositoryPath=c:\packages 
 
 # Set repositoryPath in solution-level files
@@ -72,7 +75,10 @@ nuget config -set repositoryPath=c:\packages -configfile %ProgramFiles(x86)%\NuG
 Mac/Linux:
 
 ```cli
-# Set repositoryPath in the user-level config file
+# Set globalPackagesFolder (available for PackageReference only) in the user-level config file
+nuget config -set globalPackagesFolder=/home/packages
+
+# Set repositoryPath (available for packages.config only) in the user-level config file
 nuget config -set repositoryPath=/home/packages 
 
 # Set repositoryPath in solution-level files
@@ -84,7 +90,7 @@ nuget config -set repositoryPath=/home/packages -configfile $XDG_DATA_HOME/NuGet
 ```
 
 > [!Note]
-> In NuGet 3.4 and later you can use environment variables in any value, as in `repositoryPath=%PACKAGEHOME%` (Windows) and `repositoryPath=$PACKAGEHOME` (Mac/Linux).
+> In NuGet 3.4+ you can use environment variables in any value, as in `repositoryPath=%PACKAGEHOME%` (Windows) and `repositoryPath=$PACKAGEHOME` (Mac/Linux).
 
 ### Removing a value
 
@@ -100,7 +106,10 @@ nuget config -set repositoryPath= -configfile /home/my.Config
 
 ### Creating a new config file
 
-Copy the template below into the new file and then use `nuget config -configFile <filename>` to set values:
+Using the .NET CLI, create a default nuget.config by running `dotnet new nugetconfig`.
+For more information, see [dotnet CLI commands](../reference/dotnet-commands.md#package-consumption).
+
+Alternatively, manually copy the template below into the new file and then use `nuget config -configFile <filename>` to set values:
 
 ```xml
 <?xml version="1.0" encoding="utf-8"?>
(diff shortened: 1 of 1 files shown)


docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -226,7 +235,9 @@ These files cannot be edited by the standard tooling.
 
 ## NuGet defaults file
 
-The `NuGetDefaults.Config` file exists to specify package sources from which packages are installed and updated, and to control the default target for publishing packages with `nuget push`. Because administrators can conveniently (using Group Policy, for example) deploy consistent `NuGetDefaults.Config` files to developer and build machines, they can ensure that everyone in the organization is using the correct package sources rather than nuget.org.
+The `NuGetDefaults.Config` is uncommon and can only specify package sources from which packages are installed and updated, or control the default target for publishing packages with `nuget push`.
+
+Because administrators can conveniently (using Group Policy, for example) deploy consistent `NuGetDefaults.Config` files to developer and build machines, they can ensure that everyone in the organization is using the correct package sources, whether or not that includes nuget.org.
```
donnie-msft, line 240: "ensure" is a strong word. Can't `<clear/>` still invalidate this config? I wonder if we should drop this paragraph entirely.

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -9,7 +9,7 @@ ms.topic: conceptual
 
 # Common NuGet configurations
 
-NuGet's behavior is driven by the accumulated settings in one or more `NuGet.Config` (XML) files that can exist at solution- (project if no solution is used), user-, and computer-wide levels. A global `NuGetDefaults.Config` file also specifically configures package sources. Settings apply to all commands issued in the CLI, the Package Manager Console, and the Package Manager UI.
+NuGet's behavior is driven by the accumulated settings in one or more `NuGet.Config` (XML) files that can exist at solution- (project if no solution is used), user-, and computer-wide levels.
```
zivkan, line 12:
```suggestion
NuGet's behavior is driven by the accumulated settings in one or more config (XML) files that can exist at solution- (project if no solution is used), user-, and computer-wide levels.
```

There are at least two places where `*.config` files are loaded. The machine wide config location, where VS customers will usually see the "offline" feed config file. And starting from around 2020, maybe 2021, a `config` subdirectory where the user-profile nuget.config file is located.

I expect that the machine wide config directory is documented. If the `config\*.config` pattern isn't also documented, it would be great to add it.

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -58,7 +58,10 @@ Settings are managed using the NuGet CLI [config command](../reference/cli-refer
 Windows:
 
 ```cli
-# Set repositoryPath in the user-level config file
+# Set globalPackagesFolder (available for PackageReference only) in the user-level config file
```
zivkan, line 61: The global package folder is still used by packages.config, to avoid downloading packages from the internet. It's just also copied to the solution packages folder (repository path), which is used by the compiler and build system.

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -116,7 +125,7 @@ If a command line tool is used on a project file, rather than a solution file, t
 
 Specifically, when a config file is not specified explicitly on the command line, NuGet loads settings from the different config files in the following order:
 
-1. The [`NuGetDefaults.Config` file](#nuget-defaults-file), which contains settings related only to package sources.
+1. (*Uncommon*) The [`NuGetDefaults.Config` file](#nuget-defaults-file), which contains settings related only to package sources.
 1. The computer-level file.
 1. The user-level file.
 1. Files found in every folder in the path from the drive root to the current folder (where `nuget.exe` is invoked or the folder containing the Visual Studio solution). For example, if a command is invoked in `c:\A\B\C`, NuGet looks for and loads config files in `c:\`, then `c:\A`, then `c:\A\B`, and finally `c:\A\B\C`.
```
zivkan, line 131: I believe that user-level `config\*.config` is loaded after the user-level file, but before the solution directory and parents.

docs/consume-packages/configuring-nuget-behavior.md
zivkan: No, when I was working on some config file changes for audit sources, I discovered that sources defined in `NuGetDefaults.Config` cannot be removed or disabled.  This should be documented.

I also don't like "correct package sources". I'd say "consistent package sources", "company package sources" or something similar.

zivkan (approved):
Current changes are an improvement, so good enough. But I think there are still some opportunities for improvements.

docs/consume-packages/configuring-nuget-behavior.md
donnie-msft: Is this really the place to go into that level of detail? The point here is to make it clear that the command works for a variety of settings, and not all settings apply in all scenarios.

zivkan: I think "available for PackageReference only" is somewhere between completely wrong, or just misleading.  If you don't think this is the right place to explain that the global packages folder is still used as a download cache for packages.config, that's fine.  But I hope you can understand that with the way this is currently phrased that some customers would expect that packages.config packages will never be extracted to this location, which is just not true.

donnie-msft: So, the fact that the docs for globalPackagesFolder have told customers this is also wrong, I assume?

> globalPackagesFolder (projects using PackageReference only)

https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#config-section

Nigusu-Allehu:
Team review: Let's merge this PR and create an issue for the improvements we could make

docs/consume-packages/configuring-nuget-behavior.md
zivkan: > this is also wrong, I assume?

In my opinion yes.  If you perform a restore of a packages.config project with ProcMon open, you'll see:
1. NuGet downloads the nupkg to the http-cache directory (to a temporary name then renames to the final name)
2. NuGet copies the nupkg from the http-cache directory to the global packages directory, and extracts it
3. NuGet copies the nupkg from the global packages directory to the solution packages directory and extracts it

One way to interpret "projects using PackageReference only" is that modifying the configuration file will have zero impact on projects using packages.config. However, unless what I wrote above is wrong, then the PackageReference only statement is wrong, unless the context is explained.

Do you disagree that without context, as is currently the case, that some customers might read these docs and assume that the global packages directory should remain unused by projects using packages.config, and therefore their global packages directory should be permanently empty/unused if they only ever use packages.config projects?

donnie-msft: Yes, clearly what you're saying PC does with GPF is not in alignment with the existing docs.

zivkan: FWIW, I just found a customer filed issue complaining about this precise documentation bug:
* https://github.com/NuGet/docs.microsoft.com-nuget/issues/2645

docs/consume-packages/configuring-nuget-behavior.md
donnie-msft: Please file an issue for this suggestion, per https://github.com/NuGet/docs.microsoft.com-nuget/pull/3290#issuecomment-2148077312

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -58,7 +58,10 @@ Settings are managed using the NuGet CLI [config command](../reference/cli-refer
 Windows:
 
 ```cli
-# Set repositoryPath in the user-level config file
+# Set globalPackagesFolder in the user-level config file
+nuget config -set globalPackagesFolder=c:\packages
```
zivkan, line 62: I think these examples should be changed to `dotnet nuget config` syntax, so it can be copy-pasted by more customers.

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -58,7 +58,10 @@ Settings are managed using the NuGet CLI [config command](../reference/cli-refer
...
+
+# Set repositoryPath (available for packages.config only) in the user-level config file
```
zivkan, line 64: So, I see that the nuget.exe command example already exists in the previous docs, but the comment especially saying to set this at the user level config should be considered a very poor practise. If someone wants to set the repositorypath config, they should only do it in the repo nuget.config.

I didn't test it myself, but I suspect that setting repositoryPath in the user nuget.config will cause the "solution" packages folder to be in the user's nuget config directory, and therefore any packages installed will cause project file modifications to reference files under the user's home directory and therefore will break everyone if commit into source control.

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -84,7 +90,7 @@ nuget config -set repositoryPath=/home/packages -configfile $XDG_DATA_HOME/NuGet
 ```
 
 > [!Note]
-> In NuGet 3.4 and later you can use environment variables in any value, as in `repositoryPath=%PACKAGEHOME%` (Windows) and `repositoryPath=$PACKAGEHOME` (Mac/Linux).
+> In NuGet 3.4+ you can use environment variables in any value, as in `repositoryPath=%PACKAGEHOME%` (Windows) and `repositoryPath=$PACKAGEHOME` (Mac/Linux).
```
zivkan, line 93: I expect that "3.4+" will be understood by most English speakers around the world. But I'm wondering if it will be well understood by non-English speakers who have a beginner level of English. Also, I'm wondering what the automatic translation into different languages will be like and if all the languages that learn.microsoft.com supports uses/understand the same "3.4+" convention, particularly those languages that don't use the latin alphabet.

Perhaps it's safer to leave "3.4 or later"?

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -72,7 +75,10 @@ nuget config -set repositoryPath=c:\packages -configfile %ProgramFiles(x86)%\NuG
 Mac/Linux:
 
 ```cli
-# Set repositoryPath in the user-level config file
+# Set globalPackagesFolder (available for PackageReference only) in the user-level config file
```
zivkan, line 78:
```suggestion
# Set globalPackagesFolder in the user-level config file
```

GPF isn't PR only.

docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -72,7 +75,10 @@ nuget config -set repositoryPath=c:\packages -configfile %ProgramFiles(x86)%\NuG
...
+nuget config -set globalPackagesFolder=/home/packages
+
+# Set repositoryPath (available for packages.config only) in the user-level config file
 nuget config -set repositoryPath=/home/packages
```
zivkan, line 82: Same comment as before. repositoryPath should not be set at the user level, and even if it sometimes works, I really don't feel like we should be "encouraging" it with examples like this.

docs/consume-packages/configuring-nuget-behavior.md
donnie-msft: I'm considering this out of scope as I'm not going to decide whether we remove explanations in our docs in this PR.

docs/consume-packages/configuring-nuget-behavior.md
donnie-msft: I didn't quite follow this comment (user-level is loaded after user-level?), but regardless as I mentioned in other places, I'm considering enhancements like this out of scope.

JonDouglas (approved):
Looks good and adds more modern guidance. Feel free to merge

PR #3291, opened 2024-05-09
Specify that PrivateAssets="all" also affects Publish Fixes #3270

### LINKED ISSUE #3270: PackageReference documentation doesn't include that PrivateAssets="all" turns off Publish
In this file:
https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/Package-References-in-Project-Files.md

It strongly implies that PrivateAssets="all" is equivalent to PrivateAssets="compile;runtime;..."

That isn't true. It also turns Publish off by default. That led to:
https://github.com/dotnet/sdk/issues/39465

The easiest solution to that issue was explicitly setting Publish="true" on the PackageReference. The documentation should mention both this option and the fact that it's made necessary by PrivateAssets="all" because that changes the default for Publish from true to false.

Forgind:
Fixes #3270

### DIFF
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -113,10 +113,16 @@ Allowable values for these tags are as follows, with multiple values separated b
 | analyzers | .NET analyzers |
 | native | Contents of the `native` folder |
 | none | None of the above are used. |
-| all | All of the above (except `none`) |
+| all | All of the above (except `none`). Also turns off "Publish" by default. |
+
+To set `PrivateAssets` to `all` without affecting publishing behavior, set `PrivateAssets` to `all` and `Publish` to `true` as in the first example below:
 
 ```xml
 <ItemGroup>
+    <!-- ... -->
+    <!-- This specifies all assets to be consumed without letting them flow to the parent project, -->
+    <!-- and it publishes the contents as normal -->
+    <PackageReference Include="Contoso.PrivateUtilities" Version="3.6.0" PrivateAssets="all" Publish="true" />
     <!-- ... -->
     <!-- Everything except the content files will be consumed by the project -->
     <!-- Everything except content files and analyzers will flow to the parent project-->


docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -113,10 +113,16 @@ Allowable values for these tags are as follows, with multiple values separated b
 | analyzers | .NET analyzers |
 | native | Contents of the `native` folder |
 | none | None of the above are used. |
-| all | All of the above (except `none`) |
+| all | All of the above (except `none`). Also turns off "Publish" by default. |
```
nkolev92, line 116: Is this documented anywhere on the SDK side?

Personally, I've always thought this was an incorrect decision that we unfortunately can't roll back, but it is still an SDK side decision, not NuGet. So I think the NuGet docs should reference an SDK side explanation.

Forgind: I don't think it's documented anywhere. I'm not sure where the code that does this is, but I imagine most customers who are looking for something like PrivateAssets="all" will find the NuGet docs, so I think we do need to mention it here.

I don't know where I'd add that in any SDK docs, so although I'm not against having it there somewhere there and having a link to that here, I'd probably have to dig around a bit to find where an appropriate place might be—unless you already have somewhere in mind for that?

nkolev92: I think publish docs should have it.

Forgind: Are you talking about https://learn.microsoft.com/dotnet/core/deploying/? I don't see anything that connects to how metadata (like PrivateAssets) would affect publishing...since that page was mostly talking about high-level things (what can you pass to the publish command), I feel like this would belong on a different page, but I didn't see anything better.

nkolev92: I think any doc that may be talking about publish actually does.

The PackageReference attribute Publish is a publish first thing, not a PackageReference first thing.
It never  flows through the assets file. 

PrivateAssets means what my consumers get, it never affects what gets consumed at build time. You can take a compile time or runtime dependency on a package that you've marked with PrivateAssets.
ExcludeAssets/IncludeAssets handle whether you're allowed to take that dependency.

Publish is different from build. 

That being said, i think this discussion is related to https://github.com/dotnet/sdk/issues/39400 now.

Forgind: My understanding from that discussion is that this is still correct and unlikely to change. I couldn't find anywhere on the SDK side that documented any part of this scenario. Are you ok with merging this?

nkolev92:
There's a discussion in https://github.com/dotnet/sdk/issues/47332 .

I think at this point we'd want to consider fixing that instead.

Forgind:
We never really finalized that discussion, as not enough of the right people showed up to the meeting. If we decide we do want to change the default when PrivateAssets="all", then I agree, but I don't think we ever finalized that.

marcpopMSFT:
I'm not quite following a sit sounds like 47332 isn't quite settled but also sounds like the current nuget docs are inaccurate. Would it make sense to fix the docs while continuing the ongoing discussion? I can't ping dsplaisted here to add it to his list for partner discussions.

PR #3319, opened 2024-07-20
Update NuGet Audit docs for 6.12

zivkan:
fixes: https://github.com/NuGet/Home/issues/13479

* Update the audit packages docs with info on all the NuGet 6.11 and 6.12 features, plus link to the NuGetAudit 2.0 blog.
   * Also add a mention of `dotnet list package --vulnerable` because the docs are under the "concepts" top level heading, and has a generic title, so the docs don't sound like it should be limited to the NuGetAudit on restore feature.
* Update the NU1901-NU1904 docs to mention the blog post, and `dotnet nuget why`
* Update NU1905. The warning code was originally used by the first implementation of NuGetAudit, but was removed before GA (although it did ship in the .NET SDK 8.0.100 and was removed in 8.0.101). so was unused until NuGet 6.12. Looks like I previously forgot to remove the docs from the ToC, hence why this PR doesn't add it to the ToC.
* Update nuget.config with audit sources info

### DIFF
--- a/docs/concepts/Auditing-Packages.md
+++ b/docs/concepts/Auditing-Packages.md
@@ -3,7 +3,7 @@ title: Auditing package dependencies for security vulnerabilities
 description: How to audit package dependencies for security vulnerabilities and acting on security audit reports.
 author: JonDouglas
 ms.author: jodou
-ms.date: 10/11/2023
+ms.date: 07/19/2024
 ms.topic: conceptual
 ---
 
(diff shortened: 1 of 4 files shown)


nkolev92 (approved):

docs/concepts/Auditing-Packages.md
```diff
@@ -16,24 +16,107 @@ This involves identifying vulnerabilities, evaluating risks, and making recommen
...
+Once a project is successfully restored, [`dotnet list package`](/dotnet/core/tools/dotnet-list-package) has a `--vulnerable` argument to filter the packages based on which packages have known vulnerabilities.
+Note that `--include-transitive` is not default, so should be included 
+
+## Actions when packages with known vulnerabilities are reported
+
+We also have a [blog post](https://devblogs.microsoft.com/nuget/nugetaudit-2-0-elevating-security-and-trust-in-package-management/) which discusses our recommended method for taking action when a package with a known vulnerability is found to be used by your project, and tools to help get more information.
+
 ### Security vulnerabilities found with updates
 
 If security vulnerabilities are found and updates are available for the package, you can either:
```
zivkan, line 122: What do other people think about this section of the docs (and the next 50 unmodified lines)?

It feels out of place to me. Maybe we should be doing more of it in our docs, but our docs are mostly an unopinionated reference to our features, whereas this is providing opinions on what to do.

Now that we have the blog post out, and I've linked to it in this page, should we remove this section from these docs?

I was also thinking that instead of having a single page on audit, it could be a section in the table of contents:

```
Auditing packages
+ Overview
+ Auditing during restore
+ Auditing with dotnet list package
```

And if we want to go down the route of having opinionated recomendations, there could also be "Recommendations for package consumers" which could contain the information I'm proposing to remove here (but also be a duplicate of the blog post).

Any comments?

zivkan: will consider for a different PR

PR #3320, opened 2024-07-22
Add CPM scenario for NU1604

martinrrm:
Fix: https://github.com/NuGet/Home/issues/13500

Added the scenario for users using CPM and NU1604 is raised.

### DIFF
--- a/docs/reference/errors-and-warnings/NU1604.md
+++ b/docs/reference/errors-and-warnings/NU1604.md
@@ -32,3 +32,22 @@ or
 > `<PackageReference Version="9.0.0" />`
 
 which implies a lower bound.
+
+If using [Central Package Managment](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management) you will need to update the `PackageVersion` `Version` attribute in your `Directory.Packages.Props` file.
+For example change from:
+
+> `<PackageVersion Include="PackageA" Version="(9.0.0, )" />`
+
+
+to:
+
+> `<PackageVersion Include="PackageA" Version="[9.0.0, )" />`
+
+or
+
+> `<PackageVersion Include="PackageA" Version="9.0.0" />`
+
+which implies a lower bound.
+
+> [!Note]
+> When using CPM and the file `Directory.Packages.Props` is invalid, NU1604 is raised.


docs/reference/errors-and-warnings/NU1604.md
```diff
@@ -32,3 +32,22 @@ or
...
+> `<PackageVersion Include="PackageA" Version="[9.0.0, )" />`
+
+or
+
+> `<PackageVersion Include="PackageA" Version="9.0.0" />`
+
+which implies a lower bound.
+
+> [!Note]
+> When using CPM and the file `Directory.Packages.Props` is invalid, NU1604 is raised.
```
martinrrm, line 53: I added this because the original issue is related to an invalid file but makes me think this is a bug instead of a docs issue. @jeffkl what do you think?

docs/reference/errors-and-warnings/NU1604.md
```diff
@@ -32,3 +32,22 @@ or
 > `<PackageReference Version="9.0.0" />`
 
 which implies a lower bound.
+
+If using [Central Package Managment](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management) you will need to update the `PackageVersion` `Version` attribute in your `Directory.Packages.Props` file.
```
nkolev92, line 36: Normally this would be a different scenario. 

We'd have another Issue, Solution etc. 

We should also call out that this is a special case, since in a properly onboarded CPM scenario, you'd get NU1008.

martinrrm: I wasn't sure if I should put it in another Solution section since it's the same error but yeah it's a special case so I can start with that

nkolev92: Here's prior from both restore and pack.

* https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/reference/errors-and-warnings/NU1605.md
* https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5128

donnie-msft, line 36: nit: my readability preference would be for something between these two words..
```suggestion
If using [Central Package Managment](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management) you will need to update the `PackageVersion` element's `Version` attribute in your `Directory.Packages.Props` file.
```

docs/reference/errors-and-warnings/NU1604.md
```diff
@@ -32,3 +32,22 @@ or
...
+For example change from:
+
+> `<PackageVersion Include="PackageA" Version="(9.0.0, )" />`
+
+
```
donnie-msft, line 41: extra blank line

docs/reference/errors-and-warnings/NU1604.md
```diff
@@ -32,3 +32,22 @@ or
 > `<PackageReference Version="9.0.0" />`
 
 which implies a lower bound.
+
+If using [Central Package Managment](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management) you will need to update the `PackageVersion` `Version` attribute in your `Directory.Packages.Props` file.
+For example change from:
```
donnie-msft, line 37:
```suggestion
For example, change from:
```

docs/reference/errors-and-warnings/NU1604.md
donnie-msft, line 36:
```suggestion
If using [Central Package Managment](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management), you will need to update the `PackageVersion` `Version` attribute in your `Directory.Packages.Props` file.
```

jeffkl:
Closing in favor of https://github.com/NuGet/docs.microsoft.com-nuget/pull/3350 which I'll incorporate these changes into

PR #3324, opened 2024-07-25
Document NU1702

nkolev92:
Fixes https://github.com/NuGet/Home/issues/13487


docs/reference/errors-and-warnings/NU1702.md
```diff
@@ -0,0 +1,23 @@
...
+ms.date: 07/25/2024
+ms.topic: reference
+f1_keywords: 
+  - "NU1702"
+---
+
+# NuGet Warning NU1702
+
+<pre>ProjectReference 'C:\Code\Project2.csproj'
+ was resolved using '.NETFramework,Version=v4.7.2' instead of the project target framework '.NETCoreApp,Version=v5.0'. This project may not be fully compatible with your project.
```
zivkan, line 14: viewing this with the docs preview link, the docs website doesn't wrap long lines and adds a horizontal scroll bar.  I think we should make the docs easier to read.

Nigusu-Allehu: We could do 
" > ProjectReference 'C:\Code\Project2.csproj'
 was resolved using '.NETFramework,Version=v4.7.2' instead of the project target framework '.NETCoreApp,Version=v5.0'. This project may not be fully compatible with your project.
" which would allow it to wrap. That's what I did for https://github.com/NuGet/docs.microsoft.com-nuget/pull/3325. But my worry is that it won't look the same as the other NuCodes we have. Maybe we modify all of them to look this.

zivkan: What do you mean by fragile?  learn's CSS has a max content width, no matter how wide your monitor. Maybe smaller screens scale down the width, but we can still edit the content to the known max width.

Alternatively, instead of using code/pre blocks, we could use `>` quotes, as the CSS lets those wrap. I know it's a change from the historic pattern, but if the historic pattern is hard to read, then maybe we should change.

nkolev92: I did not know about the learn's max content width. I'm not sure too many people are familiar either. I'd prefer a solution that doesn't rely on people remembering that (I'd expect they wouldn't since it's not even defined in our repo).

If there were any changes in pattern here, I think they should be done wholesale and make all of our log codes consistent.

Nigusu-Allehu (approved):

docs/reference/errors-and-warnings/NU1702.md
zivkan: > If there were any changes in pattern here, I think they should be done wholesale and make all of our log codes consistent.

Why? I don't understand the customer benefit.

I don't think that customers look at our docs frequently enough to notice if a few of the NU code pages have different formatting for the example message. Even if they did, would it matter?

If it's easier to read because it's not partially hidden my a horizontal scroll bar, it's still easier to read.

When we roll out features like package source mapping, nuget audit, etc, we do so incrementally, helping as many customers as quickly as we can, rather than trying to bake the feature until it has 100% coverage of our long term plan. For more implementation details, like analyzers, nullable annotations, etc, we also don't force the entire repo to change in one go.

nkolev92: Docs don't really carry the risks of regression that docs do. 

The other answer is, it's easy to do and we should do it all instead of piecemeal improvements. 

I'll check how it looks on this one and if it's ok, I'll do it everywhere. It'd bug me if it was all over the place.

docs/reference/errors-and-warnings/NU1702.md
```diff
@@ -0,0 +1,23 @@
...
+This project may not be fully compatible with your project.
+</pre>
+
+### Issue
+
+`AssetTargetFallback` was used to select assets from a package. The warning let users know that the assets may not be 100% compatible.
+
+### Solution
+
+Change the target framework to one of the projects to make them fully compatible.
```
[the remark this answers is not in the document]
nkolev92: I don't think this suggestion is correct. 
However, I think the current version could be better. 

I'll push a change.

docs/reference/errors-and-warnings/NU1702.md
nkolev92: I'll create a PR fixing them all.

nkolev92: https://github.com/NuGet/docs.microsoft.com-nuget/pull/3327

zivkan (approved):

docs/reference/Errors-and-Warnings.md
```diff
@@ -26,7 +26,7 @@ NuGet supports the following configuration properties.
 ## Errors
 
 | Group | Error Numbers |
-| --- | --- |
+|-------|---------------|
```
zivkan, line 29: I don't know why this is causing the docs bot to generate a warning, but it is warning, so please revert this change.

docs/reference/errors-and-warnings/NU1702.md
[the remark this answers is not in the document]
donnie-msft: Are you saying the TFM should match one of the project TFMs? If so, it's written as a possessive, so it can't just be "to one of the projects", it needs to show the possessive with the apostrophe. Disclaimer: I'm not an English major so I can be wrong :)

docs/reference/errors-and-warnings/NU1702.md
```diff
@@ -0,0 +1,23 @@
...
+> This project may not be fully compatible with your project.
+
+### Issue
+
+`AssetTargetFallback` was used to select assets from a package. The warning lets users know that the assets may not be 100% compatible.
+
+### Solution
+
+Change the target framework of the projects to make them fully compatible.
+Example: Change the target framework of Project2 or `.NET 5` or change Project2 to target both `.NET 5` and `.NET Framework 4.7.2`.
```
donnie-msft, line 23:
```suggestion
Example: Change the target framework of Project2 to `.NET 5`, or change Project2 to target both `.NET 5` and `.NET Framework 4.7.2`.
```

docs/reference/Errors-and-Warnings.md
nkolev92: This was actually my attempt at fixing it. I'll revert the change. No clue why the warning is showing up.

PR #3330, opened 2024-08-05
Release notes for 6.11

martinrrm:
Release notes for https://github.com/NuGet/Client.Engineering/issues/2907

### DIFF
--- a/docs/release-notes/NuGet-6.11.md
+++ b/docs/release-notes/NuGet-6.11.md
@@ -2,7 +2,7 @@
 title: NuGet 6.11 Release Notes
 description: Release notes for NuGet 6.11 including new features, bug fixes, and DCRs.
 author: martinrrm
-ms.date: 8/1/2024
+ms.date: 8/13/2024
 ms.topic: conceptual
 ---
 # NuGet 6.11 Release Notes
@@ -37,7 +37,7 @@ NuGet distribution vehicles:
 
 * IVsNuGetProjectUpdateEvents.ProjectUpdateStarted or ProjectUpdateFinished should only provide a list of files that will get changed. - [#13413](https://github.com/NuGet/Home/issues/13413)
 
-* `dotnet` commands should not output usage information on all errors - [#13251](https://github.com/NuGet/Home/issues/13251)
+* `dotnet nuget` commands should not output usage information on all errors - [#13251](https://github.com/NuGet/Home/issues/13251)
 
 * [DCR]: NuGet causes many ArgumentExceptions to be thrown & caught in VS - [#11535](https://github.com/NuGet/Home/issues/11535)
 
@@ -77,25 +77,25 @@ NuGet distribution vehicles:
 
 Thank you to all the contributors who helped make this NuGet release awesome!
 
-* [seclerp](https://github.com/NuGet/NuGet.Client/pull/5783)
+* [seclerp](https://github.com/seclerp)
   * [5783](https://github.com/NuGet/NuGet.Client/pull/5783) Fix `PackageSourceProvider.UpdatePackageSource` doesn't respect a value from `DisableTLSCertificateValidation` field
   * [5767](https://github.com/NuGet/NuGet.Client/pull/5767) Fix `PackageSourceProvider.UpdatePackageSource` doesn't respect a value from `AllowInsecureConnection` field
-* [mmitche](https://github.com/NuGet/NuGet.Client/pull/5749)
+* [mmitche](https://github.com/mmitche)
   * [5749](https://github.com/NuGet/NuGet.Client/pull/5749) Build NuGet from the VMR
   * [5752](https://github.com/NuGet/NuGet.Client/pull/5752) Fixup signing support conditional
-* [ToddGrun](https://github.com/NuGet/NuGet.Client/pull/5859)
+* [ToddGrun](https://github.com/ToddGrun)
   * [5859](https://github.com/NuGet/NuGet.Client/pull/5859) Reduce allocations in GetPackageInfo by conditionally calling GetNupkgMetadataPath
-* [omajid](https://github.com/NuGet/NuGet.Client/pull/5848)
+* [omajid](https://github.com/omajid)
   * [5848](https://github.com/NuGet/NuGet.Client/pull/5848) Dispose certificates on failure in SignatureUtility.GetCertificates
-* [SimonCropp](https://github.com/NuGet/NuGet.Client/pull/5842)
+* [SimonCropp](https://github.com/SimonCropp)
   * [5842](https://github.com/NuGet/NuGet.Client/pull/5842) remove redundant .GetTypeInfo()
-* [ryanmolden](https://github.com/NuGet/NuGet.Client/pull/5834)
+* [ryanmolden](https://github.com/ryanmolden)
   * [5834](https://github.com/NuGet/NuGet.Client/pull/5834) Eliminate first-chance ArgumentExceptions when querying a legacy project for properties that don't exist via DTE
-* [hickford](https://github.com/NuGet/NuGet.Client/pull/5743)
+* [hickford](https://github.com/hickford)
   * [5743](https://github.com/NuGet/NuGet.Client/pull/5743) Correct message in the case that a higher version of package is already installed
-* [ViktorHofer](https://github.com/NuGet/NuGet.Client/pull/5764)
+* [ViktorHofer](https://github.com/ViktorHofer)
   * [5764](https://github.com/NuGet/NuGet.Client/pull/5764) Use .NET SDK sourcelink integration
-* [NikolaMilosavljevic](https://github.com/NuGet/NuGet.Client/pull/5738)
+* [NikolaMilosavljevic](https://github.com/NikolaMilosavljevic)
   * [5738](https://github.com/NuGet/NuGet.Client/pull/5738) Disable CA2022 errors
-* [jv42](https://github.com/NuGet/NuGet.Client/pull/5717)
+* [jv42](https://github.com/jv42)
   * [5717](https://github.com/NuGet/NuGet.Client/pull/5717) Fixed NullReferenceException in ResolverComparer
--- a/docs/release-notes/NuGet-6.11.md
+++ b/docs/release-notes/NuGet-6.11.md
@@ -31,8 +31,6 @@ NuGet distribution vehicles:
 
 * `MSBuildRestoreUtility.GetRestoreAuditProperties` needs a breaking change to read `NuGetAuditSuppress` items - [#13313](https://github.com/NuGet/Home/issues/13313)
 
-* NuGetAudit should check transitive packages by default when the .NET 9 SDK is installed - [#13293](https://github.com/NuGet/Home/issues/13293)
-
 ### Issues fixed in this release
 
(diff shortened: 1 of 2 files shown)


docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,62 @@
...
+
+<sup>1</sup> Installed with Visual Studio 2022 with any .NET workload
+
+## Summary: What's New in 6.11
+
+* [Feature]: add dotnet nuget why to dotnet CLI - [#11943](https://github.com/NuGet/Home/issues/11943)
+
+### Breaking changes
+
+### Issues fixed in this release
```
donnie-msft, line 25: One of mine is missing as it didn't have a milestone set, which I've done now. Please add https://github.com/NuGet/NuGet.Client/pull/5766

nkolev92: Did https://github.com/NuGet/Entropy/tree/main/NuGetReleaseTool/NuGetReleaseTool/AddMilestoneCommand and https://github.com/NuGet/Entropy/tree/main/NuGetReleaseTool/NuGetReleaseTool/GenerateReleaseChangelogCommand catch this issue @martinrrm?
Did you get a change to run those? 

I'd expect that it'd find the issue Donnie is linking

docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,62 @@
...
+ms.topic: conceptual
+---
+
+# NuGet 6.11 Release Notes
+
+NuGet distribution vehicles:
+
+| NuGet version | Available in Visual Studio version | Available in .NET SDK(s) |
+|:---|:---|:---|
+| [**6.11**](https://nuget.org/downloads) | [Visual Studio 2022 version 17.11](https://visualstudio.microsoft.com/downloads/) | [<TODO: Full SDK Version>](https://dotnet.microsoft.com/download/dotnet/<TODO: SDKMajorMinorVersionOnly)<sup>1</sup> |
```
zivkan, line 15: NuGet 6.11 maps to .NET SDK 8.0.400

docs/release-notes/NuGet-6.11.md
zivkan: Most of the NuGetAuditSuppress issues were in the private repo, so I just created this one for us to include in the release notes:
* https://github.com/NuGet/Home/issues/13679

martinrrm: @nkolev92 forgot to run that, let me add the milestone to all of those

martinrrm:
@zivkan @donnie-msft @nkolev92 I updated the issues with the correct milestone, let me know if there are some that are missing.

docs/release-notes/NuGet-6.11.md
martinrrm: The command was failing because there are issues with no description, II created a PR to fix that https://github.com/NuGet/Entropy/pull/118

donnie-msft (approved):

docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,101 @@
...
+
+* `MSBuildRestoreUtility.GetRestoreAuditProperties` needs a breaking change to read `NuGetAuditSuppress` items - [#13313](https://github.com/NuGet/Home/issues/13313)
+
+* NuGetAudit should check transitive packages by default when the .NET 9 SDK is installed - [#13293](https://github.com/NuGet/Home/issues/13293)
+
+### Issues fixed in this release
+
+* IVsNuGetProjectUpdateEvents.ProjectUpdateStarted or ProjectUpdateFinished should only provide a list of files that will get changed. - [#13413](https://github.com/NuGet/Home/issues/13413)
+
+* `dotnet` commands should not output usage information on all errors - [#13251](https://github.com/NuGet/Home/issues/13251)
```
zivkan, line 40:
```suggestion
* `dotnet nuget` commands should not output usage information on all errors - [#13251](https://github.com/NuGet/Home/issues/13251)
```

zivkan (approved):

docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Summary: What's New in 6.11
+
+* Suppress NuGetAudit warnings for specific advisories for PackageReference projects - [#13679](https://github.com/NuGet/Home/issues/13679)
+
+* Add `--allow-insecure-connections` option to dotnet SDK - [#13396](https://github.com/NuGet/Home/issues/13396)
+
+* Swap authors for owners in Visual Studio Package Manager UI - [#12501](https://github.com/NuGet/Home/issues/12501)
+
+* [Feature]: add dotnet nuget why to dotnet CLI - [#11943](https://github.com/NuGet/Home/issues/11943)
```
nkolev92, line 26: Is this in .NET 8.0.4xx?

docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,101 @@
...
+
+* NuGet cannot restore from HTTPS sources that have SSL certificate problems - [#4387](https://github.com/NuGet/Home/issues/4387)
+
+### Breaking changes
+
+* `MSBuildRestoreUtility.GetRestoreAuditProperties` needs a breaking change to read `NuGetAuditSuppress` items - [#13313](https://github.com/NuGet/Home/issues/13313)
+
+* NuGetAudit should check transitive packages by default when the .NET 9 SDK is installed - [#13293](https://github.com/NuGet/Home/issues/13293)
```
nkolev92, line 34: This one is tricky. It technically matches 9.0 and VS 17.12, but it's something that could be helpful for people to know.n

docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,101 @@
...
+
+* Vulnerability InfoBar remains visible in the Solution Explorer after closing solution - [#13055](https://github.com/NuGet/Home/issues/13055)
+
+[List of commits in this release](https://github.com/NuGet/NuGet.Client/compare/6.11.0.122...6.10.1.5)
+
+### Community contributions
+
+Thank you to all the contributors who helped make this NuGet release awesome!
+
+* [seclerp](https://github.com/NuGet/NuGet.Client/pull/5783)
```
nkolev92, line 80: I have a bug in the tool here. This should be the alias of the user.

docs/release-notes/NuGet-6.11.md
zivkan: yes, it is.

https://github.com/NuGet/NuGet.Client/commit/709d913595c39d14b98dbcadfdcc96bf34c712b0

https://github.com/NuGet/NuGet.Client/commits/6.11.0.86/

docs/release-notes/NuGet-6.11.md
zivkan: yeah, I wasn't sure about commenting on this one either.

I don't think there are any scenarios where this really matters to customers. Even if the customer is using nuget.exe, I don't think it matters if they're using nuget.exe 6.11 or 6.10, in both cases I'd expect the project to load the SDK's nuget.targets, not nuget.exe's targets, and therefore it's still the version of the SDK that matters, not the version of NuGet.

But the change to nuget.targets, that adds the version check for the .NET 9 SDK, is added to nuget 6.11. 🤷

docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,101 @@
...
+
+* SignatureUtility.GetCertificates in NuGet.Client can skip calling Dispose on error - [#13535](https://github.com/NuGet/Home/issues/13535)
+
+* 'dotnet nuget why' does not work when a directory is provided for the 'Path' argument - [#13527](https://github.com/NuGet/Home/issues/13527)
+
+* NuGet IntelliCode Package Suggestions are missing Author in packages list - [#13515](https://github.com/NuGet/Home/issues/13515)
+
+* Rebuilding in VS causes unnecessary restores - [#13505](https://github.com/NuGet/Home/issues/13505)
+
+* MSB4181: The "Restore Task" task returned false but did not log an error. - [#13460](https://github.com/NuGet/Home/issues/13460)
```
zivkan, line 58: The commit that closed this issue is https://github.com/NuGet/NuGet.Client/commit/de9c314080687bfe4da4baf4b3d1a9d38b6ace3a

But this commit was reverted: https://github.com/NuGet/NuGet.Client/commit/81b4565544b0f3a90f4527b550ff3318104b1370

What's the status? Did we fix "the restore task returned false but did not log an error"?

zivkan: looks like the revert was not part of any branch (probably a private branch that was deleted). @nkolev92 can you confirm?

nkolev92: > looks like the revert was not part of any branch (probably a private branch that was deleted). @nkolev92 can you confirm?

It's not. 
Just some testing.

docs/release-notes/NuGet-6.11.md
nkolev92: > Even if the customer is using nuget.exe, I don't think it matters if they're using nuget.exe 6.11 or 6.10, in both cases I'd expect the project to load the SDK's nuget.targets, not nuget.exe's targets, and therefore it's still the version of the SDK that matters, not the version of NuGet

I actually think it's opposite. In fact, I'm pretty sure it is: https://github.com/NuGet/NuGet.Client/blob/9fbabf9a119599b62824f676407c5611c0542781/src/NuGet.Clients/NuGet.CommandLine/MsBuildUtility.cs#L263.

You *still* need .NET 9 of the SDK. 

It almost feels like this particular issue should be referenced in both 6.11 and 6.12 since 6.12 matches .NET 9.

martinrrm: Since it's in the breaking change section I think leave it is better

nkolev92: Should we call this a feature instead of a breaking change? 

@zivkan Our breaking change section was meant to be for APIs. This is more of an addition/roll out.

docs/release-notes/NuGet-6.11.md
```diff
@@ -0,0 +1,99 @@
+---
+title: NuGet 6.11 Release Notes
```
nkolev92, line 2: Add the release notes to the TOC.

nkolev92 (approved):

docs/release-notes/NuGet-6.11.md
zivkan: Sorry, I forgot to reply to this earlier. In case it's still useful, we can continue the conversation

> Our breaking change section was meant to be for APIs

Where was this communicated? If this is not obvious to customers reading the release notes, then I disagree that we should keep it for API breaking changes only. If a customer is experiencing problems with a new release, it's faster for them to check the release notes and look at a short list of breaking changes, compared to (what's hopefully) a longer list of features.

nkolev92: I feel like the new feature section gets more prominence the breaking changes section by virtue of headings. 
I added the original breaking changes section and that was pretty much my interpretation.

Another thing is that transitive vulnerabilities aren't a traditional breaking chnage, they're a new feature. I'd rather have the precedent where we can make calls like this. 

Honestly, having a flagship feature in the breaking changes section doesn't do it the service it deserves it.

PR #3331, opened 2024-08-08
Update NU1008.md

RussKie:
Explain how to override package version

### DIFF
--- a/docs/reference/errors-and-warnings/NU1008.md
+++ b/docs/reference/errors-and-warnings/NU1008.md
@@ -29,17 +29,33 @@ In your project file, you may see:
 - Remove the version from the PackageId PackageReference.
 - You may need to add or update the PackageVersion item for PackageId in Directory.Packages.props
 
-Example:
+    Example:
 
-```xml
-<!-- In the project file. -->
-<PackageReference Include="PackageId" />
-```
+    ```xml
+    <!-- In the project file. -->
+    <PackageReference Include="PackageId" />
+    ```
 
-```xml
-<!-- In the Directory.Packages.props -->
-<PackageVersion Include="PackageId" Version="5.1.0" />
-```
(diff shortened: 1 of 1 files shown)


zivkan (approved):

docs/reference/errors-and-warnings/NU1008.md
```diff
@@ -29,17 +29,33 @@ In your project file, you may see:
...
+    ```
+
+- Alternatively, you may override an individual package version by using the `VersionOverride` property on a `<PackageReference />` item.
+This overrides any `<PackageVersion />` defined centrally.
+
+    Example:
+
+    ```xml
+    <!-- In the project file. -->
+    <PackageReference Include="PackageId" VersionOverride="3.0.0" />
```
nkolev92, line 51: I think there's a chance this might've been omitted on purpose. 

While the feature is available, it's not considered a best practice. 

cc @jeffkl should have more context on this.

RussKie: > I think there's a chance this might've been omitted on purpose.

We shouldn't be hiding information. It just plain wrong.

> While the feature is available, it's not considered a best practice.

There are very valid scenarios for this feature - e.g., solutions targeting multiple TFMs.

nkolev92: There's nothing "wrong" in here by not mentioning VersionOverride. The breadth of solutions offered here doesn't make it wrong. 

> There are very valid scenarios for this feature - e.g., solutions targeting multiple TFM

PackageVersion can be conditioned as well. 

I'm not against accepting a changing like this, but I don't agree that not having is incorrect.

nkolev92 (approved):

RussKie:
7 months... not bad :)

PR #3332, opened 2024-08-10
Update NuGetAuditSuppress info in NU1901-4 docs

zivkan:
fixes: https://github.com/NuGet/Client.Engineering/issues/2899

I checked the auditing packages page already has the good information, so it doesn't need an update, just the NU1901-4 warning code page.

### DIFF
--- a/docs/reference/errors-and-warnings/NU1901-NU1904.md
+++ b/docs/reference/errors-and-warnings/NU1901-NU1904.md
@@ -47,14 +47,7 @@ For example, `<NuGetAuditLevel>high</NuGetAuditLevel>`.
 
 If you would like to suppress a specific advisory, add an MSBuild [NuGetAuditSuppress](../../concepts/Auditing-Packages.md#excluding-advisories) item.
 For example `<NuGetAuditSuppress Include="https://github.com/advisories/GHSA-g3q9-xf95-8hp5" />`.
+`NuGetAuditSuppress` is available from VS 17.11 and .NET 8.0.400 SDK for projects using `PackageReference`, and from VS 17.12 for projects using `packages.config`.
 
 If you do not want NuGet to check for packages with known vulnerabilities during restore, add `<NuGetAudit>false</NuGetAudit>` inside a `<PropertyGroup>` in your project file, or a [`Directory.Build.props` file](/visualstudio/msbuild/customize-by-directory).
 If you would like to run NuGet Audit on developer machines, but disable it on CI pipelines, you can take advantage of MSBuild importing environment variables, and create a NuGetAudit environment variable set to `false` in your pipeline definition.
-
-> [!NOTE]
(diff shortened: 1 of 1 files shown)


docs/reference/errors-and-warnings/NU1901-NU1904.md
```diff
@@ -47,14 +47,7 @@ For example, `<NuGetAuditLevel>high</NuGetAuditLevel>`.
 
 If you would like to suppress a specific advisory, add an MSBuild [NuGetAuditSuppress](../../concepts/Auditing-Packages.md#excluding-advisories) item.
 For example `<NuGetAuditSuppress Include="https://github.com/advisories/GHSA-g3q9-xf95-8hp5" />`.
+`NuGetAuditSuppress` is available from VS 17.11 and .NET 8.0.400 SDK for projects using `PackageReference`, and from VS 17.12 for projects using `packages.config`.
```
nkolev92, line 50: Link the release notes maybe?

nkolev92 (approved):

docs/reference/errors-and-warnings/NU1901-NU1904.md
```diff
@@ -47,14 +47,7 @@ For example, `<NuGetAuditLevel>high</NuGetAuditLevel>`.
...
 
 If you do not want NuGet to check for packages with known vulnerabilities during restore, add `<NuGetAudit>false</NuGetAudit>` inside a `<PropertyGroup>` in your project file, or a [`Directory.Build.props` file](/visualstudio/msbuild/customize-by-directory).
 If you would like to run NuGet Audit on developer machines, but disable it on CI pipelines, you can take advantage of MSBuild importing environment variables, and create a NuGetAudit environment variable set to `false` in your pipeline definition.
-
-> [!NOTE]
```
kartheekp-ms, line 54: I am just wondering if we need this note for customers using VS version < 17.11 or .NET SDK < 8.0.400 just in case.

kartheekp-ms (approved):

aortiz-msft (approved):

docs/reference/errors-and-warnings/NU1901-NU1904.md
zivkan: NuGetAuditSuppress is documented in multiple places, including in this document, about which versions it's available from, so I don't think it's necessary to have an additional note about from which version it's available.

About the NoWarn suggestion, it's not a recommended mitigation, it was just the best that was available at the time. It was a better alternative than customers deciding to disable NuGetAudit completely (and then forget to re-enable when suppressions was implemented).  For customers who aren't aware of NoWarn (or forget about it when trying to deal with one of these warnings), I think it's better to encourage them to upgrade to a newer version of the .NET SDK or Visual Studio, rather than add a NoWarn. Particularly when customers are using CPM, or specify package versions in another file, the chance of customers remembering to remove the NoWarn when they upgrade to a newer package version that no longer has a known vulnerability is quite low.  Even we rarely re-evaluate NoWarns in NuGet.Client.

PR #3335, opened 2024-08-27
Clarify mono isn't supported

JonDouglas:


zivkan (approved):
[There are 35 files in this docs repo that mention mono](https://github.com/search?q=repo%3ANuGet%2Fdocs.microsoft.com-nuget%20mono&type=code). This is a good first step, but I think we need to change wording on other pages too, for example, [the page on installing/obtaining nuget](https://github.com/NuGet/docs.microsoft.com-nuget/blob/50aa4c833406887f42d5d3131dfb3419d47878cd/docs/install-nuget-client-tools.md?plain=1#L29).  For better or worse, about half the mono search results look to be release notes, and I don't think we need to make any changes there, so I think the effort required is smaller than the 35 files with results makes it sound to be.

docs/resources/NuGet-FAQ.yml
```diff
@@ -23,11 +23,9 @@ sections:
...
           Does NuGet support Mono?
         answer: |
-          The command-line tool, `nuget.exe`, builds and runs under Mono 3.2+ and can create packages in Mono.
-          
-          Although `nuget.exe` works fully on Windows, there are known issues on Linux and OS X. Refer to [Mono issues](https://github.com/NuGet/Home/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+mono) on GitHub.
-          
-          A [graphical client](https://github.com/mrward/monodevelop-nuget-addin) is available as an add-in for MonoDevelop.
+          The command-line tool, `nuget.exe`, builds and runs typically under Windows. NuGet can run on Unix operating systems using `mono`, but it is not officially supported by [NuGet's Support Policy](https://devblogs.microsoft.com/nuget/announcing-nuget-exe-and-nuget-client-sdk-packages-support-policy-keeping-you-informed-and-secure/).
+
+          [Mono has transfered ownership to Wine and is no longer maintained by Microsoft](https://github.com/mono/mono/issues/21796).
```
jebriede, line 28: Typo

```suggestion
          [Mono has transferred ownership to Wine and is no longer maintained by Microsoft](https://github.com/mono/mono/issues/21796).
```

JonDouglas:
> [There are 35 files in this docs repo that mention mono](https://github.com/search?q=repo%3ANuGet%2Fdocs.microsoft.com-nuget%20mono&type=code). This is a good first step, but I think we need to change wording on other pages too, for example, [the page on installing/obtaining nuget](https://github.com/NuGet/docs.microsoft.com-nuget/blob/50aa4c833406887f42d5d3131dfb3419d47878cd/docs/install-nuget-client-tools.md?plain=1#L29). For better or worse, about half the mono search results look to be release notes, and I don't think we need to make any changes there, so I think the effort required is smaller than the 35 files with results makes it sound to be.

I agree that there are no changes really needed to be made. Older mono versions will still exist and be relevant. We can add notes later

PR #3336, opened 2024-08-28
Add transitive guidance

JonDouglas:
/cc @ericstj

### DIFF
--- a/docs/concepts/Auditing-Packages.md
+++ b/docs/concepts/Auditing-Packages.md
@@ -129,9 +129,9 @@ If security vulnerabilities are found and updates are available for the package,
 
 If a known vulnerability exists in a top-level package's transitive dependencies, you have these options:
 
-- Add the fixed package version as a direct package reference. **Note:** Be sure to remove this reference when a new package version update becomes available.
+- Add the fixed package version as a direct package reference. **Note:** Be sure to remove this reference when a new package version update becomes available and be sure to maintain the defined attributes for the expected behavior.
 - Use [Central Package Management with the transitive pinning functionality](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management#transitive-pinning).
-- [Suppress the advisory](https://learn.microsoft.com/nuget/concepts/auditing-packages#excluding-advisories) until a package version update becomes available.
+- [Suppress the advisory](https://learn.microsoft.com/nuget/concepts/auditing-packages#excluding-advisories) until it can be addressed.
 - File an issue in the top-level package's tracker to request an update.
 
 ### Security vulnerabilities found with no updates
--- a/docs/concepts/Auditing-Packages.md
+++ b/docs/concepts/Auditing-Packages.md
@@ -125,6 +125,15 @@ If security vulnerabilities are found and updates are available for the package,
 - Use the NuGet package manager user interface in Visual Studio to update the individual package.
 - Run the `dotnet add package` command with the respective package ID to update to the latest version.
 
+#### Transitive Packages
+
+If a known vulnerability exists in a top-level package's transitive dependencies, you have these options:
+
+- Add the fixed package version as a direct package reference. **Note:** Be sure to remove this reference when a new package version update becomes available.
+- Use [Central Package Management with the transitive pinning functionality](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management#transitive-pinning).
+- [Suppress the advisory](https://learn.microsoft.com/nuget/concepts/auditing-packages#excluding-advisories) until a package version update becomes available.
+- File an issue in the top-level package's tracker to request an update.
+
 ### Security vulnerabilities found with no updates
 
 In the case that a known vulnerability exists in a package without a security fix, you can do the following.


docs/concepts/Auditing-Packages.md
```diff
@@ -125,6 +125,15 @@ If security vulnerabilities are found and updates are available for the package,
 - Use the NuGet package manager user interface in Visual Studio to update the individual package.
 - Run the `dotnet add package` command with the respective package ID to update to the latest version.
 
+#### Transitive Packages
+
+If a known vulnerability exists in a top-level package's transitive dependencies, you have these options:
+
+- Add the fixed package version as a direct package reference. **Note:** Be sure to remove this reference when a new package version update becomes available.
```
ericstj, line 132: We should also tell folks to take care when doing this that they don't change the deployment behavior of their project.  I've seen folks hoist a dependency that was marked with PrivateAssets and ExcludeAssets and forget to add those same markings.

docs/concepts/Auditing-Packages.md
```diff
@@ -125,6 +125,15 @@ If security vulnerabilities are found and updates are available for the package,
...
+- Use [Central Package Management with the transitive pinning functionality](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management#transitive-pinning).
```
ericstj, line 133: Can we make this the first suggestion?

docs/concepts/Auditing-Packages.md
```diff
@@ -125,6 +125,15 @@ If security vulnerabilities are found and updates are available for the package,
...
+- [Suppress the advisory](https://learn.microsoft.com/nuget/concepts/auditing-packages#excluding-advisories) until a package version update becomes available.
+- File an issue in the top-level package's tracker to request an update.
```
ericstj, line 135: Neither of these should be recommended as alternatives to the first.

A suppression should only be used if for some reason the project owner cannot update a dependency.  For example - they aren't responsible for deploying the dependency and need to keep the version pinned to the version some host process is responsible for deploying.

Blocking on the external package to update is not recommended either.  They should inform the package owner that they'd like an update, but they shouldn't block on it.

docs/concepts/Auditing-Packages.md
JonDouglas: Something like?

Note: This does not change the deployment behavior in any way.

docs/concepts/Auditing-Packages.md
JonDouglas: Holding back on that until CPM is the default in the future.

docs/concepts/Auditing-Packages.md
JonDouglas: hopefully these read in the sense that they aren't mutually exclusive. just options people have and should consider.

some of these can be combined and enough options exist to not block.

docs/concepts/Auditing-Packages.md
ericstj: No, I mean the user needs to *take care* to make sure that they don't do this.  So if they were previously doing:
```xml
<PackageRefence Include="Direct" Version="1.0.0" PrivateAssets="All" ExcludeAssets="Runtime" />
```

They should take care when adding a new reference to give it the same:
```xml
<PackageRefence Include="Direct" Version="1.0.0" PrivateAssets="All" ExcludeAssets="Runtime" />
<!-- Note: Apply the same values for PrivateAssets and ExcludeAssets as Direct --> 
<PackageRefence Include="Indirect" Version="1.1.0" PrivateAssets="All" ExcludeAssets="Runtime"/>
```

It's not automatic here.  It is for CPM - that's a major benefit it gives.

docs/concepts/Auditing-Packages.md
ericstj: I feel strongly that we cannot suggest this:
> [Suppress the advisory](https://learn.microsoft.com/nuget/concepts/auditing-packages#excluding-advisories) until a package version update becomes available.

That implies that the user is blocked and should suppress.  That's not the case.

docs/concepts/Auditing-Packages.md
JonDouglas: ah i see. you're talking about the attribute behavior.

@jeffkl @zivkan should we explicitly mention this?

ericstj: Maybe we say: Take care when adding the dependency to match the [Assets settings](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets) of the existing dependency.  When doing so it should behave in the same way as if your dependency itself was updated.  This is not required when using CPM.

JonDouglas: yep. "whne adding a explicit reference, ensure you maintain your packagereference attributes for the asset behavior" etc

docs/concepts/Auditing-Packages.md
JonDouglas: ill work to rephrase, the idea is this:

a developer sees an advisory, they look, they don't assess they are affected, they suppress it, and next time they update packages they get a non-vulnerable one and can remove the suppression.

similar to the promotion of top-level dependency flow.

this helps them unblock in the case they may have warningsaserrors and need to make a choice of promotion or suppression.

docs/concepts/Auditing-Packages.md
zivkan, line 133: I think we should explicitly warn people that if they pack the project into a package, the pinned package will become a package dependency. The linked doc mentions that it promotes a package to a top level dependency, but the very first two customers who tested the first CPM prototype didn't understand this, complained, and stopped using CPM.

docs/concepts/Auditing-Packages.md
zivkan: I'm confused, or I think we have inconsistent behaviour (which now would be a breaking change to fix).  Today, so coincidence it's on the same day, I was pinged on Discord because someone claims that a package reference with `ExcludeAssets="runtime"` does not exclude that package's dependencies runtime assets, and they had to manually list every transitive package as direct references and add the same `ExcludeAssets="runtime"`.

If CPM's central pinning does flow through include/private/exclude assets, then it sounds like it changes behaviour compared to the same package not being pinned, which feel like a bug to me.

docs/concepts/Auditing-Packages.md
ericstj: That probably belongs in the CPM docs.  Remind folks that rolling up dependencies doesn't actually add any new dependencies than what you already have.  Maybe NuGet gallery could even help here by showing authored dependencies visually different than transitive ones?

docs/concepts/Auditing-Packages.md
ericstj: Here's a good example of the sort of thing that makes this complicated: https://github.com/dotnet/aspnetcore/pull/57560#issuecomment-2321908053

docs/concepts/Auditing-Packages.md
ericstj: Rather than say `until a package version update becomes available.` you could say `until it can be addressed.`  This means that addressing it might happen in one of many ways

docs/concepts/Auditing-Packages.md
zivkan: nuget.org (NuGetGallery) only show dependencies listed in the nuspec, so only direct dependencies, not transitive dependencies.

Any package that CPM promotes via transitive pinning becomes a direct dependency (the page on CPM explicitly states this), and therefore it will be listed in the nuspec as a dependency.  NuGet Gallery has no way to know that "this was a transitive package that was pinned by the project that it was packed from". The very first 2 customers who tried CPM (both package authors) when it was in preview and transitive pinning was default with no option to disable, immediately said they hated it and stopped using CPM. So, I think it's valuable to set expectations.

I still see transitive pinning (or direct references) as a valid way for library authors to remove transitive packages with known vulnerabilities. But if a single repo is building both packages and apps, the developer might not want to cause a transitive package like System.Text.Json to become a direct dependency of their package, when they upgrade the package for the web/gui/console apps.

nkolev92: There's another challenge here that the CPM docs aren't really complete in this scenario. 
Separate problem: https://github.com/NuGet/Home/issues/13811.

zivkan: I don't know what I was thinking with my last reply.  Let me try again to make it more clear:

Even if it's documented elsewhere, the first set of customers who tried transitive pinning hated it because it caused additional package dependencies, and told us they immediately stopped using it.  I'm sure those two people were not unique and other customers will feel similarly.  Therefore, if we're going to advise people to use the feature, I think we should set expectations so they're not surprised.

nkolev92: Oh yeah fair, I was just pointing out that we have another gap to close, which is document it for CPM in the first place.

JonDouglas:
bump. still waiting an approval to get these initial items included officially in the docs.

docs/concepts/Auditing-Packages.md
```diff
@@ -125,6 +125,15 @@ If security vulnerabilities are found and updates are available for the package,
...
 - Run the `dotnet add package` command with the respective package ID to update to the latest version.
 
+#### Transitive Packages
+
+If a known vulnerability exists in a top-level package's transitive dependencies, you have these options:
+
+- Add the fixed package version as a direct package reference. **Note:** Be sure to remove this reference when a new package version update becomes available and be sure to maintain the defined attributes for the expected behavior.
+- Use [Central Package Management with the transitive pinning functionality](https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management#transitive-pinning).
+- [Suppress the advisory](https://learn.microsoft.com/nuget/concepts/auditing-packages#excluding-advisories) until it can be addressed.
+- File an issue in the top-level package's tracker to request an update.
```
kartheekp-ms, line 135: Maybe we could add a link to `dotnet nuget why command` or `Visual Studio transitive dependencies` feature docs for customers on how to identify the top-level package.

kartheekp-ms (approved):

docs/concepts/Auditing-Packages.md
JonDouglas: @jebriede solution explorer ships what release again?

jebriede: @JonDouglas do you mean transitive dependencies in Solution-level PM UI? If so, 17.12.

JonDouglas:
I'm moving forward to merge this. Let's please address whatever other items in follow-up as my intentions here are to include basic transitive guidance and not every specific.

For example, please go update the CPM docs on specific "gotchas" and any last resort auditing advice in future PRs.

PR #3338, opened 2024-09-10
Publish Live

Mikejo5000:


Mikejo5000:
@zivkan Andy, are your changes OK to publish live?

zivkan (approved):
> are your changes OK to publish live

yes. Frankly I don't understand the purpose of having separate live and main branches. If we wanted to stage changes until a particular time, we could have used a feature branch 🤷 

I feel like the different live & main branches unnecessarily delays useful docs updates.

PR #3341, opened 2024-09-21
Update secure supply chain doc with latest NuGetAudit info

zivkan:
The first commit splits also multi-sentence lines into one sentence per line, so that future updates will have easier to review diffs. Therefore, I highly recommend that you view the changes per commit, or all the commits, excluding the first one.

I added some info about NuGetAudit into the doc, as well as some additions following our recent threat modelling review.

### DIFF
--- a/docs/concepts/Security-Best-Practices.md
+++ b/docs/concepts/Security-Best-Practices.md
@@ -3,7 +3,7 @@ title: Best practices for a secure software supply chain
 description: Best practices for securing your software supply chain using NuGet & GitHub.
 author: JonDouglas
 ms.author: jodou
-ms.date: 04/09/2024
+ms.date: 20/09/2024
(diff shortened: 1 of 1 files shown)


docs/concepts/Security-Best-Practices.md
```diff
@@ -3,62 +3,96 @@ title: Best practices for a secure software supply chain
 description: Best practices for securing your software supply chain using NuGet & GitHub.
 author: JonDouglas
 ms.author: jodou
-ms.date: 02/08/2021
```
zivkan, line 6: see: https://github.com/NuGet/docs.microsoft.com-nuget/pull/3341#issuecomment-2364776821

it says:

> If you don't specify ms.date, the date will be updated automatically.

nkolev92 (approved):

PR #3339, opened 2024-09-11
add docs for new NuGet warning NU3043

kartheekp-ms:
Tracking: https://github.com/NuGet/Client.Engineering/issues/2994

Starting with .NET 9 Preview 7 and NuGet.exe 6.12, NuGet sign commands will raise a `NU3043` warning if an invalid value or SHA-1 hash is passed for the certificate fingerprint option. The value must be a SHA-256, SHA-384, or SHA-512 certificate fingerprint (in hexadecimal).

PR #3342, opened 2024-09-24
deprecate SHA-1 usage in nuget.exe sign command

kartheekp-ms:
Tracking: https://github.com/NuGet/Client.Engineering/issues/2994

Starting with .NET 9 Preview 7 and NuGet.exe 6.12, NuGet sign commands will raise a NU3043 warning if an invalid value or SHA-1 hash is passed for the certificate fingerprint option. The value must be a SHA-256, SHA-384, or SHA-512 certificate fingerprint (in hexadecimal).

### DIFF
--- a/docs/reference/cli-reference/cli-ref-sign.md
+++ b/docs/reference/cli-reference/cli-ref-sign.md
@@ -29,7 +29,7 @@ where `<package(s)>` is one or more `.nupkg` files.
 
 - **`-CertificateFingerprint`**
 
-  Specifies the SHA-1 fingerprint of the certificate used to search a local certificate store for the certificate.
+  Specifies the fingerprint of the certificate used to search a local certificate store for the certificate.
 
   > [!NOTE]
   > Starting with `NuGet.exe 6.12`, a `NU3043` warning is raised when a SHA-1 certificate fingerprint is passed.
--- a/docs/reference/cli-reference/cli-ref-sign.md
+++ b/docs/reference/cli-reference/cli-ref-sign.md
@@ -29,7 +29,11 @@ where `<package(s)>` is one or more `.nupkg` files.
 
 - **`-CertificateFingerprint`**
 
-  Specifies the SHA-1 fingerprint of the certificate used to search a local certificate store for the certificate.
+  Specifies the fingerprint of the certificate used to search a local certificate store for the certificate.
+
+  > [!NOTE]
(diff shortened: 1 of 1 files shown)


docs/reference/cli-reference/cli-ref-sign.md
```diff
@@ -29,7 +29,12 @@ where `<package(s)>` is one or more `.nupkg` files.
 
 - **`-CertificateFingerprint`**
 
-  Specifies the SHA-1 fingerprint of the certificate used to search a local certificate store for the certificate.
+  Specifies the fingerprint of the certificate used to search a local certificate store for the certificate.
+
+  Starting with NuGet.exe 6.12, this option can be used to specify the SHA-1, SHA-256, SHA-384, or SHA-512 fingerprint of the certificate.
+  However, a `NU3043` warning is raised when a SHA-1 certificate fingerprint is used because it is no longer considered secure.
+
+  All the previous versions of the NuGet.exe continue to accept only SHA-1 certificate fingerprint.
```
donnie-msft, line 37: Consider putting this as a !NOTE to call it out more strongly.

docs/reference/cli-reference/cli-ref-sign.md
```diff
@@ -29,7 +29,12 @@ where `<package(s)>` is one or more `.nupkg` files.
 
 - **`-CertificateFingerprint`**
 
-  Specifies the SHA-1 fingerprint of the certificate used to search a local certificate store for the certificate.
+  Specifies the fingerprint of the certificate used to search a local certificate store for the certificate.
```
donnie-msft, line 32: I find this a little hard to read, can it just be "fingerprint" instead of "fingerprint of the cert" ?

```suggestion
  Specifies the fingerprint to be used to search for the certificate in a local certificate store.
```

donnie-msft (approved):

docs/reference/cli-reference/cli-ref-sign.md
kartheekp-ms: In https://github.com/NuGet/docs.microsoft.com-nuget/pull/3339#discussion_r1756275821, Andy pointed me to the following guidance from [docs' contribution guide](https://learn.microsoft.com/contribute/content/markdown-reference#alerts-note-tip-important-caution-warning).

> Avoid notes, tips, and important boxes. Readers tend to skip over them. It's better to put that info directly into the article text.

I am not sure if adding note block will have an impact on readers based on above guidance. Maybe we will merge this PR as is and update docs if needed in the future.

PR #3346, opened 2024-10-09
Append warning-behavior properties

rainersigwald:
Specifying the current value for the property is a best practice to avoid accidentally overwriting an existing value.


rainersigwald (commented):
@zivkan one small nit to the very nice how-to-keep-these-as-warnings doc section.

jeffkl (approved):

zivkan (approved):

nkolev92 (approved):

PR #3353, opened 2024-11-13
Add best practice to Package-Versioning.md

ericstj:
@JonDouglas @nkolev92

### DIFF
--- a/docs/concepts/Package-Versioning.md
+++ b/docs/concepts/Package-Versioning.md
@@ -118,7 +118,7 @@ When referring to package dependencies, NuGet supports using interval notation f
 
 ### Best Practice
 
-Always specify a version or version range for package dependencies in project files, `packages.config` files, and `.nuspec` files. Without a version or version range, NuGet 2.8.x and earlier chooses the latest available package version when resolving a dependency, whereas NuGet 3.x and later chooses the lowest package version. Specifying a version or version range avoids this uncertainty.
+Always specify a version or version range for package dependencies in project files, `packages.config` files, and `.nuspec` files. Without a version or version range, when resolving a dependency, consistent restore results are not guaranteed.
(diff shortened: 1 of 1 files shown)


nkolev92 (approved):

docs/concepts/Package-Versioning.md
```diff
@@ -116,9 +116,10 @@ When referring to package dependencies, NuGet supports using interval notation f
 | [1.0,2.0) | 1.0 ≤ x < 2.0 | Mixed inclusive minimum and exclusive maximum version |
 | (1.0)    | invalid | invalid |
 
-### Examples
+### Best Practice
 
 Always specify a version or version range for package dependencies in project files, `packages.config` files, and `.nuspec` files. Without a version or version range, NuGet 2.8.x and earlier chooses the latest available package version when resolving a dependency, whereas NuGet 3.x and later chooses the lowest package version. Specifying a version or version range avoids this uncertainty.
```
aortiz-msft, line 121: NuGet 2.8.x is out of support. Let's remove the whole of line 121.

aortiz-msft (changes requested):

docs/concepts/Package-Versioning.md
ericstj: It's still important to specify a version.  I know PackageReference requires it.  I see nuspecs still allow dependencies without a version.  I think this can be reworded and avoid the mention of 2.8.x.

nkolev92: How about: 

```suggestion
Always specify a version or version range for package dependencies in project files, `packages.config` files, and `.nuspec` files. Without a version or version range, when resolving a dependency, consistent restore results are not guaranteed.
```

nkolev92 (approved):

nkolev92:
@aortiz-msft 

Can you please take another look. 

I've made pushed a change to the line above so we can complete this PR.

zivkan (approved):

nkolev92:
I've dismissed the stale review so I can merge this, but we can make edits on top of it @aortiz-msft thinks we need them.

PR #3360, opened 2024-11-18
Clarify what a package being available means

nkolev92:
See motivation: https://github.com/NuGet/Home/issues/13930#issuecomment-2484174618. 

tldr; If a package content resource lists a package as available, it should be able to download. 
Being unavailable breaks the contract.

### DIFF
--- a/docs/api/package-base-address-resource.md
+++ b/docs/api/package-base-address-resource.md
@@ -36,8 +36,8 @@ All URLs found in the package content resource support the HTTP methods `GET` an
 
 ## Enumerate package versions
 
-If the client knows a package ID and wants to discover which package versions the package source has available, the
-client can construct a predictable URL to enumerate all package versions. Every version listen in this list must be available for download.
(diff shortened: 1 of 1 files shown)


erdembayar (approved):

docs/api/package-base-address-resource.md
```diff
@@ -37,7 +37,8 @@ All URLs found in the package content resource support the HTTP methods `GET` an
 ## Enumerate package versions
 
 If the client knows a package ID and wants to discover which package versions the package source has available, the
-client can construct a predictable URL to enumerate all package versions. This list is meant to be a "directory
+client can construct a predictable URL to enumerate all package versions. Every version listen in this list must be available for download.
```
donnie-msft, line 40:
```suggestion
client can construct a predictable URL to enumerate all package versions. Every version listed in this list must be available for download.
```

donnie-msft (changes requested):

martinrrm (approved):

donnie-msft (approved):

zivkan (approved):

PR #3380, opened 2024-12-10
Update cli-ref-sources.md

fuixx217:
Adding some clarifying items when interacting with a custom/private Azure DevOps NuGet feed.

### DIFF
--- a/docs/reference/cli-reference/cli-ref-sources.md
+++ b/docs/reference/cli-reference/cli-ref-sources.md
@@ -58,7 +58,7 @@ where `<operation>` is one of *List, Add, Remove, Enable, Disable,* or *Update*,
   
   > [!NOTE]
   > When using an Azure DevOps Artifacts NuGet feed, this value should be set to a Personal Access Token (PAT) that is set with Read&Write under Packaging.
-  > For information about provisioning Azure DevOps Personal Access Tokens, please review [Azure DevOps - Use personal access tokens](azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate)
+  > For information about provisioning Azure DevOps Personal Access Tokens, please review [Azure DevOps - Use personal access tokens](/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate)
 
   > [!NOTE]
   > Be aware that encrypted passwords are only supported on Windows.
--- a/docs/reference/cli-reference/cli-ref-sources.md
+++ b/docs/reference/cli-reference/cli-ref-sources.md
@@ -58,7 +58,7 @@ where `<operation>` is one of *List, Add, Remove, Enable, Disable,* or *Update*,
   
   > [!NOTE]
   > When using an Azure DevOps Artifacts NuGet feed, this value should be set to a Personal Access Token (PAT) that is set with Read&Write under Packaging.
-  > For information about provisioning Azure DevOps Personal Access Tokens, please review [Azure DevOps - Use personal access tokens](https://learn.microsoft.com/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=Windows)
+  > For information about provisioning Azure DevOps Personal Access Tokens, please review [Azure DevOps - Use personal access tokens](../../organizations/accounts/use-personal-access-tokens-to-authenticate)
 
   > [!NOTE]
   > Be aware that encrypted passwords are only supported on Windows.
--- a/docs/reference/cli-reference/cli-ref-sources.md
+++ b/docs/reference/cli-reference/cli-ref-sources.md
@@ -55,6 +55,10 @@ where `<operation>` is one of *List, Add, Remove, Enable, Disable,* or *Update*,
 - **`-Password`**
 
   Specifies the password for authenticating with the source.
+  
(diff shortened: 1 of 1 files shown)


fuixx217:
@microsoft-github-policy-service agree

jeffkl:
Team Review: Rather than have our own wording for instructions, we'd prefer just a link to https://learn.microsoft.com/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=Windows so that the information stays up-to-date

fuixx217:
I just fought the nugget commands for about three weeks until I figured out putting the PAT in for the password is what I needed to do. I reviewed pretty much all of these links, so I suppose I can tweak this PR to only have the nugget commands, and include a link to the PAT link you referenced. Would that be acceptable?

fuixx217:
@jeffkl - can there be any movement on this PR, please?

fuixx217:
@OliaG @donnie-msft @zivkan I noticed you on other PR's that were merged, so I thought tagging you three might get some movement on this PR.  TIA! 😄

zivkan:
My feedback is that while Azure Artifacts is an important implementation of a NuGet feed, and probably the most used server implementation by .NET projects, I personally don't like it being explicitly being called out. .NET has a policy to not give the impression that you need to use Azure to host .NET apps, and I feel like NuGet should not give the impression that you should use Azure DevOps over other NuGet servers.

If we do call out Azure Artifacts specifically, why not add GitHub Package Repository as well? In which case, what about GitLab's package repository? Artifactory? Myget? Or the other servers listed in our "hosting" docs page

I think it would better instead to link to https://learn.microsoft.com/nuget/consume-packages/consuming-packages-authenticated-feeds where we already have this snippet:

> The credentials you need to use are determined by the package source. Therefore, unless you're using a credential provider, you should check with your package source for what credentials to use. It is very common for package sources to forbid you from using your password (that you log into the website with) with NuGet. Typically you need to create a Personal Access Token to use as NuGet's password, but you should check the documentation for the NuGet server you're using. Some package sources, such as Azure DevOps and GitHub, have scoped access tokens, so you may need to ensure that any tokens you create include the required scope.

fuixx217:
@zivkan 
Appreciate the comment back. I suppose I don't care where the example goes, but I had a heck of a time getting this working, and this page has many nuget.exe examples. If one of those examples called out interaction with AzureDevOps, that would have saved me probably a month's worth of time. The page you referenced doesn't have any examples, and I don't think that would have gotten me anywhere near where I needed to ultimately be. I have to imagine there are others out there that could benefit from the Azure DevOps example and the call out. To your point, however, why wouldn't the cli reference page also be a good place if others want to contribute examples for GitLabs, Gihub Package Repository, etc...?  Seems like the examples list there is what could benefit the community. Would only including the example be acceptable an omit the notes that I put under the Password and username properties?  I searched high and low for how to get this working, until I finally figured it out on my own, and I'm simply trying to save others the headaches I went through.... Whether its on an Azure Devops page or the one you listed, it was really unclear to use a personal access token when the prompt is Password. It's also unclear what the apiKey parameter provides, if anything, when used with ADO.

zivkan:
I messaged a .NET PM earlier today, and the feedback I got is that it's fine to use Azure DevOps as an example, but it should be clear that this is a common requirement across many services. Don't make it sound in any way specific to Azure DevOps. So, something along the lines that "Many NuGet feeds, like Azure Artifacts, do not allow you to provide your password here. You need to create a token with the relevant scopes, and provide that token as the password to NuGet. See your feed's documentation for more details".

However, I just remembered that Azure Artifacts (DevOps), we recommend using [their credential provider](https://github.com/microsoft/artifacts-credprovider).  So honestly, I'm not sure that an example of using a PAT with Azure Artifacts is a good example, because it's better to use the cred provider and avoid needing to manage the secret yourself. Although, if you're using a CI pipeline that is not Azure Pipelines, but are using Azure Artifacts as your feed, then you will need to manage PATs. If you're using Azure Pipelines, you should use the `NuGetAuthenticate` task instead.

I see on [Azure Artifact's docs for using it with NuGet](https://learn.microsoft.com/en-us/azure/devops/artifacts/get-started-nuget?view=azure-devops&tabs=windows), several pages list the credential provider as a prerequisite, and the [project setup for dotnet](https://learn.microsoft.com/en-us/azure/devops/artifacts/nuget/dotnet-setup?view=azure-devops&tabs=Windows) has a screenshot of the web UI, which has in bold text instructing you to ensure you have the latest version of the credential provider installed.  But a quick search doesn't find any mention of PATs or access tokens. I'll pass your feedback onto them.

Feel free to also propose any improvements to the "using authenticated feeds" doc. There might not be examples, but if you were blocked for a month, I hope at some point you would have chosen to read the full docs, rather than searching for specific keywords. It did mention Azure DevOps, so that should have been searchable, but maybe we need to add Azure Artifacts as well, since different people use different names.

fuixx217:
I edited this to only have the example that I was suggesting. Thoughts?

fuixx217:
@zivkan Nudging my last edit. TIA.

zivkan (approved):

nkolev92 (approved):

PR #3381, opened 2024-12-17
Adding No README chapter

OliaG:
Adding a question/answer for the case when the package has no README in Visual Studio Package Manager.

### DIFF
--- a/docs/resources/NuGet-FAQ.yml
+++ b/docs/resources/NuGet-FAQ.yml
@@ -185,6 +185,6 @@ sections:
         answer: |
           If you see the message "There is no README available for the selected package version" in Visual Studio NuGet Package Manager, that means that the selected version of the NuGet package does not have an embedded README file.
           - If this is not the latest version of the package, consider checking the latest version.
-          - If the latest version doesn't have the README file embedded either, and you are not the maintainer of this package, you can ask the maintainer to add the README by filing an issue in their repo. To do so, click on the `nuget.org` link in the top right corner of the Package Manager in Visual Studio near the package name.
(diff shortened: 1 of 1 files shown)


docs/resources/NuGet-FAQ.yml
```diff
@@ -179,3 +179,12 @@ sections:
           If the `target` has been modified manually, NuGet might not be able to detect that it needs removed when migrating.
           
           If your project is `PackageReference` and you still have this target in project file, it should be safe to remove.
+
+      - question: |
+          There is no README in Visual Studio NuGet Package Manager for the selected package.
+        answer: |
+          If you see the message "There is no README available for the selected package version" in Visual Studio NuGet Package Manager, that means that the selected version of the NuGet package does not have an embedded README file.
+          - If this is not the latest version of the package, consider checking the latest version.
+          - If the latest version doesn't have the README file embedded either, and you are not the maintainer of this package, you can ask the maintainer to add the README by filing an issue in their repo. To do so, click on the `nuget.org` link in the top right corner of the Package Manager in Visual Studio near the package name.
```
zivkan, line 188: From memory, out telemetry shows that more than 50% of restores use a package source other than nuget.org (possibly in addition to nuget.org).

If you'd still like to include nuget.org instructions, I think you should rephrase it to something like "If you use nuget.org as a package source you can....".    But the instructions are missing what to do after you click the nuget.org link.

OliaG: Good point, updated.

jgonz120 (approved):

zivkan (approved):

jeffkl (approved):

PR #3382, opened 2025-01-07
Add doc for ReadmeUriTemplate

jgonz120:


docs/api/readme-template-resource.md
```diff
@@ -0,0 +1,62 @@
...
+ms.author: jongonza
+ms.date: 1/6/2025
+ms.topic: reference
+ms.reviewer: 
+---
+
+# README Uri Template
+
+It is possible for a client to build a URL that can be used to download a README for a specific package.
+This will enable the clients to render the package's README without downloading the entire package.
```
zivkan, line 14: isn't it also possible for the readme url to be provided in the package metadata (registration) blob?  Or is that not yet implemented?

I think it should be made clear to server implementors that they implement one or the other, not both.

docs/api/readme-template-resource.md
```diff
@@ -0,0 +1,62 @@
...
+The resource used for building this URL is the `ReadmeUriTemplate` resource found in the
+[service index](service-index.md).
+
+## Versioning
+
+The following `@type` values are used:
+
+@type value                       | Notes
+--------------------------------- | -----
+ReadmeUriTemplate/6.13.0          | The initial release
```
zivkan, line 25: There's a "server implementor guide" doc that has a chronological list of changes to the NuGet protocol, so it's easy for server implementors to check if they're "up to date" with all client features, and find what's new since the last time they checked.  This should be documented on that page.

If you have any suggestions on how to make sure other people who make changes to the protocol know to update that page, without needing a human like me to remind them, please share your ideas.

docs/api/readme-template-resource.md
jgonz120: Not implemented yet, we don't have any commitments from servers to implement that option.

docs/api/readme-template-resource.md
jgonz120: When do we decide to update that page? For instance, the report abuse resource wasn't added there.

zivkan: please add the report abuse resource as well

docs/api/readme-template-resource.md
zivkan: well, this makes me frustrated.  Firstly, it imples that we talked to the dozens of 3rd party NuGet server implementors. Secondly, it means that [the `ReadmeFileUrl` public APIs in NuGet.Protocol](https://github.com/NuGet/NuGet.Client/blob/bf603fb32f32478a1ce0d6a1e27b7b64d14e244a/src/NuGet.Core/NuGet.Protocol/PublicAPI/net472/PublicAPI.Unshipped.txt#L13-L19) are now leaky abstractions, which some people consider an anti-pattern and tech debt.

Considering these public APIs are in the unshipped txt, are you able to delete these public APIs before this version of Nuget ships?

jgonz120: That's true I should say we didn't get a commitment from ADO to enable this feature. I unfortunately didn't see this response before the cutoff. That said the way I implemented this would have needed to change in a way that fixing it by Wednesday would have been too risky. 

Regarding a leaky abstraction the ReadmeFileUrl is being populated by the protocol using the ReadmeUriTemplate. I did this intentionally so the client wouldn't need to concern itself with what implementation the feed has implemented.

docs/api/readme-template-resource.md
jgonz120: Hmm, so I recall having a conversation about how we want to tell the servers that if they implement this new resource type that the client will be call it anytime we try to view a readme so they should expect calls for packages that do not have README and should return 404s... 

[NuGet Server Implementation Guide](https://learn.microsoft.com/en-us/nuget/api/implementation-guide#embedded-files) seems to have these sorts of considerations. Should we continue splitting these across the two documents? I don't feel like I know enough about how report abuse works to add anything meaningful to the guide.

zivkan: the two documents have different focuses. The resource doc has all the details. The server implementation guide helps people working on NuGet feeds figure out what has changed recently.

Pretend you work on a package repository, and notice that most of them (Azure Artifacts, Artifactory, GitHub Package Registry, etc) support multiple package managers (nuget, npm, cargo, pip, etc).  The server implementation guide helps these people quickly figure out if there are any new features that they might be interested in implementing, without having to read every single resource docs page.

Hence the server implementation guide is best to have a summary or overview, with a link to the detailed resource docs. But the important part is that it's a kind of change log with dates, so the server implementors can quickly check oh, we implemented everything (we care about) in 2020, is there anything new we should implement now?

jgonz120: Added the new section

docs/api/readme-template-resource.md
```diff
@@ -0,0 +1,62 @@
...
+
+Given a known package ID and version, the client implementation can construct a URL to download the README. 
+
+The value of the `@id` is a URL string containing any of the following placeholder tokens:
+
+### URL placeholders
+
+Name        | Type    | Required | Notes
+----------- | ------- | -------- | -----
+`{lower_id}`      | string  | no       | The package ID, lowercased
```
nkolev92, line 46: What does required imply?

jgonz120: Hmm, there are many examples of it being set to "No" for other resources...
https://learn.microsoft.com/en-us/nuget/api/report-abuse-resource#url-placeholders
https://learn.microsoft.com/en-us/nuget/api/package-details-template-resource#url-placeholders


I think it makes more sense to say required "Yes", I don't know how a server could implement the ability to download a README for a package+version without being told which package and version to download.

nkolev92 (approved):

PR #3388, opened 2025-01-31
Transitive pinning and pack

nkolev92:
Fixes https://github.com/NuGet/Home/issues/13811

Document the side-effects of transitively pinned packages on the pack command.


jeffkl (approved):

nkolev92:
Just a bug. 
I had different versions in an earlier variation.
Fixing in https://github.com/NuGet/docs.microsoft.com-nuget/pull/3399

PR #3393, opened 2025-02-05
Document package pruning

nkolev92:
Fixes https://github.com/NuGet/Home/issues/14002


docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -458,3 +458,51 @@ You can leave off `$(AssetTargetFallback)` if you wish to overwrite, instead of
...
+
+## PrunePackageReference
+
+The .NET Runtime is constantly evolving, with performance improvements and new APIs each release.
+There is a lot of functionality that's available within the runtime, but also as packages, such as [System..Text.Json](https://www.nuget.org/packages/System.Text.Json). This can often lead to a `System.Text.Json 8.0.0` in a project targeting `.NET 9` or `.NET 8`. This dependency is unnecessary and the build conflict resolution would not use the assembly coming from the package since it's already available in the .NET Runtime.
+Starting in in [NuGet version 6.13](..\release-notes\NuGet-6.13.md) and .NET SDK 9.0.200, `PrunePackageReference` enables the pruning of these packages at restore time for .NET SDK based projects.
+
+Package pruning is available as an opt-in feature with the .NET 9 SDK, and will be enabled by default for all `.NET` frameworks and `>= .NET Standard 2.0` starting with .NET 10 SDK.
+
+Package pruning is only available with the default dependency resolver as released in 6.12.
```
jeffkl, line 470: I doubt anyone would know what this means, maybe a link would help?

```suggestion
Package pruning is only available with the default dependency resolver as released in 6.12.  [See here](https://learn.microsoft.com/nuget/consume-packages/package-references-in-project-files#nuget-dependency-resolver) for more details.
```

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -458,3 +458,51 @@ You can leave off `$(AssetTargetFallback)` if you wish to overwrite, instead of
...
+
+| Attributes | Description |
+|------------|-------------|
+| Version | Specifies the maximum version to be pruned. `1.0.0` means that all packages up to and including 1.0.0 will be pruned. For `1.0.0`, `0.9.0` and `1.0.0` will be pruned, but `1.0.1` would not. |
+
+The following properties can be used to modify the pruning behavior.
+
+| PropertyName | Description |
+|--------------|-------------|
+| RestoreEnablePackagePruning  | Enables package pruning for the packages specified with `PrunePackageReference`. This property is per framework and the valid values are `true` and `false`. Defaults may differ based on the .NET SDK as defined above. |
```
jeffkl, line 484:
```suggestion
| RestoreEnablePackagePruning  | Enables package pruning for the packages specified with `PrunePackageReference`. This property is per target framework and the valid values are `true` and `false`. Defaults may differ based on the .NET SDK as defined above. |
```

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -458,3 +458,51 @@ You can leave off `$(AssetTargetFallback)` if you wish to overwrite, instead of
...
+
+The .NET SDK predefines the list of packages to be pruned for you.
+
+### How PrunePackageReference works
+
+When the restore algorithm encounters a package specified to be pruned, it removes the package from the graph. This package id is not downloaded and does not appear in the assets file libraries or targets section. When a package is pruned, there is a detailed verbosity message indicating that the package id has been removed for the given framework.
```
jeffkl, line 490: This felt too technical to me, so I tried making it more like documentation rather than a comment in our code, just a suggestion.

```suggestion
When a package is specified to be pruned during restore, it is removed from the dependency graph. This package is not downloaded and does not appear in any of the outputs of NuGet. When a package is pruned, there is a detailed verbosity message indicating that the package has been removed for the given target framework.
```

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -458,3 +458,51 @@ You can leave off `$(AssetTargetFallback)` if you wish to overwrite, instead of
...
+
+Pruning is only supported for transitive packages. The following table illustrates various pruning behaviors.
```
jeffkl, line 492:
```suggestion
**NOTE:** Pruning is only supported for transitive packages, meaning packages that are referenced by other packages or projects. The following table illustrates various package pruning behaviors.
```

nkolev92: This is part of the paragraph, so we don't usually add the notes in this case. 

The rest looks great, thanks for the suggestion!

docs/reference/errors-and-warnings/NU1510.md
```diff
@@ -0,0 +1,25 @@
...
+ms.topic: reference
+f1_keywords: 
+  - "NU1510"
+---
+
+# NuGet Warning NU1510
+
+## Scenario 1
+
+> PackageReference System.Text.Json 9.0.0 will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.
```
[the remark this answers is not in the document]
nkolev92: This is verbatim from the error message. 

I can look into updating that, but I'll leave it as is for now.

nkolev92:
Great feedback @jeffkl! Should be all addressed now.
1 comment is still open so please take a look at that one as well.

jeffkl (approved):
📦

PR #3394, opened 2025-02-11
Document how to enforce audit during restore

zivkan:
fixes: https://github.com/NuGet/Home/issues/13802

### DIFF
--- a/docs/concepts/Auditing-Packages.md
+++ b/docs/concepts/Auditing-Packages.md
@@ -4,6 +4,7 @@ description: How to audit package dependencies for security vulnerabilities and 
 author: JonDouglas
 ms.author: jodou
 ms.topic: conceptual
+ms.date: 02/11/2015
 ---
 
 # Auditing package dependencies for security vulnerabilities
--- a/docs/concepts/Auditing-Packages.md
+++ b/docs/concepts/Auditing-Packages.md
@@ -4,7 +4,7 @@ description: How to audit package dependencies for security vulnerabilities and 
 author: JonDouglas
 ms.author: jodou
 ms.topic: conceptual
-ms.date: 02/11/2015
+ms.date: 02/11/2025
 ---
 
 # Auditing package dependencies for security vulnerabilities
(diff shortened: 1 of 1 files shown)


docs/concepts/Auditing-Packages.md
```diff
@@ -105,6 +106,34 @@ Alternatively, if you want to keep low and moderate vulnerabilities as warnings,
 > [!NOTE]
 > MSBuild properties for message severity such as `NoWarn` and `TreatWarningsAsErrors` are not supported for packages.config projects.
 
+## Ensure restore audited projects
+
+NuGet in MSBuild 17.13 and .NET 9.0.200 added output properties `RestoreProjectCount`, `RestoreSkippedCount` and `RestoreProjectsAuditedCount` on the restore task.
+This can be used to enforce that audit ran during a restore.
+Note that these output properties are not available with static graph restore.
```
jeffkl, line 113: Let's add a link to static graph-based restore so users can check if they're using it or not.

https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#restoring-with-msbuild-static-graph-evaluation

jeffkl (approved):
I feel like we should just ship an implementation of this that users can set a property to enable.

docs/concepts/Auditing-Packages.md
```diff
@@ -105,6 +106,34 @@ Alternatively, if you want to keep low and moderate vulnerabilities as warnings,
...
+Note that *Directory.Build.props* is commonly used, but is imported by projects.
+However, NuGet's restore target and task runs at the solution level, so needs to be in MSBuild's solution extensibility file, not the project/build file.
+
+```xml
+<Project>
+    <Target Name="AssertRestoreTaskOutputProperties"
+            AfterTargets="Restore"
+            Condition="'$(CI)' == 'true'">
+        <Error
+            Condition="'$(RestoreProjectsAuditedCount)' != '$(RestoreProjectCount)'"
```
nkolev92, line 126: This should've been more of an implementation question, but do these numbers account for pc projects?

Basically, what about PC projects when it gets restored and pc projects when they're not restored because the property isn't passed.

PR #3402, opened 2025-02-19
Remove most Xamarin references

davidbritch:
Xamarin products have been out of support since 1st May 2024, and the [docs](https://learn.microsoft.com/previous-versions/xamarin) are archived. Hence I've removed most Xamarin references.


docs/reference/target-frameworks.md
```diff
@@ -288,22 +288,5 @@ Additional frameworks defined by third parties provide compatibility with other
  | | WindowsPhone 8.1 (UWP) |
  | | WindowsPhone 8.0 (SL) |
 
-Additionally, NuGet packages targeting Xamarin can use additional Xamarin-defined frameworks. See [Creating NuGet packages for Xamarin](https://developer.xamarin.com/guides/cross-platform/advanced/nuget/).
```
nkolev92, line 291: I think this was meant to be an exhaustive list of frameworks and we'd keep it.

For example, the PCL stuff is still in this doc.

Maybe for the time being we can just mark it with a note the same way PCLs are?

Then in a different change we can rethink this doc altogether.

davidbritch: I've restored the table.

nkolev92 (approved):

zivkan (approved):

davidbritch:
#sign-off

PR #3406, opened 2025-03-04
Live publish

Mikejo5000:


docs/consume-packages/Finding-and-Choosing-Packages.md
```diff
@@ -3,7 +3,7 @@ title: Find and evaluate NuGet packages
 description: Find and evaluate publicly available NuGet packages for your project by using advanced nuget.org search filters and syntax.
 author: JonDouglas
 ms.author: jodou
-ms.date: 02/08/2024
+ms.date: 03/03/2025
```
donnie-msft, line 6: What is the purpose of updating so many files' timestamps, without changing the content of those files?

donnie-msft (changes requested):

docs/consume-packages/Finding-and-Choosing-Packages.md
Mikejo5000: @donnie-msft The quickstarts and higher pageview tutorials were re-tested and updated within the last 18-20 months or so. The doc freshness scores, reported to the product team, are calculated using ms.date. and an article is considered "stale" after 12 months. Unless there's something that changed in the product that would make these articles out of date (I'm not aware of anything), it's OK to rev the date without re-testing. For the VS articles, we would typically only re-test after a new version of VS (Dev18). Please let me know if there's anything that changed in the product that would make these articles out of date.

zivkan: @Mikejo5000 sorry for the off topic question, but can we get the docs compiler to understand ISO 8601 yyyy-MM-dd  style formats, rather than just US style date formats?   Who should I talk to?

As an Australian living in Australia, I've created and updated docs in the past using dd/MM/yyyy format that is common throughout the rest of the world, and only when the day of the month is >= 13 I get a docs compiler warning. I'm sure I've published docs with the wrong date because it didn't occur to me to use US date format, and people reviewing my PR didn't pick it up (I've gotten approvals on my PRs even when the docs bot told me the date format was wrong)

zivkan (approved):

Nigusu-Allehu (approved):

docs/consume-packages/Finding-and-Choosing-Packages.md
Mikejo5000: @zivkan Please check with Megan Bradley

PR #3390, opened 2025-02-04
Add NuGet 6.13 Release Notes

Nigusu-Allehu:
Documentations that need to be completed for 6.13

https://github.com/NuGet/Home/milestone/151

PR #3412, opened 2025-03-20
add .NET 9.0.201 to 6.13.1

Nigusu-Allehu:
Now 9.201 has shipped, address https://github.com/NuGet/docs.microsoft.com-nuget/pull/3390/files#r1950143761
In addition update the versions table to have specific .net sdk versions for all the 6.13 releases


nkolev92 (dismissed):

docs/release-notes/NuGet-6.13.md
```diff
@@ -25,7 +25,7 @@ NuGet distribution vehicles:
 
 ## Summary: What's New in 6.13.1
 
-NuGet 6.13.1 is available in Visual Studio 17.13.
+NuGet 6.13.1 is available in Visual Studio 17.13 and .NET 9.0.201 SDK.
```
zivkan, line 28: there's a table at the top of the file that contains the exact VS builds that contain the different versions of NuGet, but lists 9.0.2xx in the .NET SDK column.   Should this version be listed there?

nkolev92: It's already listed actually

We probably don't need this PR at all. 

Tbh, we probably don't even need the sentence in the first place. 

Not sure if our docs template includes it.

zivkan: the table says 9.0.2xx, not 9.0.201 which this PR is adding.  Should we be telling customers the exact version?   If not, why do we for VS?

Nigusu-Allehu: I think we should. The table looks a bit uninformative when it comes to the .NET SDK column right now.
image
If a user wants to get the correct .NET SDK that has a fix or a feature, it would be difficult to tell from this table

nkolev92: I missed the part that it doesn't say the exact version.

We can add the exact version of the .NET SDK in the table

jeffkl (dismissed):

docs/release-notes/NuGet-6.13.md
nkolev92: Can we remove this sentence altogether?
6.13.2 doesn't have it.

That way people know to rely on the table

docs/release-notes/NuGet-6.13.md
```diff
@@ -14,7 +14,7 @@ NuGet distribution vehicles:
 | NuGet version | Available in Visual Studio version | Available in .NET SDK(s) |
 |:---|:---|:---|
 | [**6.13**](https://nuget.org/downloads) | [Visual Studio 2022 version 17.13](https://visualstudio.microsoft.com/downloads/) | [9.0.2xx](https://dotnet.microsoft.com/download/dotnet/9.0)<sup>1</sup> |
```
nkolev92, line 16: Fix this too, specify 9.0.200

docs/release-notes/NuGet-6.13.md
```diff
@@ -14,7 +14,7 @@ NuGet distribution vehicles:
...
-| [**6.13.1**](https://nuget.org/downloads) | [Visual Studio 2022 version 17.13.1](https://visualstudio.microsoft.com/downloads/) | [9.0.2xx](https://dotnet.microsoft.com/download/dotnet/9.0)<sup>1</sup> |
+| [**6.13.1**](https://nuget.org/downloads) | [Visual Studio 2022 version 17.13.1](https://visualstudio.microsoft.com/downloads/) | [9.0.201](https://dotnet.microsoft.com/download/dotnet/9.0)<sup>1</sup> |
 | [**6.13.2**](https://nuget.org/downloads) | [Visual Studio 2022 version 17.13.2](https://visualstudio.microsoft.com/downloads/) | [9.0.2xx](https://dotnet.microsoft.com/download/dotnet/9.0)<sup>1</sup> |
```
nkolev92, line 18: This will be 9.0.203 I think.

cc @jeffkl

docs/release-notes/NuGet-6.13.md
Nigusu-Allehu: Note:
previously for the other releases we left the table at the top to specify the main release only. We did not add multiple rows for each patch. Is this something we want to start doing? 
image

nkolev92: I think we've just been doing it consistently.  

Take 6.8 for example:

https://learn.microsoft.com/en-us/nuget/release-notes/nuget-6.8

image

jeffkl (approved):

PR #3431, opened 2025-05-15
Fix typo

bsivanov:
Downgrades are actually *not* allowed when transitive pinning a package


nkolev92:
@bsivanov this was good change, do you want to reopen this PR?

nkolev92 (approved):

PR #3432, opened 2025-05-19
In `<PackageReference Condition/>` docs, mention `IsTargetFrameworkCompatible`

YoshiRulz:
examples based on dotnet/docs#45462

### DIFF
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -170,6 +170,20 @@ Conditions can also be applied at the `ItemGroup` level and will apply to all ch
 </ItemGroup>
 ```
 
+When you have many targets, it may be better to match ranges of TFMs, in which case you can use `IsTargetFrameworkCompatible`:
+```xml
+<ItemGroup>
+    <!-- reference 8.0 System.Text.Json when targeting things older than .NET 8 -->
+    <PackageReference Include="System.Text.Json"  Version="8.0.5" Condition=" !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0')) " />
+
+    <!-- reference 10.0 System.Linq.AsyncEnumerable when targeting things older than .NET 10 -->
+    <PackageReference Include="System.Linq.AsyncEnumerable"  Version="10.0.0-preview.2.25163.2" Condition=" !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net10.0')) " />
+
+    <!-- Reference System.Memory on frameworks not compatible with .NET Core 2.1 nor .NETStandard 2.1 -->
+    <PackageReference Include="System.Memory" Condition=" !($([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'netcoreapp2.1')) OR $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'netstandard2.1'))) " />
+</ItemGroup>
+```
+
 ## GeneratePathProperty
 
 This feature is available with NuGet **5.0** or above and with Visual Studio 2019 **16.0** or above.
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -170,7 +170,7 @@ Conditions can also be applied at the `ItemGroup` level and will apply to all ch
 </ItemGroup>
 ```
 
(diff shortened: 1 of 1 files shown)


docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -170,6 +170,20 @@ Conditions can also be applied at the `ItemGroup` level and will apply to all ch
 </ItemGroup>
 ```
 
+When you have many targets, it may be better to match ranges of TFMs, in which case you can use `IsTargetFrameworkCompatible`:
```
nkolev92, line 173: Can we link to https://learn.microsoft.com/en-us/visualstudio/msbuild/property-functions?view=vs-2022#msbuild-targetframework-and-targetplatform-functions as well?
https://github.com/NuGet/docs.microsoft.com-nuget?tab=readme-ov-file#links

It's worth noting that the more complicated the conditions are, the more challenging it is for any tooling to update these correctly. 
I'd consider these to be advanced scenarios, so my suggestion would be to call them out as such.

YoshiRulz:
I've just noticed the third example doesn't have `Version`. Should I add that, or assume the reader is using CPM and omit all of them?

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -170,6 +170,20 @@ Conditions can also be applied at the `ItemGroup` level and will apply to all ch
 </ItemGroup>
 ```
 
+When you have many targets, it may be better to match ranges of TFMs, in which case you can use [`IsTargetFrameworkCompatible`](/visualstudio/msbuild/property-functions#msbuild-targetframework-and-targetplatform-functions):
+```xml
+<ItemGroup>
```
nkolev92, line 175: The 3 examples are pretty equivalent. 

They're all showing the usage of IsTargetFrameworkCompatible with negation. 
I think it'll be easier for the reader if they were to see one example.

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -170,6 +170,20 @@ Conditions can also be applied at the `ItemGroup` level and will apply to all ch
...
+    <!-- reference 8.0 System.Text.Json when targeting things older than .NET 8 -->
+    <PackageReference Include="System.Text.Json"  Version="8.0.5" Condition=" !$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0')) " />
```
nkolev92, line 177: This is probably me overthinking it because of https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#prunepackagereference, but the idea is that we generally want to make sure people are not referencing platform packages unless they're absolutely required. 
For example, there's many System.Text.Json references when customers target .NET and that's unnecessary. 

So I'm thinking that we can maybe use a Contoso named package, or  if we're sticking with System.Text.Json we can use: GetTargetFrameworkIdentifier. 

Maybe we end up with 2 examples, 1 using `IsTargetFrameworkCompatible`, while the other one uses `GetTargetFrameworkIdentifier`

nkolev92 (commented):
Thanks for this contribution. 

Sorry for not doing as thorough of a review the first go.

docs/consume-packages/Package-References-in-Project-Files.md
YoshiRulz: As in `$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) != '.NETCoreApp'`?

nkolev92: Yes.

zivkan: Since the .NET SDK sets property `TargetFrameworkIdentifier`, and MSBuild evaluates properties top to bottom before items, doing a TFI condition is easier by using `Condition=" '$(TargetFrameworkIdentifier)' != '.NETCoreApp' "`, rather than calling the instrinsic function and passing the target framework as an argument.

nkolev92:
Closing this since we haven't progress on it in a while. 

We can always reopen it if needed.

YoshiRulz:
Rebased, and I just made a best guess at what direction you wanted to move in.

docs/consume-packages/Package-References-in-Project-Files.md
YoshiRulz: Done, but I can't leave the top of this section unchanged since it's now self-contradictory: "[only the `TargetFramework` variable is supported](https://github.com/YoshiRulz/docs.microsoft.com-nuget/blob/5e37700e67e6c199c3daa814a7cd942895a2a50e/docs/consume-packages/Package-References-in-Project-Files.md?plain=1#L146)"

PR #3433, opened 2025-05-22
Remove link from NU1302 docs that links to itself

donnie-msft:


docs/reference/errors-and-warnings/NU1302.md
```diff
@@ -10,7 +10,7 @@ f1_keywords:
 
 # NuGet Error NU1302
 
-> You are running the 'restore' operation with an 'HTTP' source: myHttpSource. NuGet requires HTTPS sources. To use an HTTP source, you must explicitly set 'allowInsecureConnections' to true in your NuGet.Config file. Please refer to https://aka.ms/nuget-https-everywhere for more information.
```
nkolev92, line 13: Should https://aka.ms/nuget-https-everywhere have pointed somewhere else?

cc @Nigusu-Allehu? 

Cause if it's pointing to this page directly, then I'd argue we should remove this text from the error message altogether (meaning change the product too)

donnie-msft: > Should https://aka.ms/nuget-https-everywhere have pointed somewhere else?

I believe a NU1302 linking to the docs for NU1302 is reasonable to do, but I see no reason to have the link on the docs page point back to itself.

nkolev92: The whole point about the error codes is that they're well known, something you can look up with your favorite search engine and you'd get to where you need to go. 

CSC warnings don't link to the page. 

I find it to be a funny precedent to "code" something and add links with the error messages.
It just clutters things, or in the case of the docs, leads to a funny loop.

donnie-msft: Absolutely, we should be using error codes. In the meantime, this PR removes a cyclical link from the docs side. In no case does that make sense to me (whether or not the source is changed to properly use error code links).

nkolev92: That's why my first question was whether this was supposed to point to somewhere else.
Because if that's the case, then we shouldn't take this PR and update the link instead.

zivkan: Put another way, all of our warning/error code doc pages have a copy-paste of the message that the product outputs, to make it easy for customers to copy/paste to find (in addition to the NU code) via a search engine.

The line being modified is the copy-paste of the product's output message. Removing it makes it no longer a copy-paste, and therefore no longer matches the "pattern" that all other NU code docs follow.

If we have an aka.ms link in the output message to take the customer to the NU code docs, by definition the link on the docs page will take the browser back to the same page.  But the point isn't to show customers a link to another page, the point is that it's a copy-paste that search engines will index so it's easy for customers to find.

zivkan: Should we start putting a "sample output:" line above the quoted line, on every single NU docs page?

Nigusu-Allehu: > Should https://aka.ms/nuget-https-everywhere have pointed somewhere else?

The intention was to provide documentation on how to opt out of the error and make it easily accessible via an aka.ms link, as discussed in https://github.com/NuGet/NuGet.Client/pull/5693#discussion_r1525585944. I considered creating a separate page about using `allowInsecureConnections` to bypass HTTP errors, but realized this would largely duplicate the existing NU1302 documentation. Therefore, it seemed logical to direct the aka.ms link to the NU1302 page, which already contains comprehensive information about HTTP errors and their mitigation.

There are two points to consider regarding this approach:

1. **Cyclic Link:** As @donnie-msft pointed out, the aka.ms link appears in the documentation and redirects to the same page. This is intentional, as the doc is meant to show a sample of the error. It's not instructing users to visit the link for further information, so I don't see it as truly cyclic. If the doc said “learn more about this error at...” and pointed to itself, that would be confusing, but that's not the case here.
2. **Link vs. Searchability:** NU error codes are designed to be easily searchable, so including a link isn't strictly necessary. Personally, I find that providing a direct link helps users quickly find relevant information about new errors, rather than requiring them to search for the NU code. That said, I’m not opposed to removing the link for consistency with other NU errors, but I don't believe doing so would improve the user experience—just make it more uniform.

nkolev92: If the link is pointing to the same thing searching NU1302 would lead to, imo it doesn't add anything new so it should be removed. 

We already have a bunch of warnings/errors that don't render well in the error list or are wordy, so I generally lean towards things that don't have us spin our wheels with questioning our documentation approaches :) 

I'd be happy to submit a PR to https://github.com/nuget/docs.microsoft.com-nuget/?tab=readme-ov-file#conventions that explicitly tells us how we documentation our log codes (ie, the example should exactly match the text you'd get from the tool) if people agree with it.

PR #3438, opened 2025-06-05
Recommend against 3rd party servers using nuget.org resources directly

zivkan:


docs/api/vulnerability-info.md
```diff
@@ -33,6 +33,16 @@ The data file schema does not allow for modification or redaction of known vulne
 Therefore if a server's vulnerability data source (for example the [GitHub Advisories Database](https://github.com/advisories)) modifies an existing advisory, the NuGet server must modify the page that the vulnerability information was previously reported.
 One way to achieve this with the suggested partition scheme is to treat all vulnerability modifications and deletions as a trigger to regenerate the complete `base.json` file, and empty `updates.json`.
 
+## Reusing nuget.org's vulnerability data
```
zivkan, line 36: I can't decide whether this content should be in the server implementation guide document, rather than here.

Any comments?

joelverhagen:
I'm curious about the context here. Did something happen with a known 3P feed linking to NuGet.org? The guidance seems reasonable enough, I just wonder what the impetus is.

docs/api/service-index.md
```diff
@@ -65,6 +65,10 @@ There is no requirement that each resource has a unique `@id` or `@type`. It is
 determine which resource to prefer over another. One possible implementation is that resources of the same or
 compatible `@type` can be used in a round-robin fashion in case of connection failure or server error.
 
+A resource can use a different host or domain than the service index, but this may cause issues in environments with strict network rules.
+If you delegate resources to api.nuget.org, your feed will not work where access to nuget.org is blocked.
```
joelverhagen, line 69:
```suggestion
If you delegate resources to nuget.org, your feed will not work where access to nuget.org is blocked.
```

Technically our service index points to several other nuget.org domains (search, gallery, symbols). Maybe this can be generalized to this 👆? Not sure.

docs/api/service-index.md
```diff
@@ -65,6 +65,10 @@ There is no requirement that each resource has a unique `@id` or `@type`. It is
...
+We recommend configuration to enable or disable features that use api.nuget.org.
```
joelverhagen, line 70: I am not sure what this means. I think I can guess it, but maybe it should be expanded a bit with an example. It seems to me like the guidance allows proxying/caching nuget.org (since this would avoid network access controls) which could still be considered "using" api.nuget.org.

I might be missing the context here and that's why I am confused.

docs/api/vulnerability-info.md
```diff
@@ -33,6 +33,16 @@ The data file schema does not allow for modification or redaction of known vulne
...
+
+NuGet does not enforce that resources in the [service index](./service-index.md), or [the vulnerability index](#vulnerability-index), must be on the same server as the service index itself.
+However, there are several reasons why some companies choose to block api.nuget.org at the firewall.
+To avoid connectivity issues, we recommend serving vulnerability data from your own web app, so that NuGet clients only make HTTP connections to a single host.
```
joelverhagen, line 40:
```suggestion
To avoid connectivity issues, we recommend serving vulnerability data from your own web app, so that NuGet clients only make HTTP connections to a host you control.
```

docs/api/vulnerability-info.md
joelverhagen: I think it should be in the implementation guide as a specific example. To me it seems like the point generalizes to many potential upstreaming modes (not just for vuln data).

docs/api/vulnerability-info.md
```diff
@@ -33,6 +33,16 @@ The data file schema does not allow for modification or redaction of known vulne
 Therefore if a server's vulnerability data source (for example the [GitHub Advisories Database](https://github.com/advisories)) modifies an existing advisory, the NuGet server must modify the page that the vulnerability information was previously reported.
 One way to achieve this with the suggested partition scheme is to treat all vulnerability modifications and deletions as a trigger to regenerate the complete `base.json` file, and empty `updates.json`.
 
+## Reusing nuget.org's vulnerability data
+
+NuGet does not enforce that resources in the [service index](./service-index.md), or [the vulnerability index](#vulnerability-index), must be on the same server as the service index itself.
```
nkolev92, line 38: nit: I'd flip this. NuGet does not require....

My suggestion would also be to add some explicit information that every resource is expected to be available. 
I don't know where to add that but feels important.

zivkan: Is there anyone who would assume it's ok to use URLs that aren't available?  It seems weird to explicitly write "don't do this thing that is obviously wrong".

nkolev92: I was think more like available to the end user. 

Maybe this is more of a client thing. 

Basically, if you block a part of a feed, and it breaks, it's not a supported scenario.

docs/api/vulnerability-info.md
zivkan: In the case of a .NET developer using a self-hosted feed, it's the feed implementation that needs to implement vulnerability info, but they don't control the server that it's installed on.   I'll push a change soon with different wording, so I'd like your opinion once it's pushed.

docs/api/vulnerability-info.md
zivkan: I added something to the "host your own feed" page.  What do you think?

docs/api/service-index.md
zivkan: I mean putting api.nuget.org URLs in the service index, which causes clients to try to access api.nuget.org directly. proxying/caching is great, and what feeds should do, to avoid situations where the server is installed on-prem, and the company firewall blocks nuget.org.

I've updated the PR, let me know what you think now.

docs/hosting-packages/Overview.md
```diff
@@ -42,3 +42,7 @@ There are also several other NuGet hosting products such as [Azure Artifacts](ht
 - [RepoFlow](https://www.repoflow.io), a simple and easy-to-use package management platform.
 
 Regardless of how packages are hosted, you access them by adding them to the list of available sources in `NuGet.Config`. This can be done in Visual Studio as described in [Package Sources](../consume-packages/install-use-packages-visual-studio.md#package-sources), or from the command line using [`nuget sources`](../reference/cli-reference/cli-ref-sources.md). The path to a source can be a local folder pathname, a network name, or a URL.
+
+[NuGet's V3 protocol](../api/overview.md) uses a [service index](../api//service-index.md) that contains a list of URLs used by various NuGet operations.
+All of the URLs in the service index must be accessible by developer and CI machines to avoid unexpeced errors.
```
nkolev92, line 47:
```suggestion
All of the URLs in the service index must be accessible by developer and CI machines to avoid unexpected errors.
```

joelverhagen (approved):

joelverhagen (approved):

nkolev92 (approved):

PR #3444, opened 2025-07-17
Improve pruning documentation

nkolev92:
Fixes https://github.com/NuGet/Home/issues/14410
Fixes https://github.com/NuGet/Home/issues/14330
Fixes https://github.com/NuGet/Home/issues/14254

### DIFF
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -498,16 +498,16 @@ For transitive packages, meaning packages that are referenced by other packages 
 
 For direct packages, `PrivateAssets='all'` and `IncludeAssets='none'` are implicitly applied.
 
-- `IncludeAssets='none'` ensures that the assemblies from this package are used during the build.
+- `IncludeAssets='none'` ensures that the assemblies from this package are used during the build. Before pruning existing, the conflict resolution during build ensured the platform assemblies are preferred over the ones coming from the packages.
 - `PrivateAssets='all'` ensures that the packages aren't included in packages or through project references.
 
 Example:
 
-A project file like below:
+A project like below:
 
 ```xml
   <PropertyGroup>
-    <TargetFramework>net9.0;netstandard2.0</TargetFramework>
+    <TargetFrameworks>net9.0;netstandard2.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
@@ -527,7 +527,10 @@ will have a nuspec with the following dependencies:
 </dependencies>
 ```
 
-The following table illustrates various package pruning behaviors.
+When a direct PackageReference can be completely removed from your project, and one of the project frameworks are .NET 10 or newer, [NU1510](../reference/errors-and-warnings/NU1510.md) will be raised asking you to remove the package. 
+Following this suggestion will reduce the complexity of your project graph.
+
+The following table summarizes all the package pruning behaviors.
 
 | Dependency disposition | Behavior |
 |-----------------|----------|
@@ -535,10 +538,6 @@ The following table illustrates various package pruning behaviors.
 | Matches the ID of a transitive package coming through another project | Prune |
 | Matches the ID of a direct `PackageReference` | Apply `PrivateAssets='all'` and `IncludeAssets='none'` and raise the [NU1510](../reference/errors-and-warnings/NU1510.md) when the package can be removed from all frameworks and the project targets .NET 10. |
 | Matches the ID of a `ProjectReference` | Do not prune and raise the [NU1511](../reference/errors-and-warnings/NU1511.md) warning when the project targets .NET 10 |
-
-TODO NK: Clean-up, better documentation.
-
-Separate file?
 
 ### PrunePackageReference applications
 
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -464,10 +464,12 @@ You can leave off `$(AssetTargetFallback)` if you wish to overwrite, instead of 
 ## PrunePackageReference
 
 The .NET Runtime is constantly evolving, with performance improvements and new APIs each release.
-There is a lot of functionality that's available within the runtime, but also as packages, such as [System.Text.Json](https://www.nuget.org/packages/System.Text.Json). This can often lead to a `System.Text.Json 8.0.0` in a project targeting `.NET 9` or `.NET 8`. This dependency is unnecessary and the build conflict resolution would not use the assembly coming from the package since it's already available in the .NET Runtime.
-Starting in in [NuGet version 6.13](..\release-notes\NuGet-6.13.md) and .NET SDK 9.0.200, `PrunePackageReference` enables the pruning of these packages at restore time for .NET SDK based projects.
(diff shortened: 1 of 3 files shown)


docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -464,10 +464,12 @@ You can leave off `$(AssetTargetFallback)` if you wish to overwrite, instead of
 ## PrunePackageReference
 
 The .NET Runtime is constantly evolving, with performance improvements and new APIs each release.
-There is a lot of functionality that's available within the runtime, but also as packages, such as [System.Text.Json](https://www.nuget.org/packages/System.Text.Json). This can often lead to a `System.Text.Json 8.0.0` in a project targeting `.NET 9` or `.NET 8`. This dependency is unnecessary and the build conflict resolution would not use the assembly coming from the package since it's already available in the .NET Runtime.
-Starting in in [NuGet version 6.13](..\release-notes\NuGet-6.13.md) and .NET SDK 9.0.200, `PrunePackageReference` enables the pruning of these packages at restore time for .NET SDK based projects.
+There is a lot of functionality that's available within the runtime, but also as packages, such as [System.Text.Json](https://www.nuget.org/packages/System.Text.Json).
```
zivkan, line 467: very opinionated: It looks like my suggestion is not shorter than what's already written here, but to me "There is a lot of functionality that's available within the runtime" reads funny to me.

I'm thinking something along the lines:

```suggestion
New features added to .NET sometimes is also provided as packages, so that developers using older target frameworks can use the library, such as [System.Text.Json](https://www.nuget.org/packages/System.Text.Json).
```

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -489,16 +491,53 @@ The .NET SDK predefines the list of packages to be pruned for you.
 
 ### How PrunePackageReference works
 
-When a package is specified to be pruned during restore, it is removed from the dependency graph. This package is not downloaded and does not appear in any of the outputs of NuGet. When a package is pruned, there is a detailed verbosity message indicating that the package has been removed for the given target framework.
+When a package is specified to be pruned during restore, it is removed from the dependency graph.
+When a package is pruned, there is a detailed verbosity message indicating that the package has been removed for the given target framework.
```
zivkan, line 495: very opinionated: 

```suggestion
When a package is pruned, there is a message, visible at detailed verbosity, indicating that the package has been removed for the given target framework.
```

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -489,16 +491,53 @@ The .NET SDK predefines the list of packages to be pruned for you.
...
 
-Pruning is only supported for transitive packages, meaning packages that are referenced by other packages or projects. The following table illustrates various package pruning behaviors.
+For transitive packages, meaning packages that are referenced by other packages or projects, the packages are not downloaded and do not appear in any of the outputs of NuGet.
```
zivkan, line 497:
```suggestion
For transitive packages, meaning dependencies of other packages or projects, the packages are not downloaded and do not appear in any of the outputs of NuGet.
```

zivkan (approved):
just a few suggestions that make it a little easier for me to read, but maybe it's worse for other people 😁

docs/consume-packages/Package-References-in-Project-Files.md
nkolev92:
```suggestion
New features added to .NET sometimes are also provided as packages, so that developers using older target frameworks can use the library, such as [System.Text.Json](https://www.nuget.org/packages/System.Text.Json).
```

docs/reference/errors-and-warnings/NU1510.md
```diff
@@ -11,15 +11,74 @@ f1_keywords:
 
 # NuGet Warning NU1510
 
-## Scenario 1
+## Scenario
 
-> PackageReference System.Text.Json 9.0.0 will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.
+> PackageReference System.Text.Json will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.
```
kartheekp-ms, line 16:
```suggestion
> PackageReference System.Text.Json will be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.
```

I thought this warning is raised to notify customers that a package reference will be pruned by .NET SDK later during build. Hence, we are suggesting customers to remove the package reference. Please let me know if I am missing something.

Frulfump: I'm also interested in a clarification to make sure I understand it but I think from the docs

> The following table summarizes all the package pruning behaviors.
> 
> | Dependency disposition | Behavior |
> |-----------------|----------|
> | Matches the ID of a direct `PackageReference` | Apply `PrivateAssets='all'` and `IncludeAssets='none'` and raise the [NU1510](../reference/errors-and-warnings/NU1510.md) warning when the package can be removed from all frameworks and the project targets .NET 10. |

Means that it can be pruned and it is (to some degree pruned, but not "fully") by setting `PrivateAssets` and `IncludeAssets` to `none` and then gives you the warning to gain the benefits described below. 

>Performance benefits, by virtue of reducing the number of packages within a dependency graph
>Reduction of false positives by component scanners such as NuGetAudit

i.e we would still see the package in project.assets.json and that is why some poorly? configured scanners will still flag a vuln? (that's what I'm seeing with
>dotnet new console
>dotnet package add System.Memory@4.6.3
<check obj\project.assets.json>
)

Maybe the effect of setting those options to none could be explained I think that would clear it up a bit?

nkolev92: The pruning for directs is more limited and not full. 
The end goal is for a user to remove this package from their project.

I think the message should remain as is (it matches the actual warning message text)

docs/reference/errors-and-warnings/NU1510.md
```diff
@@ -11,15 +11,74 @@ f1_keywords:
...
+
 ### Solution
 
 Remove the PackageReference as it's unnecessary.
+
+> [!NOTE]
+> In order to allow for easier adoption of the PrunePackageReference feature, this warning is raised by default when a project targets the .NET 10 framework or newer.
+
+> [!NOTE]
+> The warning will not be raised in scenarios in which at least one of the frameworks still needs the package, such as `net48` in the below example.
```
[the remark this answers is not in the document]
nkolev92: We actually explicitly moved away from that because it can lead authors to a scenario where they build packages that could eventually cause runtime issues for their customers. 

Please read: https://github.com/NuGet/Home/blob/dev/accepted/2025/PrunePackageReference-with-direct-PackageReference.md#motivation and it's tracking issue: https://github.com/NuGet/Home/issues/14196

nkolev92:
> When doing dotnet package add System.Memory@4.6.3 I see the warning text but colored as info

That sounds like an add package issue, rather than something specific to the pruning warning. 
Can you please file an issue for that? (nuget/home repo)

> The dotnet package add command seem to trigger an implicit build

That's unusual, that shouldn't be the case. Similarly please file an issue about that (dotnet/sdk repo for this one)

Frulfump (commented):
Looks good to me, nice improvements/clarifications appreciated the examples and the updated table.

PR #3450, opened 2025-07-25
Add conceptual document of MCP and NuGet packages

joelverhagen:
This should help folks understand more about why we think NuGet and .NET are great for building MCP servers. I didn't want the document to come across as too push, or be too direct since some of the points are subjective and depend on team/author context. Also, until we have a `dnx` that is more easily acquired (either via https://github.com/dotnet/sdk/issues/49796 or a standalone `dnx`), the user will continue to need _some_ toolchain to do the package download and tool process launch.


baronfel (dismissed):
This is a solid take 👍 I suspect that node will be the most popular MCP server shipping format, but we can put up a heck of a fight on the raw performance axis. And once we get a native `dnx` then IMO node will have nothing on us!

docs/concepts/nuget-mcp.md
```diff
@@ -0,0 +1,91 @@
...
+ms.date: 07/23/2025
+---
+
+# MCP servers in NuGet packages
+
+NuGet provides a convenient way to package and distribute MCP servers written in .NET. The C# MCP SDK and .NET offer a robust platform for building MCP servers, and NuGet is ideal for delivering your MCP server to end users as a local tool. Self-contained, platform-specific packages reduce runtime compatibility issues, and AOT compilation can further improve the end-user experience.
+
+For more information about the Model Context Protocol (MCP) in general, see the [introduction on the MCP website](https://modelcontextprotocol.io/introduction). To create your own MCP server and package it using NuGet, see the [quickstart guide](/dotnet/ai/quickstarts/build-mcp-server).
+
+## Key considerations when building an MCP server
```
jeffkl, line 16: This section seems redundant to the "Applicable scenarios" section and that section I think does a better job of helping a user determine whether or not they should use a .NET-based MCP server via NuGet.  Consider folding this section into that one.

docs/concepts/nuget-mcp.md
```diff
@@ -0,0 +1,91 @@
...
+
+All of these tools support installing alternate sources. For example, `dnx` supports installing from Azure DevOps using the `--source` parameter, allowing consumption of private MCP servers, as long as the needed credential or credential providers are configured.
+
+## Runtime requirements
+
+Once the package is downloaded, a runtime is needed to execute the code inside the package. .NET tool packages (and by extension NuGet-based MCP servers) support a variety of options for how the tool is compiled and packaged. These options allow you, the MCP server author, to decide which runtime requirements should be placed on the users of your MCP server.
+
+There are three main options for how to package your MCP server:
+
+1. **Framework-dependent**: Requires that the MCP client has access to a compatible .NET runtime. If `dnx` is being used to download and execute the package, a runtime will be available.
```
jeffkl, line 74: I would include a link to the docs so a user can read more: https://learn.microsoft.com/en-us/dotnet/core/deploying/#platform-specific-and-framework-dependent

docs/concepts/nuget-mcp.md
```diff
@@ -0,0 +1,91 @@
...
+2. **Self-contained**: Bundles the runtime with the package. [Using trimming](/dotnet/core/deploying/trimming/trimming-options) can reduce the size of the package.
+3. **Ahead-of-time (AOT) compiled**: A self-contained package with AOT compilation enabled.
```
jeffkl, line 76:
```suggestion
3. **Ahead-of-time (AOT) compiled**: A self-contained package with AOT compilation enabled. More information is available at https://learn.microsoft.com/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8
```

PR #3451, opened 2025-07-28
Mention audit sources are supported starting 9.0.300 for package list…

Nigusu-Allehu:
… command

Fixes: https://github.com/NuGet/Home/issues/14371

### DIFF
--- a/docs/concepts/Auditing-Packages.md
+++ b/docs/concepts/Auditing-Packages.md
@@ -74,7 +74,8 @@ Note that the [V2 protocol is deprecated](../nuget-org/overview-nuget-org.md#api
 
 Audit sources are available from [NuGet 6.12, .NET 9.0.100 SDK, and Visual Studio 2022 17.12](../release-notes/NuGet-6.12.md).
 Prior to this version, NuGet Audit will only use package sources to download vulnerability information.
-Audit sources are not used by `dotnet list package --vulnerable` at this time.
+Starting [NuGet 6.14, .NET 9.0.300 SDK, and Visual Studio 2022 17.14](../release-notes/NuGet-6.14.md) the `dotnet package list --vulnerable` command also uses audit sources.
+In earlier versions, the command relied solely on package sources.
 
 #### Excluding advisories
 
--- a/docs/concepts/Auditing-Packages.md
+++ b/docs/concepts/Auditing-Packages.md
@@ -72,10 +72,13 @@ Note that the [V2 protocol is deprecated](../nuget-org/overview-nuget-org.md#api
 </configuration>
 ```
 
-Audit sources are available from [NuGet 6.12, .NET 9.0.100 SDK, and Visual Studio 2022 17.12](../release-notes/NuGet-6.12.md).
-Prior to this version, NuGet Audit will only use package sources to download vulnerability information.
-Starting [NuGet 6.14, .NET 9.0.300 SDK, and Visual Studio 2022 17.14](../release-notes/NuGet-6.14.md) the `dotnet package list --vulnerable` command also uses audit sources.
-In earlier versions, the command relied solely on package sources.
+**Note**: The table below lists features that support Audit Sources.
+
+| Introduced In                                                                                | Feature Supporting Audit Sources                                   |
+| -------------------------------------------------------------------------------------------- | ------------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


docs/concepts/Auditing-Packages.md
```diff
@@ -74,7 +74,8 @@ Note that the [V2 protocol is deprecated](../nuget-org/overview-nuget-org.md#api
 
 Audit sources are available from [NuGet 6.12, .NET 9.0.100 SDK, and Visual Studio 2022 17.12](../release-notes/NuGet-6.12.md).
 Prior to this version, NuGet Audit will only use package sources to download vulnerability information.
-Audit sources are not used by `dotnet list package --vulnerable` at this time.
+Starting [NuGet 6.14, .NET 9.0.300 SDK, and Visual Studio 2022 17.14](../release-notes/NuGet-6.14.md) the `dotnet package list --vulnerable` command also uses audit sources.
```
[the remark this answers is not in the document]
Nigusu-Allehu: Following this PR https://github.com/dotnet/sdk/pull/45384, the command order has changed.

nkolev92, line 77: I think this is good, but I wonder if having a table would be better. 

In particular, there are 3 scenarios I could think of where audit sources applies. 

1. restore (packages.config vs PackageReference, but ideally wer don't document that.
2. dotnet package list --vulnerable
3. PM UI - https://github.com/NuGet/Home/issues/13954

PM UI is scheduled in https://github.com/NuGet/Home/issues/13954. 

Thoughts @zivkan?

zivkan: I guess a table could be easier to understand.  For example

|Version|Feature|
|--|--|
|VS 17.8, NuGet 6.8, .NET 9.0.100|Restore & VS's Package Management installed tab for transitive packages
|.NET 9.0.300|`dotnet list package --vulnerable`|

It's less wordy than a paragraph of text.  But at the same time, right now there's only 2 relevant versions.

I could go either way, to be honest.

zivkan, line 77: the dotnet CLI is part of the .NET SDK.  I don't think that the VS or nuget.exe versions help customers understand when `dotnet list package --vulnerable` supports audit sources

nkolev92: > VS's Package Management installed tab for transitive packages

Is this actually supported though? I thought it wasn't?

I think a table for 2 probably makes less sense, but my thinking was a table helps cause we have 3, one of whic his not done yet.

zivkan: You're right. I'm getting mixed up. PM UI transitive packages have vulnerabilities checked with the vulnerability info resource (as opposed to direct packages getting checked with the package metadata resource). But it's not using audit sources, it's using package sources.

Nigusu-Allehu: A table would definitely be easier to consume. There is only two scenarios now, but we could add the PMUI part to the table when it comes out

zivkan: probably, I was just writing a table as an example so we could see what it would look like.

nkolev92 (approved):

docs/concepts/Auditing-Packages.md
```diff
@@ -72,9 +72,13 @@ Note that the [V2 protocol is deprecated](../nuget-org/overview-nuget-org.md#api
...
-Prior to this version, NuGet Audit will only use package sources to download vulnerability information.
-Audit sources are not used by `dotnet list package --vulnerable` at this time.
+**Note**: The table below lists features that support Audit Sources.
+
+| Introduced In                                                                                | Feature Supporting Audit Sources                                   |
+| -------------------------------------------------------------------------------------------- | ------------------------------------------------------------------ |
+| [NuGet 6.12, .NET 9.0.100 SDK, and Visual Studio 2022 17.12](../release-notes/NuGet-6.12.md) | Restore                                                            |
+| [NuGet 6.14, .NET 9.0.300 SDK](../release-notes/NuGet-6.14.md)                               | `dotnet package list --vulnerable`                                 |
+| Not yet supported                                                                            | NuGet AuditSources support in the Visual Studio Package Manager UI |
```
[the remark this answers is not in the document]
nkolev92: https://github.com/nuget/home/issues/13954

PR #3516, opened 2025-12-15
Add note about third-party pull requests for GitHub Actions

eXpl0it3r:
I think it's important to communicate to the user, that pull requests by strangers won't be able to circumvent the trust publishing policy.

Without such a hint, it's unclear whether an "approved" user (someone who has contributed before and won't need a manual approval for the GitHub Actions run), would be able to create a PR that changes the GitHub Actions step in such a way, that malicious code is released as new package, because the trust publishing policy is set up for the given repository.

Feel free to make suggestions on the words or change the wording directly.

Pinging @joelverhagen, given that I have more or less used his wording from the .NET Foundation Discord server.

### DIFF
--- a/docs/nuget-org/trusted-publishing.md
+++ b/docs/nuget-org/trusted-publishing.md
@@ -56,7 +56,11 @@ To get started:
        > This corresponds to your workflow at `.github/workflows/build.yml`. Enter the **file name only** (`build.yml`)—do not include the `.github/workflows/` path.  
      - **Environment (optional):** `release` 
        > Enter environment if your workflow uses e.g. `environment: release` and you want to restrict this policy to that environment. Leave this empty if you do not use GitHub Actions environments.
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Thank you for looking at this PR and clarifying your statement, @joelverhagen!

Happy to update this!

Do you have any doc links regarding the "However, the pull_request event will fail to produce an OIDC token due to protections in the GitHub Actions platform." statement? Is this documented somewhere?

joelverhagen:
> Do you have any doc links regarding the "However, the pull_request event will fail to produce an OIDC token due to protections in the GitHub Actions platform." statement? Is this documented somewhere?

No, I was unable to find a doc. However, I have tested it and it is referenced on https://docs.pypi.org/trusted-publishers/security-model/. When I tested it, they implemented the block by not setting `ACTIONS_ID_TOKEN_REQUEST_TOKEN` and `ACTIONS_ID_TOKEN_REQUEST_URL`.

PR #3561, opened 2026-04-22
Add info about data.nuget.org

zivkan:
The following issue has been implemented, so document it making it easier to discover.
* https://github.com/NuGet/NuGetGallery/issues/10656

### DIFF
--- a/docs/reference/nuget-config-file.md
+++ b/docs/reference/nuget-config-file.md
@@ -161,6 +161,11 @@ Audit sources support the same attributes as `packageSources` (`protocolVersion`
 </auditSources>
 ```
 
+> [!TIP]
+> nuget.org also provides `https://data.nuget.org/v3/index.json`, a service index that only contains vulnerability data and doesn't serve packages.
+> Network administrators who block `api.nuget.org` to prevent package downloads might be willing to allow `data.nuget.org` if asked.
+> For more information, see [audit sources](../concepts/Auditing-Packages.md#audit-sources).
+
 ### packageSourceCredentials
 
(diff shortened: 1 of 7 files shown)


docs/reference/nuget-config-file.md
```diff
@@ -161,6 +161,11 @@ Audit sources support the same attributes as `packageSources` (`protocolVersion`
 </auditSources>
 ```
 
+> [!TIP]
+> nuget.org also provides `https://data.nuget.org/v3/index.json`, a service index that only contains vulnerability data and doesn't serve packages.
```
donnie-msft, line 165: What do you think about updating VS Options docs to also have a similar callout? 
This is linked from the Sources -> Audit Sources VS Options page today:

https://learn.microsoft.com/en-us/nuget/consume-packages/nuget-visual-studio-options#audit-sources

donnie-msft (dismissed):
Nice updates! Left one idea to update the VSOptions as well. I can do that separately if you want to leave your ideas on whether any UI changes should occur (eg, the VS Options watermark textbox could be changed to show the audit.nuget.org, or a message could indicate that it's an option, etc). 
Otherwise, just updating the docs page works for me. Let me know your thoughts on what to put there and I can help if needed.

docs/reference/nuget-config-file.md
zivkan: @copilot add a similar message in the managing audit sources section of the visual studio options docs

nkolev92 (approved):

donnie-msft (approved):

PR #3571, opened 2026-05-01
Add recommendation to pin .NET SDK version

### LINKED ISSUE #14867: Lock file docs should mention it's recommended to use `global.json` with `rollForward = disable` when using lock files
### Docs Page URL

https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#locking-dependencies

### Description of the Issue

See comment from Microsoft(Chet) here https://github.com/dotnet/sdk/issues/48795#issuecomment-3001803076
> Our current recommendation for users that use lock files is to also lock their SDK versions via global.json with no rollforward, so they(sic) their entire toolchain stays in lockstep.

Discovery of this recommendation is currently hard

### Suggested Fix

Add some version of: Our current recommendation for users that use lock files is to also lock their SDK versions via global.json with no rollforward, so they their entire toolchain stays in lockstep.

Ideally linking to an issue to track for when that recommendation changes or at least the specific issues which most users run into which require pinning the SDK in relation to lock-files
https://github.com/dotnet/aspnetcore/issues/64897
https://github.com/dotnet/sdk/issues/48795

Ideally the blog post https://devblogs.microsoft.com/dotnet/enable-repeatable-package-restores-using-a-lock-file/ is also updated

Opened related `global.json` docs issue here https://github.com/dotnet/docs/issues/53347

### Additional Context

https://github.com/dotnet/sdk/issues/48795
https://github.com/dotnet/aspnetcore/issues/65061
https://github.com/microsoft/azure-pipelines-tasks/issues/22040

Frulfump:
Fix https://github.com/NuGet/Home/issues/14867
Added a note about pinning the .NET SDK version to ensure consistency in the dependency graph.

### DIFF
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -341,7 +341,7 @@ Input to NuGet restore is a set of `PackageReference` items from the project fil
 - A given package version is removed from the repository. Though nuget.org does not allow package deletions, not all package repositories have this constraint. This results in NuGet finding the best match when it cannot resolve to the deleted version.
 
 > [!Note]
-> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
+> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](/dotnet/core/tools/global-json). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
 
 ### Enabling the lock file
 
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -341,7 +341,7 @@ Input to NuGet restore is a set of `PackageReference` items from the project fil
 - A given package version is removed from the repository. Though nuget.org does not allow package deletions, not all package repositories have this constraint. This results in NuGet finding the best match when it cannot resolve to the deleted version.
 
 > [!Note]
-> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](/dotnet/core/tools/global-json). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
+> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](/dotnet/core/tools/global-json#rollforward). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
 
 ### Enabling the lock file
 
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -340,6 +340,9 @@ Input to NuGet restore is a set of `PackageReference` items from the project fil
 
 - A given package version is removed from the repository. Though nuget.org does not allow package deletions, not all package repositories have this constraint. This results in NuGet finding the best match when it cannot resolve to the deleted version.
 
+> [!Note]
+> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
+
 ### Enabling the lock file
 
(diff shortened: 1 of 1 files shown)


Frulfump:
@dotnet-policy-service agree

Frulfump:
The links by the bot comment above https://learn.microsoft.com/en-us/help/platform/validation-ref/docs-link-absolute , https://learn.microsoft.com/en-us/help/platform/validation-ref/hard-coded-locale, https://learn.microsoft.com/en-us/help/platform/faq?tabs=pr-validation-error-solution%2Cpr-merge-conflict-error-solution#build--publish gives me a 404

And the preview URL asks me to login

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -340,6 +340,9 @@ Input to NuGet restore is a set of `PackageReference` items from the project fil
 
 - A given package version is removed from the repository. Though nuget.org does not allow package deletions, not all package repositories have this constraint. This results in NuGet finding the best match when it cannot resolve to the deleted version.
 
+> [!Note]
+> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
```
Frulfump: Fix validation issues https://github.com/NuGet/docs.microsoft.com-nuget/pull/3571#issuecomment-4358167269
```suggestion
> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](/dotnet/core/tools/global-json). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
```

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -340,6 +340,9 @@ Input to NuGet restore is a set of `PackageReference` items from the project fil
 
 - A given package version is removed from the repository. Though nuget.org does not allow package deletions, not all package repositories have this constraint. This results in NuGet finding the best match when it cannot resolve to the deleted version.
 
+> [!Note]
+> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](/dotnet/core/tools/global-json). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
```
Frulfump: Not sure how to validate the relative links as they give me a 404 but the validation bot says everything is green, I'd still like to verify for myself

Frulfump: Fix link
```suggestion
> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](/dotnet/core/tools/global-json#rollforward). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
```

docs/consume-packages/Package-References-in-Project-Files.md
```diff
@@ -340,6 +340,9 @@ Input to NuGet restore is a set of `PackageReference` items from the project fil
 
 - A given package version is removed from the repository. Though nuget.org does not allow package deletions, not all package repositories have this constraint. This results in NuGet finding the best match when it cannot resolve to the deleted version.
 
+> [!Note]
+> It's also recommended to pin the version of the .NET SDK that's used so the SDK version and dependency graph stay in lockstep. See [global.json](/dotnet/core/tools/global-json) and especially the `rollForward` section with the `disable` value [global.json rollForward Policy](/dotnet/core/tools/global-json#rollforward). For related issues when not pinning, see [ASP.NET Core GitHub issue](https://github.com/dotnet/aspnetcore/issues/65061) and [.NET Core SDK GitHub issue](https://github.com/dotnet/sdk/issues/48795)
```
nkolev92: I think we have too many notes, so let's have this content as just a regular paragraph.
Let's do one sentence per line. 

My take on this is that I think we should provide some explanation on the motivation. The SDK version pinning is something customers may need, but I don't necessarily want customers using lock file to be tying themselves to potentially older versions of the SDK unless they absolutely needed. 

So in there I'd basically talk about the fact that .NET SDK wills ometimes automatically reference packages and those package versions may change as the .NET SDK tooling changes and when customers see these issues, then they should pin.

cc @baronfel for a gut check.

Frulfump: I can reformat.

More explanations are always good, for lock files specifically it might be the current (see below) recommendation but for things like reproducible build (outside of this document but somewhat related?) it would still make sense to pin the version used. Yes for those not using tools like Depandbot or Renovate it might be more easily missed but with knowledgeable/skilled devs you would always check `global.json` when first starting to check out a new repo. (Maybe something should be built(built-in) that warns when the SDK contains known vulns or is just outdated? Both CLI and IDE (Visual Studio(/Rider)) (and VS Code))

It's based on this comment from Chet ~1 year ago https://github.com/dotnet/sdk/issues/48795#issuecomment-3001803076 
> It's the recommended way today, but it is a gap in the overall NuGet lockfile story. Our current recommendation for users that use lock files is to also lock their SDK versions via global.json with no rollforward, so they(sic) their entire toolchain stays in lockstep.

But yeah very interested in hearing Chet's updated takes here, will give it some time before making any adjustments

zivkan: Sorry it took me so long to review. I've had time management challenges following a week of being sick. Anyway, I don't like that this is suggesting everyone who uses lock files to also use a global.json. This issue only affects a limited set of project types. For example, I did `dotnet new mvc`, and that project didn't have any implicit package references, so those projects don't gain any benefits from locking the sdk via global.json, and only have the downsides.

Here's my suggestion:

```suggestion
> For some project types, the .NET SDK adds implicit `PackageReference` items, and using a different SDK version will change the package version, leading to locked mode restore failures.
> To mitigate this issue, you can use a [global.json file](/dotnet/core/tools/global-json), setting the [global.json rollForward Policy](/dotnet/core/tools/global-json#rollforward) to disable.
> This means that all developers working in this repo, and your CI pipelines, will need to use the same version of the SDK at all times, but it allows you to control when the implicit packages change versions.
```

Frulfump: Ah no worries. Sure that's fine I'll take your suggestion but still interested in the SDK teams thoughts as well and it's not Only downside as you get a guaranteed version of the SDK and the risk for issues is more limited then when compared to rolling forward. There's been cases of new warnings or other bugs in new patch versions.

nkolev92: @baronfel 
Do you have any guidance here?

Frulfump: Suggestion taken, still interested in the SDK teams thoughts and input on my statement about the not only downside thing.

zivkan (approved):

PR #3576, opened 2026-06-03
Fixing typo in the configuration example

exoosh:
* The system-wide NuGet.config should be in %ProgramFiles(x86)% as per the documentation
  - Also, %appdata% would not require any elevation

-----

Correct me, if I am wrong, but looking at the corresponding mac/Linux example the intent here is to use the global (not user-writable) `.config`.


exoosh:
Also, whoever is maintaining this, I think the path for user-level and computer-level is inconsistent.

* computer-level ⇒ **directory path**:
  * `%ProgramFiles(x86)%\NuGet\Config`
  * `/etc/opt/NuGet/Config`
* user-level ⇒ **full file path**:
  * `%appdata%\NuGet\NuGet.Config`
  * `~/.config/NuGet/NuGet.Config`

Nigusu-Allehu (approved):

donnie-msft (approved):

PR #3578, opened 2026-06-04
Add docs for Deterministic and DeterministicTimestamp

omajid:
For details about this, see:
https://github.com/NuGet/Home/blob/dev/accepted/2026/deterministic-pack-revisited.md

### DIFF
--- a/docs/reference/cli-reference/cli-ref-pack.md
+++ b/docs/reference/cli-reference/cli-ref-pack.md
@@ -38,6 +38,14 @@ where `<nuspecPath>` and `<projectPath>` specify the `.nuspec` or project file, 
 
   The NuGet configuration file to apply. If not specified, `%AppData%\NuGet\NuGet.Config` (Windows), or `~/.nuget/NuGet/NuGet.Config` or `~/.config/NuGet/NuGet.Config` (Mac/Linux) is used.
 
+- **`-Deterministic`**
+
+  *(7.9+)* Build packages in deterministic mode. This produces bit-identical packages for the same input (source code, project file, build configuration), with a known caveat of timestamps. See `DeterministicTimestamp` on how to reproduce with identical date/times.
+
+- **`-DeterministicTimestamp`**
+
+  *(7.9+)* Specify a timestamp that's used as the file modification date/time for all files added to the NuGet package. It can be a date/time in RFC 3339 format (such as `1996-12-19T16:39:57-08:00`), it can be the number of non-leap seconds since 00:00:00 UTC on 1 Jan 1970 (also known as the Unix time), or it can be the special string `true` (default, use the current wall clock time), or `false` to not modify file times when creating deterministic packages. This option is ignored if `Deterministic` is not enabled.
+
 - **`-Exclude`**
 
   Specifies one or more wildcard patterns to exclude when creating a package. To specify more than one pattern, repeat the -Exclude flag. See example below.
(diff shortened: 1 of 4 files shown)


docs/reference/cli-reference/cli-ref-pack.md
```diff
@@ -38,6 +38,14 @@ where `<nuspecPath>` and `<projectPath>` specify the `.nuspec` or project file,
 
   The NuGet configuration file to apply. If not specified, `%AppData%\NuGet\NuGet.Config` (Windows), or `~/.nuget/NuGet/NuGet.Config` or `~/.config/NuGet/NuGet.Config` (Mac/Linux) is used.
 
+- **`-Deterministic`**
+
+  Build packages in deterministic mode. This produces bit-identical packages for the same input (source code, project file, build configuration), with a known caveat of timestamps. See `DeterministicTimestamp` on how to reproduce with identical date/times.
+
+- **`-DeterministicTimestamp`**
```
zivkan, line 45: both of these new arguments need to specify that it's available from NuGet 7.9. See options `-ForceEnglishOutput`, `-MSBuildPath`, `-MSBuildVersion`, and `-Suffix` for examples how we did that for other options.

docs/reference/msbuild-targets.md
```diff
@@ -116,6 +116,9 @@ The following table describes the MSBuild properties that can be added to a proj
...
+| `Deterministic` | This Boolean value specifies whether to enable deterministic mode and use the same timestamp for all files added to a package. |
+| `DeterministicTimestamp` | This string value specifies a timestamp that's used as the file modification date/time for all files added to the NuGet package. It can be a date/time in RFC 3339 format (such as `1996-12-19T16:39:57-08:00`), it can be the number of non-leap seconds since 00:00:00 UTC on 1 Jan 1970 (also known as the Unix time), it can be the special string `true` (default, use the current wall clock time), or `false` to not modify file times when creating deterministic packages. This option is ignored if `Deterministic` is not `true`. |
+| `SOURCE_DATE_EPOCH` | Specifies the number of non-leap seconds since 00:00:00 UTC on 1 Jan 1970 (also known as the Unix time). This is used automatically in place of `DeterministicTimestamp` if that's not defined and `SOURCE_DATE_EPOCH` is. See [the `SOURCE_DATE_EPOCH` specification](https://reproducible-builds.org/docs/source-date-epoch/) |
```
zivkan: These also need comments to specify that it's only available from NuGet 7.9. Some of the other rows have examples of how it's specified in this document.

docs/reference/cli-reference/cli-ref-pack.md
nkolev92, line 45: Can we include the version somewhere to indicate this is the default in .NET 11.

docs/reference/cli-reference/cli-ref-pack.md
```diff
@@ -38,6 +38,14 @@ where `<nuspecPath>` and `<projectPath>` specify the `.nuspec` or project file,
 
   The NuGet configuration file to apply. If not specified, `%AppData%\NuGet\NuGet.Config` (Windows), or `~/.nuget/NuGet/NuGet.Config` or `~/.config/NuGet/NuGet.Config` (Mac/Linux) is used.
 
+- **`-Deterministic`**
```
nkolev92, line 41: I was wondering if we can add the general deterministic documentation somewhere in https://github.com/NuGet/docs.microsoft.com-nuget/tree/main/docs/create-packages, but I wasn't quite sure where.

Copilot suggested docs/create-packages/Package-authoring-best-practices.md, but I think the best guidance is not obvious. It's probably just use the defaults.

nkolev92 (commented):
Thanks for starting this doc.

docs/reference/cli-reference/cli-ref-pack.md
omajid: I am quite unfamiliar with how mixing/matching.NET and NuGet versions work.

This is the NuGet cli (`nuget`), right? It will ignore the .NET default, no?

nkolev92: ah fair, call out 6.9 here then.

nkolev92: or rather 7.9 whioch is done. 
Resolving

nkolev92 (dismissed):

nkolev92 (approved):

omajid:
Not sure what the next steps here are. Can this be merged?

PR #3580, opened 2026-06-12
Document IsImplicitlyDefined for MSBuild project SDKs

### LINKED ISSUE #13529: Document IsImplicitlyDefined
Document the `IsImplicitlyDefined` metadata on `PackageReference` items, that MSBuild SDKs are expected to set to any PackageReference the SDK itself defines, so that Visual Studio's PackageManager doesn't try to offer it as an upgradable package that will fail if the upgrade is attempted.  It also makes the PackageReference work for projects using CPM.

donnie-msft:
Fixes: https://github.com/NuGet/Home/issues/13529

### DIFF
--- a/docs/consume-packages/Package-References-in-Project-Files.md
+++ b/docs/consume-packages/Package-References-in-Project-Files.md
@@ -261,6 +261,35 @@ namespace PackageReferenceAliasesExample
 
 ```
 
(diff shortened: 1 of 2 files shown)


donnie-msft:
Moved to https://github.com/NuGet/docs.microsoft.com-nuget/pull/3586

PR #3583, opened 2026-06-24
Fix config paths

zivkan:
in https://github.com/NuGet/docs.microsoft.com-nuget/pull/3576, a customer noticed some of the paths documented were wrong.  I pointed copilot at both the docs repo and the NuGet.Client repo, and told it to find all the other errors. This PR are the fixes it found.

### DIFF
--- a/docs/consume-packages/configuring-nuget-behavior.md
+++ b/docs/consume-packages/configuring-nuget-behavior.md
@@ -17,8 +17,8 @@ NuGet's behavior is driven by the accumulated settings in one or more config (XM
 | Scope | `NuGet.Config` file location | Description |
 | --- | --- | --- |
 | Solution | Current folder (aka Solution folder) or any folder up to the drive root.| In a solution folder, settings apply to all projects in subfolders. Note that if a config file is placed in a project folder, it has no effect on that project. When restoring a project on the command line, the project's directory is treated as the solution directory, which can lead to differences in behaviour when restoring the project vs solution. |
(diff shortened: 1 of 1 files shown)


docs/consume-packages/configuring-nuget-behavior.md
```diff
@@ -87,7 +87,8 @@ dotnet nuget config set repositoryPath /home/projects/packages --configfile /hom
 dotnet nuget config set repositoryPath /home/packages --configfile home/myApp/NuGet.Config
 
 # Set repositoryPath in the computer-level file (requires elevation)
-dotnet nuget config set repositoryPath /home/packages --configfile $XDG_DATA_HOME/NuGet.Config
+# On Linux the computer-level config is under /etc/opt/NuGet/Config; on Mac it's under /Library/Application Support/NuGet/Config
```
donnie-msft, line 90: Should we quote paths for readability?

donnie-msft (approved):
Looks good, and I have these questions:

1. So, is `XDG_DATA_HOME` not used by NuGet at all? If it is, should we keep some mention of it?
2. Does this PR also address https://github.com/NuGet/Home/issues/13596 ?

docs/consume-packages/configuring-nuget-behavior.md
donnie-msft, line 1: Should we delete this part?
> On Linux, this previous location was `/etc/opt`, and on Mac, `/Library/Application Support`. 

...from
https://github.com/NuGet/docs.microsoft.com-nuget/blob/25db6815b3e6d1ff76c813a9ae2ea76a352c7c09/docs/consume-packages/configuring-nuget-behavior.md?plain=1#L42

...to potentially fix this issue?
https://github.com/NuGet/Home/issues/13596

PR #3584, opened 2026-06-29
Document NUGET_SCRATCH environment variable

### LINKED ISSUE #13567: Document NUGET_SCRATCH environment variable
Hi,
The variable has been introduced in this pull request - https://github.com/NuGet/NuGet.Client/pull/4511
The associated issue should describe the rationale.
Thank you.

zivkan:
Fixes NuGet/Home#13567

Documents the `NUGET_SCRATCH` environment variable, which overrides the default temporary folder NuGet uses for its operations (including the lock files that synchronize concurrent NuGet processes).

The variable was introduced in NuGet 6.2 via [NuGet/NuGet.Client#4511](https://github.com/NuGet/NuGet.Client/pull/4511). When unset, NuGet defaults to `NuGetScratch` under the system temp folder (for example, `%TEMP%\NuGetScratch` on Windows and `/tmp/NuGetScratch<username>` on Linux and macOS).

Added a row to the NuGet CLI environment variables reference table in `docs/reference/cli-reference/cli-ref-environment-variables.md`.


Nigusu-Allehu (dismissed):

nkolev92 (dismissed):
At some point we should probably have 2 docs, CLI ref env vars and just general env var. 

Maybe we could use the `include` composition that docs repo allows so that we don't have to lose the content here. 

Seet the bottom https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/concepts/Dependency-Resolution.md

kartheekp-ms (dismissed):
There is a mention about this environment variable [here](https://github.com/NuGet/docs.microsoft.com-nuget/blob/2113360bbc746c28f19da74a5c7336972b49d5e4/docs/consume-packages/managing-the-global-packages-and-cache-folders.md?plain=1#L19) but it is great to add to this doc also.

Nigusu-Allehu (dismissed):

nkolev92 (approved):

PR #3594, opened 2026-07-16
Clarify that Package Source Mapping applies only to restore/install/update

### LINKED ISSUE #15001: [Docs] Security Best Practices & .NET CLI commands are unclear about Package Source Mapping
### Docs Page URL

https://learn.microsoft.com/en-us/nuget/concepts/security-best-practices

### Description of the Issue

Package Source Mapping applies to package download.
The CLI commands that query with package ID but do not currently respect Package Source Mapping should have this called out in their docs.

[Best practices for a secure software supply chain](https://learn.microsoft.com/en-us/nuget/concepts/security-best-practices) also doesn't explain current limitations of package source mapping.

### Suggested Fix

_No response_

### Additional Context

_No response_

donnie-msft:
Fixes: https://github.com/NuGet/Home/issues/15001

- **`Package-Source-Mapping.md`** — Added a scope note (and expanded the limitations list) stating PSM only affects the restore/install/update download path. Sets clear expectations up front so users know metadata commands aren't covered.
- **`Security-Best-Practices.md`** — Clarified in the **NuGet feeds** and **Package Source mapping** sections that metadata commands send package IDs to all sources, with tracking issues grouped by tool. Gives security-minded readers informed consent about the privacy implication where they'd look for it.
- **`nuget-config-file.md`** — Added a scope note to the `packageSourceMapping` section. Warns users at the point of configuration that PSM doesn't cover metadata operations.
- **`Deprecate-packages.md`** — Noted that `dotnet list package --deprecated` queries all sources. 
- **Scope correctness** — Excluded non-leaking commands (`nuget search`/`nuget list`, `dotnet package search`, `dotnet nuget why`) and dropped `dotnet package update` (already honors PSM per https://github.com/NuGet/Home/issues/14307).

### DIFF
--- a/docs/concepts/Security-Best-Practices.md
+++ b/docs/concepts/Security-Best-Practices.md
@@ -190,7 +190,7 @@ To ensure your build is predictable and secure from known attacks such as [Depen
 You can use a single feed or private feed with upstreaming capabilities for protection.
 
 When you configure both a public and a private source, client commands that query package metadata (for example, `dotnet package add` and `dotnet list package --outdated/--deprecated/--vulnerable`) send the requested package IDs to every configured source, including public ones.
-[Package Source Mapping](../consume-packages/package-source-mapping.md) doesn't prevent this, because it filters only the package download path used by restore, install, and update.
+[Package Source Mapping](../consume-packages/package-source-mapping.md) doesn't prevent this, because it applies only when NuGet downloads packages during restore, install, and update.
 For the current status of these feature gaps, see the following tracking issues:
 
 * dotnet.exe package commands (`add`, `list package`): [NuGet/Home#12766](https://github.com/NuGet/Home/issues/12766) and [NuGet/Home#11380](https://github.com/NuGet/Home/issues/11380).
@@ -225,8 +225,8 @@ To enable lock files, [see the following documentation](../consume-packages/pack
 
 Package Source Mapping allows you to centrally declare which source each package in your solution should restore from in your nuget.config file.
 
-Package Source Mapping applies to the package download path used by restore, install, and update.
-It doesn't filter the sources that other client commands query for package metadata, so it isn't a complete privacy control on its own.
+Package Source Mapping currently applies only when NuGet downloads packages during restore, install, and update.
+A current limitation is that it doesn't yet filter the sources that other client commands query for package metadata.
 For more information, see [NuGet feeds](#nuget-feeds).
 
 To enable package source mapping, [see the following documentation](../consume-packages/package-source-mapping.md).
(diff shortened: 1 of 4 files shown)


docs/concepts/Security-Best-Practices.md
```diff
@@ -216,6 +225,10 @@ To enable lock files, [see the following documentation](../consume-packages/pack
 
 Package Source Mapping allows you to centrally declare which source each package in your solution should restore from in your nuget.config file.
 
+Package Source Mapping applies to the package download path used by restore, install, and update.
```
nkolev92: I'd just say, PSM has a limitation with the follow commands. 

I think splitting the 2 makes it sound like this is a behavior that will remain, but in practice, we actually want to fix it.

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -18,6 +18,19 @@ With Package Source Mapping, you can filter, per package, which source(s) NuGet
 
 We also have suggestions for other [best practices](..\concepts\Security-Best-Practices.md) to help you fortify your supply chain against attacks.
 
+> [!IMPORTANT]
+> Package Source Mapping currently applies only to the package download path used by restore, install, and update operations.
+> It doesn't yet filter the sources that other client commands query for package metadata.
+> Commands such as [`dotnet package add`](/dotnet/core/tools/dotnet-package-add) and [`dotnet list package`](/dotnet/core/tools/dotnet-list-package) (including `--outdated`, `--deprecated`, and `--vulnerable`) and the Package Manager UI's browse and update metadata lookups send the requested package IDs to *all* configured sources, including public ones.
+> If you use a mix of public and private sources, the names of your private packages can reach a public source through these metadata operations even when Package Source Mapping is configured.
```
nkolev92: nit: this is restating line 24 with different words.

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -204,4 +217,5 @@ For an idea of how your source mappings may look like, refer to our [samples rep
 >
 > * There are no nuget.exe or dotnet.exe commands for managing the package source mapping configuration, see [NuGet/Home#10735](https://github.com/NuGet/Home/issues/10735).
 > * There are no means of mapping packages at package installation time, see [NuGet/Home#10730](https://github.com/NuGet/Home/issues/10730).
+> * Package Source Mapping applies only to the package download path used by restore, install, and update. It doesn't apply to commands that query source metadata, such as `dotnet package add` and `dotnet list package --outdated/--deprecated/--vulnerable`. Those commands send package IDs to all configured sources.
```
nkolev92: I see the term "download" path being used and while technically correct, it is not something we've defined, so I wonder if there's simpler phrasing here.

docs/consume-packages/Package-Source-Mapping.md
```diff
@@ -18,6 +18,18 @@ With Package Source Mapping, you can filter, per package, which source(s) NuGet
 
 We also have suggestions for other [best practices](..\concepts\Security-Best-Practices.md) to help you fortify your supply chain against attacks.
 
+> [!IMPORTANT]
```
nkolev92: The content looks good now, thanks for iterating and addressing all the admittedly opinionated feedback. 

Do we need to put this is an important section or should just be part of the general documentation. 

I like how it ready in Security-Best-Practices. I think it sohuld just be general documentation in the other places too.

donnie-msft: Changed it to a "Scope and limitations". Other docs have "Limitations" sections, so this seems in line with that.

nkolev92 (approved):

PR #3597, opened 2026-07-24
Doc for package signing certificates on NuGet.org

pranathibora14:
Adds a new server-side doc page describing the account certificates feature and its enrollment consequence, addressing https://github.com/NuGet/NuGetGallery/issues/10910.

### DIFF
--- a/docs/nuget-org/package-signing-certificates.md
+++ b/docs/nuget-org/package-signing-certificates.md
@@ -9,11 +9,14 @@ ms.topic: how-to
 
 # Package signing certificates on NuGet.org
 
-NuGet.org lets you register one or more code signing certificates on your individual account or organization. You manage registered certificates in the **Certificates** section of the account or organization settings page.
+NuGet.org lets you register one or more code signing certificates on your individual account or organization.
+You manage registered certificates in the **Certificates** section of the account or organization settings page.
 
-Once you register at least one certificate, NuGet.org requires every package that account pushes to be author-signed with one of the registered certificates. This is referred to as **required author signing**.
+Once you register at least one certificate, NuGet.org requires every package that account pushes to be author-signed with one of the registered certificates.
+This is referred to as **required author signing**.
 
-This article describes the effect of registering a certificate and how to manage certificates on NuGet.org. For how package signatures work and the requirements NuGet.org enforces on each signature, see [Signed package reference](../reference/Signed-Packages-Reference.md).
+This article describes the effect of registering a certificate and how to manage certificates on NuGet.org.
+To learn how package signatures work and the requirements NuGet.org enforces on each signature, see [Signed package reference](../reference/Signed-Packages-Reference.md).
 
 ## What happens when you register a certificate
 
@@ -25,10 +28,14 @@ Once you register **one or more** certificates on an account:
 - NuGet.org **rejects packages author-signed with a certificate that isn't registered** to the account.
 - Packages author-signed with **any one** of the registered certificates are accepted, provided the signature also meets the [signature requirements on NuGet.org](#signature-requirements-on-nugetorg).
 
-This applies to **new versions** you push after enrollment. Package versions that were already published remain available and are not affected.
+This applies to **new versions** you push after enrollment.
+Package versions that were already published remain available and are not affected.
 
 > [!IMPORTANT]
-> After you register a certificate, you must author-sign every subsequent push from that account with a registered certificate. If you push an unsigned package, or a package signed with an unregistered certificate, the push is rejected. Author signing is produced with `nuget.exe sign`. For a walkthrough, see [Sign a package](../create-packages/Sign-a-Package.md).
+> After you register a certificate, you must author-sign every subsequent push from that account with a registered certificate.
+> If you push an unsigned package, or a package signed with an unregistered certificate, the push is rejected.
+> Author signing is produced with `nuget.exe sign`.
+> For a walkthrough, see [Sign a package](../create-packages/Sign-a-Package.md).
 
 ## Signature requirements on NuGet.org
 
@@ -43,7 +50,7 @@ NuGet.org has additional requirements for accepting a signed package:
   - Must be valid for its purpose:
     - The author signing certificate must be valid for code signing.
     - The timestamp certificate must be valid for timestamping.
-  - Must not be revoked at signing time. (This may not be knowable at submission time, so NuGet.org periodically rechecks revocation status.)
+  - Must not be revoked at signing time. (This might not be knowable at submission time, so NuGet.org periodically rechecks revocation status.)
 
 ## Register a certificate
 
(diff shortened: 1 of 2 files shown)


docs/nuget-org/package-signing-certificates.md
```diff
@@ -0,0 +1,83 @@
...
+description: Register signing certificates on your NuGet.org account or organization to require author-signed packages.
+author: pranathibora14
+ms.author: prabora
+ms.date: 07/23/2026
+ms.topic: how-to
+---
+
+# Package signing certificates on NuGet.org
+
+NuGet.org lets you register one or more code signing certificates on your individual account or organization. You manage registered certificates in the **Certificates** section of the account or organization settings page.
```
zivkan: markdown is source code for pages that will get rendered as html. As source code, I suggest optimizing it for editing, not reading, so put 1 line per sentence. The final html will be the same.

Sample applies for the whole document, not just this line, I see some other paragraphs on a single line, that would be nice to change to 1 line per sentence

zivkan (dismissed):

docs/nuget-org/package-signing-certificates.md
pranathibora14: That makes sense. I have changed all paragraphs to one line per sentence

docs/nuget-org/package-signing-certificates.md
```diff
@@ -0,0 +1,95 @@
...
+- NuGet.org **rejects unsigned packages** pushed by that account.
+- NuGet.org **rejects packages author-signed with a certificate that isn't registered** to the account.
+- Packages author-signed with **any one** of the registered certificates are accepted, provided the signature also meets the [signature requirements on NuGet.org](#signature-requirements-on-nugetorg).
+
+This applies to **new versions** you push after enrollment.
+Package versions that were already published remain available and are not affected.
+
+> [!IMPORTANT]
+> After you register a certificate, you must author-sign every subsequent push from that account with a registered certificate.
+> If you push an unsigned package, or a package signed with an unregistered certificate, the push is rejected.
```
donnie-msft, line 36: How is it rejected? Is there an HTTP response code that fails the push command? If so, I'd like to see the HTTP code or similar here.
Does it look like it gets pushed but fail validation later?

docs/nuget-org/package-signing-certificates.md
```diff
@@ -0,0 +1,95 @@
...
+> Author signing is produced with `nuget.exe sign`.
+> For a walkthrough, see [Sign a package](../create-packages/Sign-a-Package.md).
+
+## Signature requirements on NuGet.org
```
dtivel, line 40: There's a lot of information from here onwards that is duplicative of existing articles, such as:

- https://learn.microsoft.com/nuget/reference/signed-packages-reference#signature-requirements-on-nugetorg
- https://learn.microsoft.com/nuget/create-packages/sign-a-package#register-the-certificate-on-nugetorg
 
It makes more sense to me to either refactor docs to collocate topics in one doc or link to existing sections.

docs/nuget-org/package-signing-certificates.md
```diff
@@ -0,0 +1,95 @@
...
+ms.topic: how-to
+---
+
+# Package signing certificates on NuGet.org
+
+NuGet.org lets you register one or more code signing certificates on your individual account or organization.
+You manage registered certificates in the **Certificates** section of the account or organization settings page.
+
+Once you register at least one certificate, NuGet.org requires every package that account pushes to be author-signed with one of the registered certificates.
+This is referred to as **required author signing**.
```
dtivel, line 16: This is too broad. The enforcement path computes the author-signing policy **per package registration** from that package ID's owners and optional required signer. It is not determined by the individual or organization account that happens to perform the push.

- **No required signer:** signing is allowed when any owner has a registered certificate and required only when every owner has one. A signed package may use a certificate registered to any owner.
- **A required signer is set:** only that account's certificate state matters for these signing calculations. If it has certificates, signing is required and only its certificates are accepted. If it has none, asynchronous signature validation rejects every author-signed submission, even when another owner has certificates.
- The quoted behavior is the common single-owner case when the package is solely owned by that account and no different required signer remains set. It does not hold for co-owned packages or packages with a different required signer. A publisher may also push on behalf of multiple organizations, each with independent certificate registrations, while each package ID retains its own policy.

The authoritative calculations are [`IsSigningAllowed`, `IsSigningRequired`, and `IsAcceptableSigningCertificate`](https://github.com/NuGet/NuGetGallery/blob/9f94d54037e88c1e0e106d373aa804988984d5ec/src/NuGet.Services.Entities/Extensions/PackageRegistrationExtensions.cs#L18-L103). The asynchronous enforcement is in [`ValidatePackageRegistrationSigningRequirementsAsync`](https://github.com/NuGet/NuGetGallery/blob/9f94d54037e88c1e0e106d373aa804988984d5ec/src/Validation.PackageSigning.ProcessSignature/SignatureValidator.cs#L603-L659). The upload flow separately resolves the account represented by the API key, which may be an organization: [`ApiController`](https://github.com/NuGet/NuGetGallery/blob/9f94d54037e88c1e0e106d373aa804988984d5ec/src/NuGetGallery/Controllers/ApiController.cs#L643-L664).

One qualification: an account subscribed to the internal `AutomaticallyOverwriteRequiredSignerPolicy` has itself set as required signer across all package IDs it owns when its first certificate is registered ([controller](https://github.com/NuGet/NuGetGallery/blob/9f94d54037e88c1e0e106d373aa804988984d5ec/src/NuGetGallery/Controllers/AccountsController.cs#L513-L519), [service](https://github.com/NuGet/NuGetGallery/blob/9f94d54037e88c1e0e106d373aa804988984d5ec/src/NuGetGallery.Services/PackageManagement/PackageService.cs#L1077-L1129)). That broad effect is implemented by updating each package registration. It is not the general owner-certificate rule described above.

I suggest replacing these sentences with a per-package explanation, then describing the ordinary single-owner case as the simple special case.

> Registering a certificate can change the author-signing requirements for package IDs associated with that account. NuGet.org determines these requirements separately for each package ID, based on its owners and any configured required signer. When future versions must be author-signed, this is referred to as required author signing.

dtivel (changes requested):

docs/nuget-org/package-signing-certificates.md
pranathibora14: The package does get pushed and then fails the async validation phase. The package does not get published and shows up as 'Failed Validation' on package status page. 

I've changed the wording in the doc to better reflect this.

docs/nuget-org/package-signing-certificates.md
pranathibora14: That's a good point. I've linked to existing sections to avoid duplication.

docs/nuget-org/package-signing-certificates.md
pranathibora14: I've made the explanation per-package and also explained the no required signer vs required signer conditions to be more accurate.

dtivel (dismissed):

docs/nuget-org/package-signing-certificates.md
```diff
@@ -0,0 +1,83 @@
...
+- If the account still has **at least one** registered certificate after removal, required author signing remains in effect and packages must be signed with one of the remaining certificates.
+- If you remove the **last** registered certificate, the account is no longer enrolled in required author signing. The account can once again push unsigned packages or packages signed with any valid certificate.
+
+## Packages with multiple owners
+
+A package can have multiple owners, and each owner can independently register certificates.
+NuGet.org computes the author-signing policy per package ID from that package's owners and any configured required signer:
+
+- **No required signer:** signing is allowed when any owner has a registered certificate and required only when every owner has one. A signed package can use a certificate registered to any owner.
+- **A required signer is set:** only that account's certificate state matters for these signing calculations. If it has certificates, signing is required and only its certificates are accepted. If it has none, asynchronous signature validation rejects every author-signed submission, even when another owner has certificates.
```
donnie-msft:
```suggestion
- **A required signer is set:** only that account's certificate state matters for these signing calculations. If it has certificates, signing is required and only if its certificates are accepted. If it has none, asynchronous signature validation rejects every author-signed submission, even when another owner has certificates.
```

donnie-msft (dismissed):

docs/nuget-org/package-signing-certificates.md
pranathibora14: I think the "if" is making the sentence a little hard to understand. How about 

> **A required signer is set:** only that account's certificate state matters for these signing calculations. If it has certificates, signing is required, and only its certificates are accepted.

donnie-msft: My bad, must have read it wrong yesterday.

donnie-msft (approved):

